PR #313, opened 2024-12-13
bump test infra to v20241213-212009

jeanschmidt:
Includes (AFAIK):

* https://github.com/pytorch/test-infra/commit/1577e6b8541089412294fa71554af1b89be19cfe

PR #314, opened 2024-12-16
making IMDSv2 mandatory

wdvr:
See pytorch/test-infra#6061 for changes to the runners


zxiiro (approved):

zxiiro:
I think this obsoletes #313 so we likely can close the other one.

PR #320, opened 2025-01-13
Update AMI pinning  to al2023-ami-2023.6.*

jeanschmidt:
update the AMI pinning so it reflects what is already on `pytorch-labs/pytorch-gha-infra` and is on all variants of `scale-config.yml`


huydhn (approved):

malfet (commented):
Any docs what this change supposed to be doing, especially the  `amzn2`->`al2023` part?

jeanschmidt:
@malfet just outdated code that was not updated by laziness after the migration to amzn2023.

PR #334, opened 2025-05-30
Reduce minimum runners to 4

zxiiro:
Back in September we increased this from 0 to 6 to mitigate issues for runner types that were not frequently used where the Autoscaler dropped scale-up requests resulting in no runners being available for jobs. At time we were also testing LF runners at 95% - 100% load balance which is no longer the case.

Let's reduce the min_available_runners again to try to reduce idle costs on the account. Lets try 4 for now and reduce it further if no issues reported.


ZainRizvi (approved):
Sounds like a reasonable experiment to me, especially given the change a few months ago to setup a cron job that checks for jobs stuck in a queued state and provision runners accordingly.  

Would be good to keep an eye on queue times though to see if they go up after this

PR #335, opened 2025-06-02
Reduce the amount of time we wait before reclaiming an instance

ZainRizvi:
The original value for the minimum running time in production was actually 10 mins ([source PR](https://github.com/pytorch-labs/pytorch-gha-infra/pull/644)), so reverting back to that value to reduce costs.

Pairs with changes from https://github.com/pytorch/test-infra/pull/6684/


zxiiro (approved):

jeanschmidt (approved):

jeanschmidt:
we use the same values internally, and seems safe, so, I am merging and releasing this change

PR #340, opened 2025-06-23
Bump configure-aws-credentials to v4.2.1

zxiiro:
We're getting warnings regarding set-output being deprecated in GitHub so bump to a version of configure-aws-credentials that resolves that.

Reference https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Also update all actions to use sha1sum as is best practice.


jordanconway (approved):
Looks good, but I guess we have to wait for https://github.com/pytorch/ci-infra/pull/341 to merge

PR #343, opened 2025-06-24
Backup 2024 ARC attempt

zxiiro:
Rename the folder to arc-backup-2024 so that we can still reference it while making room for the new multicloud WG attempt at ARC. Also remove the old workflows related to the old ARC deployment.

Relates to pytorch-fdn/multicloud-ci-infra#15


zxiiro:
I have a follow up PR that bootstraps the Terraform for the new ARC folder. Will push once this is agreed on.

ZainRizvi (approved):
Looks reasonable to me. Do you want to keep the old workflows around somewhere else as a reference as well?

zxiiro:
> Looks reasonable to me. Do you want to keep the old workflows around somewhere else as a reference as well?

No I'm going to bootstrap the new ARC using the ALI model so we won't be needing all the same jobs. I'm thinking copy the VPC & S3 state bucket as a good bootstrapping point.

PR #344, opened 2025-06-25
Bootstrap new ARC Terraform

zxiiro:
This bootstraps a new Terraform setup for the multicloud WG ARC work. This reuses some of the configuration from the ALI folder to bootstrap the  S3 state bucket.

Relates to pytorch-fdn/multicloud-ci-infra#15


jordanconway (approved):

seemethere (approved):

arc/modules/backend-file/backend.tf
```diff
@@ -0,0 +1,8 @@
+terraform {
+  backend "s3" {
+    bucket         = "tfstate-pyt-arc-prod"
+    key            = "runners/terraform.tfstate"
+    region         = "#AWS_REGION"
+    dynamodb_table = "tfstate-lock-pyt-arc-prod"
```
afrittoli, line 6: If we add more clouds in future, we would probably store their terraform state in a storage on that cloud (to avoid mixing credentials of multiple clouds), but I guess for thise we can create dedicated backend modules e.g. `backend-file-gcp` etc.

.github/workflows/arc-on-pr.yml
```diff
@@ -0,0 +1,62 @@
+name: TFLint & Plan - ARC
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/arc-*'
+      - 'arc/**'
+      - 'modules/**'
+      - 'scripts/**'
```
afrittoli, line 9: What's in scripts that we need here?

.github/workflows/arc-on-pr.yml
```diff
@@ -0,0 +1,62 @@
...
+        terraform_version: 1.5.7
+        terraform_wrapper: false
+
+    - uses: terraform-linters/setup-tflint@444635365d380c7363d1eaee4267317c2445717d # v2.0.1
+      name: Setup TFLint
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Install virtualenv
+      run: pip install virtualenv
```
afrittoli, line 39: What is this for? Also we may need something like
```
echo PATH=${GITHUB_WORKSPACE}/.venv/bin:$PATH >> $GITHUB_ENV
```
to activate the virtual environment for following steps

afrittoli (commented):
Thanks! It looks good to me

PR #345, opened 2025-06-26
Add Network for MultiCloud ARC

zxiiro:
This adds the networking components necessary for us to deploy an EKS cluster.

Relates to pytorch-fdn/multicloud-ci-infra#17.

### DIFF
--- a/arc/aws/391835788720/us-east-1/main.tf
+++ b/arc/aws/391835788720/us-east-1/main.tf
@@ -6,7 +6,7 @@ terraform {
       source  = "hashicorp/random"
     }
     aws    = {
-      version = ">= 6.0"
+      version = ">= 5.95, < 6.0"
       source  = "hashicorp/aws"
     }
   }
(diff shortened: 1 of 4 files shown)


arc/aws/391835788720/us-east-1/vpc.tf
```diff
@@ -0,0 +1,16 @@
+module "arc_runners_vpc" {
+  source = "terraform-aws-modules/vpc/aws"
```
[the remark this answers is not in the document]
zxiiro: Locked the version to:

```
      version = ">= 5.95, < 6.0"
```

arc/aws/391835788720/us-east-1/main.tf
```diff
@@ -6,7 +6,7 @@ terraform {
       source  = "hashicorp/random"
     }
     aws    = {
-      version = ">= 6.0"
```
malfet, line 9: This feels very unrelated to PR title. Why is this needed? Why do we want to downgrade AWS version?

zxiiro: Yes we need to downgrade as I discovered 6.0 was only released last week and many Terraform modules do not work with it yet. My change makes it match the requirements of the EKS module which is what we want to deploy.

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/versions.tf

This is a low risk change since this Terraform stack is new. Nothing production uses it yet.

ZainRizvi (approved):

PR #350, opened 2025-07-28
Set min-available runners to 0

zxiiro:
Considering there is now a scale-up cron job that periodically checks Clickhouse if there are jobs that have been queued for a long time and automatically scales up new runners. We should be able to set min-available to 0 again.

The cron checks this Clickhouse URL
https://hud.pytorch.org/api/clickhouse/queued_jobs_aggregate?parameters=%5B%5D


zxiiro:
A couple of weeks ago @ZainRizvi mentioned that this should work now due to the new cron job. Let's give it a try again and see if using min-runners 0 works.

jordanconway (approved):

PR #361, opened 2025-09-02
WIP Layered terraform infrastructure for ARC

afrittoli:
Setup a layered terraform infrastructure for ARC, with three layers: infra, helm and argocd.

Each layer has a dedicated terraform state, using a dedicated key within the same bucket, to avoid bloating of the terraform state. They all share the same dynamodb table for the lock.

The helm layer takes the inputs required for the helm provider configuration from the infra layer remote state, and the argocd layer similarly takes the inputs from the helm provider outputs in the remote state.

Make files are used to orchestrate deployment of the layered infrastructure. The root make file provides the usual targets that loop through the layers in order, so that one layer may complete provisioning before the next one is started.

The available make targets are the same of the previous single layered setup (init, plan, apply, destroy, clean, tflint), so that the multilayer set up is transparent to CI jobs and do not alter the end-user interface.

Individual layer's makefile import a share makefile and only implement what's different in each layer. The shared make file is almost identical to the existing makefile, with only a few differences required to manage the layer name in the backend bucket key.

The infrastructure is split in layers as follows:
- infra: aws IAM, vpc, eks
- helm: ARC control plane, argocd control plane
- argocd: ArgoCD app for the RunnerSets. The runner sets definitions will be maintained in git and ArgoCD will update them automatically to the cluster. If we have a staging environment, we could use ArgoCD with branches to promote changes from staging to prod via PR.

The helm and argocd layers are not implemented for now, as the PR is already large as it is. If the team agress on this approach to manage the infrastructure for ARC, I will make follow-up PRs to add the various missing part one by one.


afrittoli:
TF lint and plan fail because I cannot access credentials from a fork PR.

zxiiro:
Is the plan here to have separate GitHub Actions jobs that run for the 3 different layers: infra, helm, and argocd?

zxiiro:
> helm: ARC control plane, argocd control plane

I think if we're using ArgoCD then we don't need helm to deploy ARC, we would just bootstrap ArgoCD with Helm, and then use ArgoCD to deploy everything.

PR #362, opened 2025-09-02
Layered terraform infrastructure for ARC

afrittoli:
Setup a layered terraform infrastructure for ARC, with three layers: infra, helm and argocd.

Each layer has a dedicated terraform state, using a dedicated key within the same bucket, to avoid bloating of the terraform state. They all share the same dynamodb table for the lock.

The helm layer takes the inputs required for the helm provider configuration from the infra layer remote state, and the argocd layer similarly takes the inputs from the helm provider outputs in the remote state.

Make files are used to orchestrate deployment of the layered infrastructure. The root make file provides the usual targets that loop through the layers in order, so that one layer may complete provisioning before the next one is started.

The available make targets are the same of the previous single layered setup (init, plan, apply, destroy, clean, tflint), so that the multilayer set up is transparent to CI jobs and do not alter the end-user interface.

Individual layer's makefile import a share makefile and only implement what's different in each layer. The shared make file is almost identical to the existing makefile, with only a few differences required to manage the layer name in the backend bucket key.

The infrastructure is split in layers as follows:
- infra: aws IAM, vpc, eks
- helm: ARC control plane, argocd control plane
- argocd: ArgoCD app for the RunnerSets. The runner sets definitions will be maintained in git and ArgoCD will update them automatically to the cluster. If we have a staging environment, we could use ArgoCD with branches to promote changes from staging to prod via PR.

The helm and argocd layers are not implemented for now, as the PR is already large as it is. If the team agress on this approach to manage the infrastructure for ARC, I will make follow-up PRs to add the various missing part one by one.


arc/aws/391835788720/us-east-1/03_argocd/Makefile
```diff
@@ -0,0 +1,9 @@
+# Include shared logic
+include ../../../../makefiles/layer-common.mk
+
+# Override remote-state.tf to depend on infra layer
+.PHONY: remote-state.tf
+remote-state.tf:
+	@echo "Generating remote state data sources for $(LAYER)..."
+	@> remote-state.tf  # Clear file first
+	@$(call generate-remote-state,helm,runners)
```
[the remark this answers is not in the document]
afrittoli: I refactored this.

afrittoli:
@zxiiro, any more thoughts on this?

arc/templates/backend-file/remote-state.tf
```diff
@@ -0,0 +1,17 @@
+# Handle the case of "terraform plan" on all layers beforce the
```
zxiiro, line 1:
```suggestion
# Handle the case of "terraform plan" on all layers before the
```

zxiiro (approved):
I found a typo but otherwise I think I'm fine with this approach.

arc/templates/backend-file/remote-state.tf
afrittoli: Fixed, thank you!

PR #365, opened 2025-09-10
Add cert-manager and ingress-nginx to ARC

afrittoli:
Add cert-manager to the k8s cluster used for ARC.
Add a ClusterIssuer. The email address required for that is pulled from a GitHub secret called CERT_MANAGER_EMAIL that needs to be created. GitHub workflows have been updated accordingly.

Add ingress-nginx to the k8s cluster used for ARC. This ingress works nicely with cert-manager, ArgoCD and the AWS LB infra.

Configure ArgoCD to use an ingress with the domain name argocd.pytorch.org, annotated for cert-manager and nginx. ArgoCD prefers SSL passthrough, because it transports HTTPs and grpc on the same port. Using SSL passthrough also enhances security as the traffic stays encrypted within the cluster as well.

Using SSL passthrough is a bit of an headache for the HTTP01 challenge. Adding a dedicated ingress template to the cluster issuer should solve that be creating a dedicated route in the ingress backed by the challenge pod and with SSL pass-through disabled.

The DNS name needs to be provisioned manually, so the ingress won't have a valid certificate initially, since Let's Encrypt challange will fail. Once the DNS record is provisioned, the certificate will be created.

Order of execution (after the PR is reviewed):
- merge this PR
- check the public IP address associated with the new ingress
- provision the DNS record
- verify that the certificate is provisioned correctly

### DIFF
--- a/arc/aws/391835788720/us-east-1/02_helm/variables.tf
+++ b/arc/aws/391835788720/us-east-1/02_helm/variables.tf
@@ -9,3 +9,21 @@ variable "argocd_namespace" {
   type        = string
   default     = "argocd"
 }
+
+variable "cert_manager_email" {
+  type        = string
+  description = "Email to be used for the cert-manager cluster issuer"
+  default     = "hostmaster+pytorch-ci-argocd@linuxfoundation.org"
+}
+
+variable "letsencrypt_issuer" {
+  type        = string
+  description = "Name of the cert-manager cluster issuer"
+  default     = "letsencrypt-prod"
+}
+
+variable "argocd_ingress_host" {
+  type        = string
+  description = "Public ArgoCD endpoint"
+  default     = "argocd.pytorch.org"
+}
--- a/arc/aws/391835788720/us-east-1/02_helm/argocd.tf
+++ b/arc/aws/391835788720/us-east-1/02_helm/argocd.tf
@@ -10,8 +10,13 @@ resource "helm_release" "argocd" {
 
   values = [
     # Nothing to pass to the template for now
-    templatefile("${path.module}/values/argocd.yaml.tftpl", {})
+    templatefile("${path.module}/values/argocd.yaml.tftpl", {
+      ingress_host       = var.argocd_ingress_host,
+      letsencrypt_issuer = var.letsencrypt_issuer,
+    })
   ]
+
+  depends_on = [null_resource.k8s_infra_ready]
 }
 
 # Verify that the service exists with expected name and namespace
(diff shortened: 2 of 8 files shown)


.github/workflows/arc-deploy-prod.yml
```diff
@@ -40,3 +40,4 @@ jobs:
         AWS_DEFAULT_REGION: us-east-1
         GITHUB_TOKEN: ${{ secrets.LIST_PYTORCH_RUNNERS_GITHUB_TOKEN }}
         TERRAFORM_EXTRAS: -auto-approve -lock-timeout=15m
+        TF_VAR_cert_manager_email: ${{ secrets.CERT_MANAGER_EMAIL }}
```
afrittoli, line 43: This secret needs to be created on GitHub and include the email address used for cert-manager `ClusterIssuer`. That email address is sent to Let's Encrypt when the certificate is requested, but it's not associated with an account, nor is it stored in the certificate itself. 

Let's Encrypt used to send notification emails related to certificate expirations, but it no longer does.
The cert-manager component on the cluster takes care of renewing the certificate when needed.

afrittoli: If you don't mind storing the email address in git, it can be added as a default value to `variables.tf` and then this environment variable and the need for a secret can be removed

arc/aws/391835788720/us-east-1/02_helm/variables.tf
```diff
@@ -9,3 +9,20 @@ variable "argocd_namespace" {
...
+variable "letsencrypt_issuer" {
+  type        = string
+  description = "Name of the cert-manager cluster issuer"
+  default     = "letsencrypt-prod"
+}
+
+variable "argocd_ingress_host" {
+  type        = string
+  description = "Public ArgoCD endpoint"
+  default     = "argocd.pytorch.org"
```
afrittoli, line 27: @zxiiro This is the name that will need a new DNS record. Once this is merged and applied, I can get the IP that should go in the record. Please let me know if the name is ok - it can also be changed later but it's easier if we stick to the same name once configured.

zxiiro: Name looks fine to me. Let's go with this one.

.github/workflows/arc-deploy-prod.yml
zxiiro: I think it should be okay to store it in Git. Emails aren't really "secret" anyway. You can use this email address `hostmaster+pytorch-ci-argocd@linuxfoundation.org`.

@jordanconway and I should get any emails that come in on that if we need to check.

zxiiro (approved):

PR #378, opened 2025-09-22
Update cloud-account-access for LFID auth

zxiiro:
We are now using IAM Identity Center for authentication into PyTorch AWS Account. This updates the instructions on how to access PyTorch AWS using the LFID login portal for AWS.

Update the pytorch-infra-admins role to use PowerUserAccess as it is effectively administrator access minus the ability to manage users / groups. LFIT staff will handle account access moving forward as it hooks into LFID SSO so needs a different level of account access to manage.

Also remove the pytorch-infra-contributor access role. We currently do not have any members of this group and this group should be more restricted than is defined. PowerUserAccess essentially grants admin privileges minus the ability to manage users / groups.

### DIFF
--- a/cloud-account-access.md
+++ b/cloud-account-access.md
@@ -16,28 +16,13 @@ Infrastructure weekly meeting and are able to administer the account.
 Policy:
 
 * Administrator level permissions will be granted
-  (arn:aws:iam::aws:policy/AdministratorAccess).
+  (arn:aws:iam::aws:policy/PowerUserAccess).
 * Permissions automatically expire after 6 months if not renewed.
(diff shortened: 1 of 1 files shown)


zxiiro:
This is my attempt to document the new access method using LFID SSO to gain access to AWS Console. I have not rolled this out broadly yet.

At some point we'd like to stop using AWS local accounts and grant access to folks using LFID SSO (one less account for folks to remember). I will slowly reach out to folks to migrate their AWS accounts to using LFID accounts as we test and make sure it works well.

jordanconway (approved):

PR #431, opened 2026-04-08
Add release runner class to naming convention

jeanschmidt:
- Add `rel` optional prefix to runner label format
- Document `rel` field: dedicated runner group and node isolation
- Add two release runner examples (arm64g4, x86iavx512)

The `rel-` segment sits between provider and OS in the label, enabling dedicated `release-runners` runner group with `osdc.io/runner-class: release` node isolation — separating release builds from CI workloads.


huydhn (approved):
LGTM!

PR #404, opened 2026-03-19
Add gharts to the arc dev cluster

afrittoli:
Deploys the GitHub Actions Runner Token Service (gharts) to the arc dev cluster in us-east-1, using RDS PostgreSQL with IAM authentication via IRSA.

- 02_helm: gharts Helm release (v0.0.3) using OCI chart from ghcr.io, GitHub App credentials reused from ARC secrets, OIDC via Linux Foundation SSO, ingress at gharts.pytorch.org

PR #444, opened 2026-04-14
Add gharts RDS instance and IRSA role to 01_infra

afrittoli:
Provisions the AWS infrastructure required by the gharts service on the arc dev cluster.

- RDS db.t3.micro PostgreSQL instance with IAM authentication enabled
- Security group allowing PostgreSQL access from within the VPC
- DB subnet group covering private subnets
- IAM role and rds-db:connect policy for IRSA
- Outputs: `gharts_rds_host`, `gharts_irsa_role_arn` (consumed by 02_helm)

## Stack

1. **This PR** — 01_infra: RDS instance, security group, subnet group, IRSA role and policy, outputs
2. **PR #404** — 02_helm: gharts Helm release (depends on this PR)

### DIFF
--- a/arc/aws/391835788720/us-east-1/01_infra/gharts-rds.tf
+++ b/arc/aws/391835788720/us-east-1/01_infra/gharts-rds.tf
@@ -1,0 +1,120 @@
+/*
+ * RDS PostgreSQL instance for gharts (GitHub Actions Runner Token Service).
+ *
+ * Placed in the private subnets of the cluster VPC. Access is restricted to
+ * within the VPC CIDR so only pods running on the cluster can connect.
+ *
+ * Authentication uses IAM RDS auth — no static password is required.
+ * The gharts service account is bound to an IAM role (IRSA) that carries the
+ * rds-db:connect permission.
+ */
+
+# Subnet group covering all private subnets in the VPC
+resource "aws_db_subnet_group" "gharts" {
+  name       = "${var.arc_dev_environment}-gharts"
+  subnet_ids = module.arc_runners_vpc.private_subnets
+
+  tags = {
+    Environment = var.arc_dev_environment
+    Project     = var.arc_dev_environment
+  }
+}
+
+# Allow inbound PostgreSQL from within the VPC (EKS nodes and pods)
+resource "aws_security_group" "gharts_rds" {
+  name        = "${var.arc_dev_environment}-gharts-rds"
+  description = "Allow PostgreSQL access from EKS nodes"
+  vpc_id      = module.arc_runners_vpc.vpc_id
+
+  ingress {
+    description = "PostgreSQL from VPC"
+    from_port   = 5432
+    to_port     = 5432
+    protocol    = "tcp"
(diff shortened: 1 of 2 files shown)


arc/aws/391835788720/us-east-1/01_infra/gharts-rds.tf
```diff
@@ -0,0 +1,120 @@
...
+  }
+}
+
+resource "aws_db_instance" "gharts" {
+  identifier        = "${var.arc_dev_environment}-gharts"
+  engine            = "postgres"
+  engine_version    = "16"
+  instance_class    = "db.t3.micro"
+  allocated_storage = 20
+  storage_type      = "gp2"
```
zxiiro: Should we use `gp3` instead? I think that's the newer one and if I understand it correctly it's cheaper too.

https://docs.aws.amazon.com/emr/latest/ManagementGuide/emr-plan-storage-compare-volume-types.html

arc/aws/391835788720/us-east-1/01_infra/gharts-rds.tf
```diff
@@ -0,0 +1,120 @@
...
+  manage_master_user_password         = true
+
+  db_subnet_group_name   = aws_db_subnet_group.gharts.name
+  vpc_security_group_ids = [aws_security_group.gharts_rds.id]
+
+  # Dev instance — no Multi-AZ, no automated backups needed
+  multi_az                = false
+  backup_retention_period = 0
+  skip_final_snapshot     = true
+  deletion_protection     = false
```
zxiiro: I think this is fine for a dev system but when we create a prod system we should probably set a backup_retention_period and deletion_protection to true.

arc/aws/391835788720/us-east-1/01_infra/gharts-rds.tf
```diff
@@ -0,0 +1,120 @@
...
+  tags = {
+    Environment = var.arc_dev_environment
+    Project     = var.arc_dev_environment
+  }
+}
+
+resource "aws_db_instance" "gharts" {
+  identifier        = "${var.arc_dev_environment}-gharts"
+  engine            = "postgres"
+  engine_version    = "16"
```
zxiiro: Do we want to pin the minor version too? I think Postgres sometimes creates breaking changes in minor releases if I'm not wrong.

zxiiro (approved):
Looks good. I left a few review comments but I think we could proceed without them too. The gp3 one I think we should switch to though.

arc/aws/391835788720/us-east-1/01_infra/gharts-rds.tf
afrittoli: Definitely, will update

arc/aws/391835788720/us-east-1/01_infra/gharts-rds.tf
afrittoli: I added deletion protection and backup retention of 1 day, even in dev it would be nice to preserve the data.

PR #448, opened 2026-04-15
Enable EKS provisioned control plane (XL tier) for production

huydhn:
The standard EKS control plane is constantly throttling the production
cluster.  Analysis of 7-day Mimir metrics (2026-04-15) showed:

  - POST pods requests throttled at ~0.4 req/s (0.36% of traffic,
    ~34K rejected requests/day) — 429s present in 100% of samples
  - Peak API rate: 638 req/s (p99: 478, p95: 321, avg: 130)
  - Cluster scale: up to 171 nodes / 4,220 pods / 223 node creations/hr

The XL provisioned tier ($1.75/hr, ~$15K/yr) provides:

  - 2,000 API concurrency seats (10x standard ~200 seats)
  - 167 pods/s scheduling rate
  - 99.99% SLA (up from 99.95%)

4XL ($6.90/hr, ~$61K/yr) was considered but ruled out — the current
concurrency deficit is a few dozen seats, not thousands.  Pod scheduling
rate caps at 400/s for both 4XL and 8XL, identical to 2XL, so the extra
$46K/yr buys no practical benefit at current scale.

Implementation wires a new `control_plane_scaling_tier` variable from
clusters.yaml through cluster-config.py and the tofu module chain.
Uses a dynamic block so clusters on "standard" (the default) emit no
control_plane_scaling_config block, keeping existing behavior unchanged.

### DIFF
--- a/osdc/clusters.yaml
+++ b/osdc/clusters.yaml
@@ -166,6 +166,15 @@ clusters:
         - ossci_gha_terraform
     base:
       vpc_cidr: "10.1.0.0/16"
+      # Provisioned control plane XL ($1.75/hr, ~$15K/yr).
+      # Analysis (2026-04-15, 7-day window) showed the standard tier constantly
+      # throttling POST-pods requests (~0.4 req/s of 429s, 0.36% of traffic,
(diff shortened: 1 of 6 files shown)


huydhn:
* (to be filled)

PR #454, opened 2026-04-16
Add `just clean` recipe

jeanschmidt:
- Remove tofu init caches (.terraform), lock files, and plan files
- Remove Python bytecode, pytest/ruff caches, coverage artifacts, and .venv
- Remove generated module outputs and integration test scratch dirs
- Operate on both consumer (ROOT) and upstream (UPSTREAM) trees
- Report item count and suggest `just setup` to re-initialize


huydhn (approved):
LGTM!

PR #455, opened 2026-04-16
Add zombie-cleanup module to garbage-collect stuck ARC runner pods

jeanschmidt:
**Impact:** ARC runner clusters (arc-cbr-production, upstream staging/production)
**Risk:** low

## What
Adds a new `zombie-cleanup` OSDC module that runs as a Kubernetes CronJob every hour, detecting and deleting runner pods that are stuck Pending or Running beyond configurable age thresholds. 

It removes pending pods older than 12h and running pods older than 24h. This prevents zombie runner pods to start when the github job have long been canceled and prevents workflow pods to stick wasting resources when the runner pod failed and didn't clean them up properly.

## Why
ARC runner pods can become zombies — stuck in Pending (e.g. unschedulable due to resource pressure or node failures) or workflow pods can stay Running indefinitely (e.g. a job that never terminates). These orphaned pods waste node capacity, block Karpenter from scaling down nodes, and can accumulate over time. There was no automated mechanism to clean them up.

## How
- The cleanup script uses `lightkube` to list all pods in the `arc-runners` namespace, filtering by age and phase
- Controller-managed pods (owned by ReplicaSet, DaemonSet, StatefulSet, Job) are explicitly skipped so long-lived infrastructure like ARC listener pods and hooks-warmer DaemonSets are never touched
- Pods already being terminated (have `deletionTimestamp`) are also skipped to avoid redundant delete calls
- Pods in `Unknown` phase (typically from node failures) use the running threshold
- 404 on delete is treated as success (pod may have been cleaned up between list and delete)
- Content-addressed image tagging: the deploy script hashes source files to compute the image tag, skipping rebuild when nothing changed
- Image is built locally and pushed to Harbor (not ECR), consistent with how other in-cluster tools are deployed

## Changes
- **`modules/zombie-cleanup/scripts/python/zombie_cleanup.py`** — Python script: lists pods, identifies zombies by phase+age, deletes them. Configurable via env vars (`TARGET_NAMESPACE`, `PENDING_MAX_AGE_HOURS`, `RUNNING_MAX_AGE_HOURS`, `DRY_RUN`)
- **`modules/zombie-cleanup/scripts/python/test_zombie_cleanup.py`** — 431-line test suite covering all helper functions, edge cases (naive timestamps, missing timestamps, 404 races, partial failures), and the `main()` entrypoint
- **`modules/zombie-cleanup/kubernetes/cronjob.yaml`** — CronJob running hourly on base-infra nodes with hardened security context (non-root, read-only rootfs, all caps dropped, seccomp)
- **`modules/zombie-cleanup/kubernetes/rbac.yaml`** — Namespace-scoped ServiceAccount + Role + RoleBinding granting only `get`, `list`, `delete` on pods in `arc-runners`
- **`modules/zombie-cleanup/docker/Dockerfile`** — Minimal Alpine-based image using uv for dependency installation
- **`modules/zombie-cleanup/docker/pyproject.toml`** — Single dependency: `lightkube==0.15.5`
- **`modules/zombie-cleanup/deploy.sh`** — Build script: checks enabled flag, computes content-hash tag, builds/pushes to Harbor, applies RBAC + CronJob with sed-substituted config values
- **`clusters.yaml`** — Adds `zombie-cleanup` to module list for `arc-cbr-production`; adds default config (`pending_max_age_hours: 24`, `running_max_age_hours: 12`, `dry_run: false`)

## Notes
- The module is enabled by default in `clusters.yaml` defaults with `dry_run: false`. Set `zombie_cleanup.dry_run: true` per-cluster to observe before enabling deletion.
- The CronJob has `backoffLimit: 0` and `activeDeadlineSeconds: 300` — a single failed run won't retry and won't block the next hourly execution.
- RBAC is namespace-scoped to `arc-runners` only — the cleanup pod cannot touch pods in other namespaces.

### DIFF
--- a/osdc/modules/zombie-cleanup/deploy.sh
+++ b/osdc/modules/zombie-cleanup/deploy.sh
@@ -47,6 +47,7 @@ fi
 PENDING_MAX_AGE=$(uv run "$CFG" "$CLUSTER" zombie_cleanup.pending_max_age_hours "24")
 RUNNING_MAX_AGE=$(uv run "$CFG" "$CLUSTER" zombie_cleanup.running_max_age_hours "12")
 DRY_RUN=$(uv run "$CFG" "$CLUSTER" zombie_cleanup.dry_run "false")
+PUSHGATEWAY_URL=$(uv run "$CFG" "$CLUSTER" zombie_cleanup.pushgateway_url "http://prometheus-pushgateway.monitoring:9091")
 
 # --- Compute content-based image tag ---
 TAG=$(find "$MODULE_DIR/docker" "$MODULE_DIR/scripts/python" \
@@ -138,6 +139,7 @@ sed \
   -e "s|PENDING_MAX_AGE_PLACEHOLDER|${PENDING_MAX_AGE}|" \
   -e "s|RUNNING_MAX_AGE_PLACEHOLDER|${RUNNING_MAX_AGE}|" \
   -e "s|DRY_RUN_PLACEHOLDER|${DRY_RUN}|" \
+  -e "s|PUSHGATEWAY_URL_PLACEHOLDER|${PUSHGATEWAY_URL}|" \
   "$MODULE_DIR/kubernetes/cronjob.yaml" | kubectl_apply_if_changed -f -
 
 echo "  zombie-cleanup deployed."
--- a/osdc/modules/zombie-cleanup/deploy.sh
+++ b/osdc/modules/zombie-cleanup/deploy.sh
@@ -1,0 +1,143 @@
+#!/usr/bin/env bash
+set -euo pipefail
+#
+# Deploy the zombie-cleanup CronJob.
+# Called by: just deploy-module <cluster> zombie-cleanup
+#
+# Args: $1=cluster-id  $2=cluster-name  $3=region
+#
+# Builds the container image, pushes it to Harbor, and applies the
+# RBAC and CronJob manifests with config values substituted.
+
+CLUSTER="$1"
+_CNAME="$2"  # unused but required by deploy-module interface
+_REGION="$3" # unused but required by deploy-module interface
+MODULE_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="${OSDC_ROOT:-$(cd "$MODULE_DIR/../.." && pwd)}"
+UPSTREAM_ROOT="${OSDC_UPSTREAM:-$REPO_ROOT}"
+
+# shellcheck source=/dev/null
+source "$UPSTREAM_ROOT/scripts/mise-activate.sh"
+# shellcheck source=/dev/null
+source "$UPSTREAM_ROOT/scripts/kubectl-apply.sh"
+
+CFG="$UPSTREAM_ROOT/scripts/cluster-config.py"
+
+# --- Cleanup trap ---
+PF_PID=""
+BUILD_CONTEXT=""
+NETRC_FILE=""
+IMAGE_TAR=""
+cleanup() {
+  [[ -n "$PF_PID" ]] && kill "$PF_PID" 2>/dev/null || true
+  [[ -n "$BUILD_CONTEXT" ]] && rm -rf "$BUILD_CONTEXT" 2>/dev/null || true
+  [[ -n "$NETRC_FILE" ]] && rm -f "$NETRC_FILE" 2>/dev/null || true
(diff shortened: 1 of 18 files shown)


osdc/modules/zombie-cleanup/deploy.sh
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env bash
```
malfet, line 1: I don't know much about Kubernetes, but this file looks like something that could be avoided or replaced with 1 liner

osdc/modules/zombie-cleanup/scripts/python/zombie_cleanup.py
```diff
@@ -0,0 +1,179 @@
...
+        config["pending_max_hours"],
+        config["running_max_hours"],
+        config["dry_run"],
+    )
+
+    try:
+        client = Client()
+    except Exception:
+        log.exception("Failed to create Kubernetes client")
+        return 1
```
malfet: Any reason why one is better than the other?
```suggestion
    client = Client()
```

osdc/modules/zombie-cleanup/scripts/python/zombie_cleanup.py
```diff
@@ -0,0 +1,179 @@
...
+            log.info("No zombie pods found")
+            return 0
+
+        log.info("Found %d zombie pod(s)", len(zombies))
+        deleted, failed = delete_zombies(client, zombies, config)
+        log.info("Cleanup complete: %d deleted, %d failed", deleted, failed)
+        return 1 if failed > 0 else 0
+    except Exception:
+        log.exception("Cleanup failed")
+        return 1
```
malfet: This would make debugging really hard, please remove try/catch block

malfet (approved):

osdc/modules/zombie-cleanup/scripts/python/zombie_cleanup.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+
+    try:
+        client = Client()
+    except Exception:
+        log.exception("Failed to create Kubernetes client")
+        return 1
+
+    try:
+        zombies = find_zombie_pods(client, config)
```
huydhn, line 176: As this module could kill pods by their age, do you want a safety valve here to see how many zombies it finds or the % of zombie pods and check if that number is sane before actually do the deleting?

My thought process here is that if an abnormal number is captured, i.e. 50% of the pods are zombies, do you trust the module to automatically kill half of the pods?

osdc/modules/zombie-cleanup/scripts/python/zombie_cleanup.py
```diff
@@ -0,0 +1,179 @@
...
+        if not zombies:
+            log.info("No zombie pods found")
+            return 0
+
+        log.info("Found %d zombie pod(s)", len(zombies))
```
huydhn: Are these logs going to Grafana? I assume yes, so we can get the logs there to debug issues? For example, to check if a pod is killed as a zombie or exit normally

huydhn (approved):

osdc/modules/zombie-cleanup/deploy.sh
jeanschmidt: Yeahp, this is a explicit technical debt.

I dediced to not go with plumi right away and coordinate deployments with just. So, when we need to do some templating it goes to bash or python scripts. Mostly I try to rely on Python, but a few simple things end up just in bash.

Let me land this one and I'll follow up with replacing them with python, this will be a step stone to move to plumi and have a simpler release script/process

osdc/modules/zombie-cleanup/scripts/python/zombie_cleanup.py
jeanschmidt: excelent finding

osdc/modules/zombie-cleanup/scripts/python/zombie_cleanup.py
jeanschmidt: no, let me add this.

osdc/modules/zombie-cleanup/scripts/python/zombie_cleanup.py
jeanschmidt: Good point, we can put some safety guardrails and alert in case those are violated. At worst intervening manually would force us to be aware if something is causing excesive leackages

jeanschmidt:
the change became much more significant due to adding monitoring

PR #458, opened 2026-04-16
Add node-fleet abstraction for multi-size instance pools

jeanschmidt:
**Impact:** All runner scheduling, NodePool generation, and DaemonSets on runner/GPU nodes
**Risk:** high

## What
Introduces a "node-fleet" layer between runners and instance types so that a single runner definition can schedule across multiple instance sizes within the same family (e.g. r7a.8xlarge through r7a.48xlarge), with Karpenter preferring the largest available size and falling back to smaller ones when capacity is constrained.

## Why
Previously each runner was pinned 1:1 to a specific instance type via `instance-type` nodeSelector and taint. If that exact instance size was unavailable (capacity limits, spot exhaustion, regional stock-outs), runner pods were stuck pending with no fallback. This change gives Karpenter a ranked list of sizes per family so it can provision the best available node, improving scheduling resilience without changing runner resource requests.

## How
- A new `node-fleet` label and taint replaces `instance-type` as the primary scheduling constraint for runners. The `instance-type` taint is kept (with `operator: Exists`) so nodes remain isolated, but runners no longer require a specific size.
- Karpenter `weight` on each NodePool within a fleet controls preference order: largest/most-efficient sizes get the highest weight and are tried first.
- GPU families use a fleets: (plural) format that groups by GPU count (e.g. g5-1gpu, g5-4gpu, g5-8gpu), so Karpenter provisions instances with the right GPU count rather than wasting GPUs on oversized nodes.
- Legacy `nodepool:` defs auto-derive a fleet name so they get the new label/taint without manual migration.
- Fleet name derivation uses actual GPU count from `INSTANCE_SPECS` (not runner-requested GPUs), since multiple runner sizes share the same multi-GPU node.

### How runners and nodepools connect

Runners no longer target a specific instance type. They target a **fleet** — a group of instance sizes within the same family that Karpenter can choose from based on availability.

**Matching rule**: runner pod gets `nodeSelector: { node-fleet: "<fleet>" }` + a toleration for that fleet's taint. The fleet name is derived automatically from the runner's instance type using `instance_specs.py`:
- CPU instances: family name → `c7i`
- GPU instances: family + GPU count → `g5-4gpu`

### Defining a fleet (nodepools)

One YAML per instance family instead of one per size. Each size gets a `weight` — Karpenter prefers higher weights, falls back to lower ones.

**CPU example** (`c7i.yaml`):
```yaml
fleet:
  name: c7i
  arch: amd64
  gpu: false
  instances:
    - type: c7i.48xlarge   # preferred
      weight: 100
      node_disk_size: 3750
    - type: c7i.8xlarge    # fallback
      weight: 20
      node_disk_size: 650
```

**GPU example** (`g5.yaml`) — uses `fleets:` (plural) to split by GPU count:
```yaml
fleets:
  - name: g5-1gpu
    gpu: true
    instances:
      - type: g5.8xlarge
        weight: 100
  - name: g5-4gpu
    gpu: true
    instances:
      - type: g5.12xlarge
        weight: 100
```

### What gets generated

Each instance in a fleet produces its own NodePool + EC2NodeClass, all sharing:
- Label: `node-fleet: c7i`
- Taint: `node-fleet=c7i:NoSchedule`

Runner pods tolerate and select by fleet, so they land on whichever size Karpenter provisions.

### DaemonSets

Base DaemonSets (git-cache, nvidia-device-plugin, etc.) gained `node-fleet` tolerations so they still schedule on fleet-managed nodes.

#### The Three-Layer Chain

Three pieces work together: **Karpenter NodePool**, **Node Taints**, and **Runner Pod Tolerations + NodeSelector**.

#### 1. NodePool defines what nodes to create and taints them

When a NodePool is created from a def file like `defs/c7i.yaml`:

```yaml
fleet:
  name: c7i
  arch: amd64
  instances:
    - type: c7i.48xlarge
```

The generator produces a NodePool that **labels** and **taints** every node it creates:

```yaml
# Labels on the node (for nodeSelector matching)
labels:
  workload-type: github-runner
  node-fleet: "c7i"

### Taints on the node (repel everything by default)
taints:
  - key: node-fleet
    value: "c7i"
    effect: NoSchedule        # "don't schedule here unless you tolerate me"
  - key: instance-type
    value: "c7i.48xlarge"
    effect: NoSchedule
```

Think of taints as a **"keep out" sign**. No pod lands on these nodes unless it explicitly says "I'm okay with this."

#### 2. Runner pod says "I belong on that fleet"

A runner def like `defs/l-x86iavx512-8-64.yaml`:

```yaml
runner:
  name: l-x86iavx512-8-64
  instance_type: r7a.48xlarge
  vcpu: 8
  memory: 64Gi
```

Gets wired into the pod template with two mechanisms:

```yaml
# nodeSelector = "only schedule me on nodes with THESE labels"
nodeSelector:
  workload-type: github-runner
  node-fleet: "r7a"              # derived from instance_type family

# tolerations = "I'm okay with these taints"
tolerations:
  - key: node-fleet
    operator: Equal
    value: "r7a"                 # matches the exact fleet taint
    effect: NoSchedule
  - key: instance-type
    operator: Exists             # tolerates ANY instance-type value
    effect: NoSchedule
  - key: git-cache-not-ready     # startup taint (removed when cache is warm)
    operator: Equal
    value: "true"
    effect: NoSchedule
```

#### 3. Karpenter sees the pending pod and provisions the right node

```
GitHub job queued with `runs-on: linux.r7a.8xlarge.64gb`
         │
         ▼
ARC creates a runner pod with:
  nodeSelector: { node-fleet: "r7a" }
  tolerations:  [ node-fleet=r7a, instance-type=Exists ]
  resources:    { cpu: 8, memory: 64Gi }
         │
         ▼
Pod is Pending (no matching node exists yet)
         │
         ▼
Karpenter sees the pending pod, checks which NodePool can satisfy it:
  - nodeSelector "node-fleet=r7a" → only NodePool "r7a" has this label
  - tolerations match r7a's taints → ✓
  - resources fit on r7a instances → ✓
         │
         ▼
Karpenter launches an r7a.48xlarge EC2 instance
  → node gets labels: node-fleet=r7a, workload-type=github-runner
  → node gets taints: node-fleet=r7a, instance-type=r7a.48xlarge
         │
         ▼
Pod scheduled onto the new node (nodeSelector matches, tolerations match)
```

### New abstraction: fleet definition format
- New `fleet:` (single fleet) and `fleets:` (multi-fleet) YAML formats in `modules/nodepools/defs/`
- Replaced 22 single-instance def files with 14 fleet def files covering the same instance types plus smaller fallback sizes
- `generate_nodepools.py`: added `_process_fleet`, `_build_fleet_nodepool_def`, `_process_nodepool` to handle all three formats
- Generated NodePools now include `spec.weight`, `node-fleet` label, and `node-fleet` taint when fleet info is present

### Runner scheduling changes
- `runner.yaml.tpl`: nodeSelector changed from `instance-type` to `node-fleet`; tolerations now include both `node-fleet` (exact match) and `instance-type` (Exists); job pod affinity targets `node-fleet` instead of `instance-type`
- `generate_runners.py`: added `derive_fleet_name()` and `{{NODE_FLEET}}` template variable
- Both generators now import from shared `instance_specs.py`

### DaemonSet tolerations
- Added `node-fleet` toleration (operator: Exists) to all DaemonSets that run on runner nodes: git-cache, image-cache-janitor, node-performance-tuning, nvidia-device-plugin, hooks-warmer, cache-enforcer, dcgm-exporter

### Instance specs expansion
- `instance_specs.py`: added ~30 new instance types (fleet fallback sizes) with vCPU, memory, GPU count, and ENI max-pod values

### Tests
- `test_generate_nodepools.py`: new `TestFleetNodepoolGeneration` and `TestBuildFleetNodepoolDef` classes; updated `TestRealDefFiles` to expand fleet defs; added `TestMain` cases for fleet, multi-fleet, release, and mixed formats
- `test_generate_runners.py`: new `TestDeriveFleetName` class; updated template and assertions to use `node-fleet`
- `test_nodepools.py` (smoke): updated `_load_all_defs` to expand fleet/fleets formats

## Notes
- The `instance-type` label on nodes is preserved (for observability and debugging) — only the scheduling constraint moved to `node-fleet`.
- Legacy `nodepool:` format is still supported and auto-derives a fleet name, so consumer repos with custom single-instance defs continue to work without changes.
- Deploying this requires redeploying both nodepools (`just deploy nodepools <env>`) and runners (`just deploy arc-runners <env>`) together — the new `node-fleet` taint on nodes will block runner pods that still use the old `instance-type`-only tolerations.


malfet:
@claude can you propose a review strategy of this PR? I.e. highlight files that change the algorithm, how big are the algorithmic change compared to config/test-harness changes

osdc/modules/arc-runners/scripts/python/generate_runners.py
```diff
@@ -23,6 +23,15 @@
...
 
+# instance_specs lives in scripts/python/ at the repo root.  Add it to
+# sys.path so the import works both when run directly (deploy.sh) and
+# when run via pytest (pyproject.toml testpaths also adds it).
+_scripts_python = str(Path(__file__).resolve().parents[4] / "scripts" / "python")
+if _scripts_python not in sys.path:
+    sys.path.insert(0, _scripts_python)
+
+from instance_specs import INSTANCE_SPECS  # noqa: E402
+
```
malfet, line 34: Frankly, this feels like a bandaid/incorrect code layout

osdc/modules/arc-runners/scripts/python/generate_runners.py
```diff
@@ -42,6 +51,25 @@ def normalize_name(name):
     return name.replace(".", "-").replace("_", "-")
 
 
+def derive_fleet_name(instance_type):
```
malfet, line 54: Is there anything wrong with type annotations when they are straighforward?
```suggestion
def derive_fleet_name(instance_type: str) -> str:
```

malfet (commented):
Do we have to migrate all runners to the new schema simultaneously or can we have mixed fleets? For example `.metal` instances can not probably be replaced, and same is true for something like H100

osdc/modules/arc-runners/scripts/python/generate_runners.py
jeanschmidt: please don't review the code, just the overall strategy and plan. I want initial feedback on this before polishing the code and implementation

PR #463, opened 2026-04-20
Require known instance types in fleet derivation

jeanschmidt:
**Impact:** nodepool and runner generation — config-time validation only
**Risk:** low — fails fast on bad config instead of producing silent wrong output

## What
Make `derive_fleet_name()` and nodepool generation fail explicitly when an instance type is missing from `INSTANCE_SPECS`, instead of silently falling back to a potentially wrong fleet name.

## Why
Previously, an unknown instance type (e.g. a typo like `g5.8xlareg` or a new size not yet in `INSTANCE_SPECS`) would silently produce a fleet name with 0 GPUs or use the runner's requested GPU count. On multi-GPU instances, this could cause a runner requesting 1 GPU to schedule on a node with 8 GPUs (or vice versa), wasting resources or failing at runtime. Failing early surfaces these errors at generation time, before they reach the cluster.

## How
- Removed the `gpu_count` fallback parameter from `derive_fleet_name()` — GPU count is always read from `INSTANCE_SPECS`, never from the runner's requested count
- Added explicit `ValueError` when an instance type is not found in `INSTANCE_SPECS`
- Updated tests to use instance types that exist in `INSTANCE_SPECS`
- Documented the `node-fleet` taint in the scheduling gates doc

## Code examples

**Before** — silent fallback:
```python
derive_fleet_name("g5.8xlareg", gpu_count=1)  # → "g5-1gpu" (wrong, typo not caught)
derive_fleet_name("z9.xlarge")                 # → "z9" (unknown family, no error)
```

**After** — explicit failure:
```python
derive_fleet_name("g5.8xlareg")  # → ValueError: Unknown instance type 'g5.8xlareg'
derive_fleet_name("z9.xlarge")   # → ValueError: Unknown instance type 'z9.xlarge'
derive_fleet_name("g5.8xlarge")  # → "g5-1gpu" (correct, from INSTANCE_SPECS)
```

## Changes
- `generate_runners.py`: removed `gpu_count` parameter from `derive_fleet_name()`, added `ValueError` for unknown types
- `generate_nodepools.py`: added validation that all fleet instance types exist in `INSTANCE_SPECS`
- `test_generate_runners.py`: updated test instance types to known entries, added unknown-type error tests
- `test_generate_nodepools.py`: added tests for unknown instance type rejection
- `docs/node-warmup-and-scheduling-gates.md`: documented `node-fleet` taint in the node initialization sequence

### DIFF
--- a/osdc/docs/node-warmup-and-scheduling-gates.md
+++ b/osdc/docs/node-warmup-and-scheduling-gates.md
@@ -123,6 +123,7 @@ Runs on nodes with `workload-type` in `[github-runner, buildkit]`.
 |-----------|------|--------|-------|------------|
 | `git-cache-not-ready=true` | Startup taint | `NoSchedule` | All Karpenter NodePools | git-cache-warmer DaemonSet |
 | `instance-type={type}` | Permanent | `NoSchedule` | ARC runner NodePools | Never (scheduling constraint) |
+| `node-fleet={fleet}` | Permanent | `NoSchedule` | ARC runner NodePools | Never (fleet-based scheduling) |
 | `workload/buildkit-{arch}=true` | Permanent | `NoSchedule` | BuildKit NodePools | Never (scheduling constraint) |
 | `nvidia.com/gpu=true` | Permanent | `NoSchedule` | GPU NodePools only | Never (scheduling constraint) |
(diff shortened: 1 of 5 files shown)


jeanschmidt:
Superseded by new ghstack stack (#464-#468) with proper descriptions.

PR #467, opened 2026-04-20
Add node-fleet taint toleration to DaemonSets and runners

jeanschmidt:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #486
* #476
* #474
* #485
* #473
* #469
* #468
* __->__ #467
* #466
* #465
* #464

**Impact:** all DaemonSets on runner/GPU nodes + runner pod scheduling
**Risk:** low — adding a toleration for a taint that doesn't exist yet is a no-op; safe to land before or after fleet NodePools

## What
Update all workloads that run on runner nodes to tolerate the new `node-fleet` taint, and update runner scheduling to select nodes by fleet name.

## Why
Fleet NodePools (PRs #460–#461) add a `node-fleet` taint. Without matching tolerations, DaemonSet pods won't schedule on fleet-tainted nodes, and runners won't be able to target the right fleet.

## How
- DaemonSets use `operator: Exists` — they run on *all* fleet nodes regardless of fleet name
- Runner pods use exact-match tolerations and `nodeSelector` — each runner targets a *specific* fleet
- `derive_fleet_name()` in `generate_runners.py` computes the fleet name from the runner's instance type (e.g. `c7i.48xlarge` → `c7i`, `g5.8xlarge` → `g5-1gpu`)
- The runner template injects the fleet name into both `nodeSelector` and tolerations

## Code examples

**DaemonSet toleration** (wildcard — matches any fleet):
```yaml
tolerations:
  - key: "node-fleet"
    operator: "Exists"
    effect: "NoSchedule"
```

**Runner pod scheduling** (targets specific fleet):
```yaml
spec:
  nodeSelector:
    node-fleet: "c7i"
  tolerations:
    - key: "node-fleet"
      value: "c7i"
      operator: "Equal"
      effect: "NoSchedule"
    - key: "instance-type"
      operator: "Exists"
      effect: "NoSchedule"
```

**Fleet name derivation**:
```python
derive_fleet_name("c7i.48xlarge")  # → "c7i"
derive_fleet_name("g5.8xlarge")    # → "g5-1gpu"
derive_fleet_name("g5.48xlarge")   # → "g5-8gpu"
```

## Changes

**Base DaemonSets** (added `node-fleet` toleration):
- `base/kubernetes/git-cache/daemonset.yaml`
- `base/kubernetes/image-cache-janitor/daemonset.yaml`
- `base/kubernetes/node-performance-tuning.yaml`
[...]

PR #468, opened 2026-04-20
Require known instance types in fleet derivation

jeanschmidt:
* #486
* #476
* #474
* #485
* #473
* #469
* __->__ #468
* #467
* #466
* #465
* #464

**Impact:** nodepool and runner generation — config-time validation only
**Risk:** low — fails fast on bad config instead of producing silent wrong output

Make `derive_fleet_name()` and nodepool generation fail explicitly when an instance type is missing from `INSTANCE_SPECS`, instead of silently falling back to a potentially wrong fleet name.

Previously, an unknown instance type (e.g. a typo like `g5.8xlareg` or a new size not yet in `INSTANCE_SPECS`) would silently produce a fleet name with 0 GPUs or use the runner's requested GPU count. On multi-GPU instances, this could cause a runner requesting 1 GPU to schedule on a node with 8 GPUs (or vice versa). Failing early surfaces these errors at generation time.

- Removed the `gpu_count` fallback parameter from `derive_fleet_name()` — GPU count is always read from `INSTANCE_SPECS`
- Added explicit `ValueError` when an instance type is not found
- Updated tests to use instance types that exist in `INSTANCE_SPECS`

**Before** — silent fallback:
```python
derive_fleet_name("g5.8xlareg", gpu_count=1)  # → "g5-1gpu" (wrong, typo not caught)
derive_fleet_name("z9.xlarge")                 # → "z9" (unknown family, no error)
```

**After** — explicit failure:
```python
derive_fleet_name("g5.8xlareg")  # → ValueError: Unknown instance type 'g5.8xlareg'
derive_fleet_name("g5.8xlarge")  # → "g5-1gpu" (correct, from INSTANCE_SPECS)
```

- `generate_runners.py`: removed `gpu_count` parameter from `derive_fleet_name()`, added `ValueError` for unknown types
- `generate_nodepools.py`: added validation that all fleet instance types exist in `INSTANCE_SPECS`
- `test_generate_runners.py`: updated test instance types, added unknown-type error tests
- `test_generate_nodepools.py`: added tests for unknown instance type rejection
- `docs/node-warmup-and-scheduling-gates.md`: documented `node-fleet` taint in node init sequence

### DIFF
--- a/osdc/docs/node-warmup-and-scheduling-gates.md
+++ b/osdc/docs/node-warmup-and-scheduling-gates.md
@@ -123,6 +123,7 @@ Runs on nodes with `workload-type` in `[github-runner, buildkit]`.
 |-----------|------|--------|-------|------------|
 | `git-cache-not-ready=true` | Startup taint | `NoSchedule` | All Karpenter NodePools | git-cache-warmer DaemonSet |
 | `instance-type={type}` | Permanent | `NoSchedule` | ARC runner NodePools | Never (scheduling constraint) |
+| `node-fleet={fleet}` | Permanent | `NoSchedule` | ARC runner NodePools | Never (fleet-based scheduling) |
 | `workload/buildkit-{arch}=true` | Permanent | `NoSchedule` | BuildKit NodePools | Never (scheduling constraint) |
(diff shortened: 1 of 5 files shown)


huydhn (approved):
LGTM!

PR #469, opened 2026-04-20
Add harbor-cache-recovery module to auto-purge corrupted Harbor cache

jeanschmidt:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #486
* #476
* #474
* #485
* #473
* __->__ #469
* #468
* #467
* #466
* #465
* #464

Signed-off-by: Jean Schmidt <contato@jschmidt.me>

PR #473, opened 2026-04-20
Add fleet-aware grouping to node compactor

jeanschmidt:
* #486
* #476
* #474
* #485
* __->__ #473
* #469
* #468
* #467
* #466
* #465
* #464

----

- Group nodes by `node-fleet` label instead of NodePool for capacity
  decisions, cooldown, and metrics; fall back to NodePool when absent
- Add `group_key` callable to `compute_taints()` for late-binding
  group strategy (backward compatible: None = per-NodePool)
- Taint smaller instances first within a fleet (`allocatable_cpu` in
  sort key) so larger, more-packable nodes are preserved
- Re-group by NodePool inside `select_reserved_nodes()` since capacity
  reservations are per-instance-type, not per-fleet
- Build both `fleet_groups` and `pool_groups` in compactor so each
  metric uses the correct grouping dimension
- Add `LABEL_NODE_FLEET` constant as single source of truth; guard
  against empty-string values

Nodes sharing a `node-fleet` label now form a single capacity domain
for bin-packing, spare capacity, and fleet cooldown. This lets the
compactor consolidate across heterogeneous instance types (e.g. drain
all m8g.8xlarge nodes when pods fit on the m8g.48xlarge in the same
fleet). Reservations intentionally stay per-NodePool because they map
to AWS capacity reservation IDs by instance type.

### DIFF
--- a/osdc/base/node-compactor/scripts/python/compactor.py
+++ b/osdc/base/node-compactor/scripts/python/compactor.py
@@ -27,7 +27,7 @@ import time
 import metrics as m
 from discovery import build_node_states, discover_managed_nodes
 from lightkube import ApiError, Client
-from models import Config
+from models import LABEL_NODE_FLEET, Config, NodeState
 from packing import _count_spare_nodes, compute_taints, select_reserved_nodes
 from phantom import apply_pending_phantom_load
 from prometheus_client import start_http_server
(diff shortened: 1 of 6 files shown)


huydhn (approved):
This change would probably be a bit simpler if we don't need to support the legacy `nodepool` syntax. So, let's confirm if that is still needed before landing this

PR #479, opened 2026-04-21
osdc: add `just plan <cluster>` recipe

huydhn:
* #482
* #483
* #481
* #480
* #484
* __->__ #479

Run `tofu plan` for the base terraform plus every terraform-backed
module of a cluster. Read-only — no apply, no kubectl, no helm side
effects. Used by the PR plan workflow to preview prod infra changes
before they merge.

Suppresses `tofu init` output for a cleaner plan dump (stderr is
preserved for errors) and uses `-no-color` / `-input=false` so the
output is safe to capture into a PR comment.

### DIFF
--- a/osdc/justfile
+++ b/osdc/justfile
@@ -498,6 +498,61 @@ deploy-module cluster module:
     deploy_log_finish module "$CLUSTER" "$MODULE" "$DEPLOY_LOG_MOD_START"
     deploy_log_finish cmd "$CLUSTER" "deploy-module-$MODULE" "$DEPLOY_LOG_CMD_START"
 
+# Run `tofu plan` for base + every terraform-backed module of a cluster.
+# Read-only: no apply, no k8s/helm side effects. Safe to run from CI on PRs.
+plan cluster:
+    #!/usr/bin/env bash
+    set -euo pipefail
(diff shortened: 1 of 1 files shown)


osdc/justfile
```diff
@@ -498,6 +498,61 @@ deploy-module cluster module:
     deploy_log_finish module "$CLUSTER" "$MODULE" "$DEPLOY_LOG_MOD_START"
     deploy_log_finish cmd "$CLUSTER" "deploy-module-$MODULE" "$DEPLOY_LOG_CMD_START"
 
+# Run `tofu plan` for base + every terraform-backed module of a cluster.
+# Read-only: no apply, no k8s/helm side effects. Safe to run from CI on PRs.
+plan cluster:
+    #!/usr/bin/env bash
```
seemethere, line 504: Maybe this is a nit on the general code base but why write shell script in a Justfile?

We have the technology to blast as much python as we want and that's easier to test we should just do that instead of relying on primitives like this.

jeanschmidt (approved):

osdc/justfile
huydhn: Let me add a follow-up task to swap justfile + bash to justfile + uv + python.  Agree that it makes more sense that way instead of using bash here

PR #480, opened 2026-04-21
Add reusable OSDC deploy workflow

huydhn:
* #482
* #483
* #481
* __->__ #480
* #484
* #479

Inputs: cluster (required), environment (default ""), taint_nodes (default true), run_integration (default true), skip_lint_test (default false, firefight escape hatch).

Secrets: META_AWS_ACC_ID, META_AWS_DEPLOY_ROLE (caller picks staging vs prod role), optional CANARY_GITHUB_TOKEN for integration.

Jobs:
1. deploy — checkout → just/mise/uv → just lint + just test (skippable) → OIDC AWS creds → just deploy $cluster
2. smoke — needs: deploy, runs just smoke
3. compactor — needs: deploy, runs just test-compactor
4. integration — needs: deploy, if: inputs.run_integration, runs just integration-test with CANARY_GITHUB_TOKEN

All four jobs inherit inputs.environment, so the OIDC sub carries the environment:X qualifier — which is how the staging and prod IAM
roles gate fork PRs / require the osdc-production approval.

### DIFF
--- a/.github/workflows/_osdc-deploy.yml
+++ b/.github/workflows/_osdc-deploy.yml
@@ -22,6 +22,11 @@ on:
         required: false
         type: boolean
         default: true
+      skip_lint_test:
+        description: "Skip `just lint` and `just test` pre-flight checks (firefighting only)"
+        required: false
+        type: boolean
+        default: false
     secrets:
       META_AWS_ACC_ID:
         required: true
@@ -61,9 +66,11 @@ jobs:
         run: uv sync
 
       - name: Lint
+        if: ${{ !inputs.skip_lint_test }}
         run: just lint
 
       - name: Test
+        if: ${{ !inputs.skip_lint_test }}
         run: just test
 
       - name: Configure AWS credentials
--- a/.github/workflows/_osdc-deploy.yml
+++ b/.github/workflows/_osdc-deploy.yml
@@ -16,11 +16,6 @@ on:
         required: false
         type: string
         default: ""
-      ref:
-        description: "Git ref to check out (empty = caller's ref)"
-        required: false
-        type: string
-        default: ""
       run_smoke:
         description: "Run smoke tests as part of `just deploy`"
         required: false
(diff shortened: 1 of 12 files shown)


.github/workflows/_osdc-deploy.yml
```diff
@@ -0,0 +1,70 @@
...
+      role_to_assume:
+        description: "Full ARN of the AWS role to assume via OIDC"
+        required: true
+        type: string
+      environment:
+        description: "GitHub environment for deployment gating (empty = no gate)"
+        required: false
+        type: string
+        default: ""
+      run_smoke:
```
jeanschmidt: It makes more sense for this to be a separate step on the workflow

.github/workflows/_osdc-deploy.yml
```diff
@@ -0,0 +1,70 @@
...
+        run: uv sync
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@b47578312673ae6fa5b5096b330d9fbac3d116df # v4.2.1
+        with:
+          role-to-assume: ${{ inputs.role_to_assume }}
+          aws-region: us-west-1
+          role-duration-seconds: 7200
+
+      - name: Deploy ${{ inputs.cluster }}
```
jeanschmidt, line 83: add a `just integration-test` step, it needs a cross-repo token so we might end up having to set a PAT here, but it can be nice to automatically trigger the integration tests after deployment with the smoke tests to flag anything outside of the usual

jeanschmidt, line 83: gate the release on lint and test, 

either by gating on a workflow or by adding a step here, for firefight reasons this should be skipeable

huydhn: Add all the smoke, node compactor, and integration test (for staging) as part of this workflow so that we can run deplot and test in one unit

huydhn: Done, the deployment now runs just lint and just test first. They are skippable on prod deployment in the case of emergency firelight

jeanschmidt (approved):

PR #483, opened 2026-04-21
Refactor osdc-pre-merge to use reusable deploy, gate on env

huydhn:
* #482
* __->__ #483

The reusable workflow owns the full deploy → smoke → compactor → integration sequence, so this file is intentionally thin — just a trigger, path filter, and one call.


.github/workflows/osdc-pre-merge.yml
```diff
@@ -102,7 +87,7 @@ jobs:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@b47578312673ae6fa5b5096b330d9fbac3d116df # v4.2.1
         with:
-          role-to-assume: arn:aws:iam::${{ secrets.META_AWS_ACC_ID }}:role/${{ secrets.META_AWS_DEPLOY_ROLE }}
```
huydhn, line 105: I don't know why we need to keep this a secret.  We have been using OIDC publicly in PyTorch, i.e. https://github.com/pytorch/pytorch/blob/main/.github/workflows/_linux-build.yml#L528. The protection will come from `environment:` instead where we can have branch protection rule

jeanschmidt: security by obscurity?

huydhn: Sure, we can keep the secrets then

jeanschmidt (approved):

huydhn:
Testing looks ok now https://github.com/pytorch/ci-infra/actions/runs/24749964236/job/72412191637, going to merge this. The compator failure is unrelated, but I can take a look at this in a follow-up PR

PR #487, opened 2026-04-22
Run integration tests after prod deployment

huydhn:
Run integration tests after prod deployment and skip other post-deployment tests that manipulating the clusters

### Testing

https://github.com/pytorch/ci-infra/actions/runs/24802393183

### DIFF
--- a/.github/workflows/_osdc-deploy.yml
+++ b/.github/workflows/_osdc-deploy.yml
@@ -17,6 +17,11 @@ on:
         required: false
         type: boolean
         default: true
+      run_smoke:
+        description: "Run smoke tests after deploy"
+        required: false
+        type: boolean
+        default: true
       run_integration:
         description: "Run integration tests after deploy (requires CANARY_GITHUB_TOKEN)"
         required: false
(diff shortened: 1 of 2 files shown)


jeanschmidt:
In prod we should not run any of the tests.

Compactor can put the cluster in a unhealthy state (potentially), janitor as well.

Only lint, integration and potentially workload tests should run IMO.

.github/workflows/osdc-deploy-prod.yml
```diff
@@ -37,6 +37,10 @@ jobs:
       # Integration tests target pytorch-canary via CANARY_GITHUB_TOKEN —
       # they don't belong in a prod deploy. Always skip.
       run_integration: false
```
jeanschmidt: true here

.github/workflows/_osdc-deploy.yml
jeanschmidt, line 135: add integration tests here we had this, this is important, we should run on staging and on production

.github/workflows/osdc-deploy-prod.yml
huydhn: I can just remove this line, the integration tests are run by default

jeanschmidt (approved):

PR #494, opened 2026-04-23
Move OSDC workflow secrets to GitHub environments 

huydhn:
META_AWS_ACC_ID, META_AWS_DEPLOY_ROLE, and CANARY_GITHUB_TOKEN now live in the osdc-staging and osdc-production environments, so callers no longer need to pipe them through workflow_call. Each job in _osdc-deploy.yml already sets environment:, which makes the env-scoped secrets resolve automatically.

### DIFF
--- a/.github/workflows/_osdc-deploy.yml
+++ b/.github/workflows/_osdc-deploy.yml
@@ -37,15 +37,6 @@ on:
         required: false
         type: boolean
         default: false
-    secrets:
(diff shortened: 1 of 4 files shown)


zxiiro (approved):

malfet (approved):
PR title is misleading. It does not move anything, just deletes them

jeanschmidt (approved):

PR #498, opened 2026-04-23
Apply maxRunners cap to B200 runner defs

huydhn:
* __->__ #498
* #496
* #495

Size each B200 runner scale set to the current reserved capacity
(1 × p6-b200.48xlarge = 8 GPUs):

  l-x86iamx-22-225-b200      (1 GPU):  max_runners: 8
  l-x86iamx-44-450-b200-2    (2 GPUs): max_runners: 4
  l-x86iamx-88-900-b200-4    (4 GPUs): max_runners: 2
  l-bx86iamx-176-1800-b200-8 (8 GPUs): max_runners: 1

The cap gates ARC runner creation so surplus jobs queue server-side at
GitHub's Actions Service instead of running into workflow-pod scheduling
failures when the reserved GPU pool is full.

### DIFF
--- a/osdc/modules/arc-runners-b200/defs/l-bx86iamx-176-1800-b200-8.yaml
+++ b/osdc/modules/arc-runners-b200/defs/l-bx86iamx-176-1800-b200-8.yaml
@@ -7,3 +7,5 @@ runner:
   vcpu: 176
   memory: 1800Gi
   gpu: 8
+  # Fixed-capacity cap: 1 reserved 8-GPU node / 8 GPUs per runner = 1 concurrent runner.
+  max_runners: 1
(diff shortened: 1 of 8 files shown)


jeanschmidt (approved):

jeanschmidt:
This would still be a problem if 1 job for -8 is running and 1 job for -2 arrives. But, it is sound.

I am implementing something similar now, but as you are there already, will just merge your changes on mine.

huydhn:
> This would still be a problem if 1 job for -8 is running and 1 job for -2 arrives. But, it is sound.
> 
> I am implementing something similar now, but as you are there already, will just merge your changes on mine.

Yeah, I thought about that point too, and decide to just go ahead with this as the initial iteration because this is better than not setting anything and allows us to safely run all 1xB200 jobs.  My thought process is that even if there is no better solution, we could still do the same fixed allocation like what we have now with X 1xB200 and Y 8xB200 to sum to to the total GPUs we have in the capacity reservations.

PR #500, opened 2026-04-24
Add proactive capacity PoC for ARC runners

jeanschmidt:
**Impact:** ARC runner infrastructure, all clusters (staging excluded)
**Risk:** low (feature is wired but disabled — `CAPACITY_AWARE_ENABLED=false` in all generated manifests)

## What

Adds a proactive capacity system that pre-warms cluster nodes by maintaining placeholder pods ahead of demand, and adds a single-label mode to the load test runner for targeted runner-type testing.

## Why

Stock ARC autoscaling is count-based and capacity-unaware — it tells GitHub "I can run N runners" without checking whether the cluster can actually schedule those pods. This causes job delays when Karpenter must provision new nodes from cold. Proactive capacity reserves node resources via low-priority placeholder pods that get preempted by real workloads, ensuring warm capacity is ready when jobs arrive.

Tracking issue: https://github.com/pytorch/ci-infra/issues/499

## How

- Uses a **forked ARC controller** (`jeanschmidt/actions-runner-controller`, branch `jeanschmidt/placeholder_run_poc`) that adds a `CapacityMonitor` goroutine to the listener binary. The fork is built and pushed to the cluster's local Harbor registry.
- Implements a **two-tier priority class preemption chain**: placeholder-runner (-10) → arc-runner (0) → placeholder-workflow (10) → arc-workflow (20). Workflow placeholders survive runner pod creation (higher priority), preventing cascading scheduling failures.
- Each runner definition declares its own `proactive_capacity` count (5–30 depending on runner size), injected as env vars on the listener container via the existing template generation pipeline.
- Staging clusters force `proactive_capacity=0` regardless of runner def values — safety net for non-production environments.
- Feature is globally disabled (`CAPACITY_AWARE_ENABLED=false`) — this PR wires the infrastructure; enabling is a separate step.

## Changes

**Proactive capacity infrastructure:**
- New `modules/arc/kubernetes/priority-classes.yaml` — four PriorityClasses (placeholder-runner, arc-runner, placeholder-workflow, arc-workflow) with preemption ordering
- New `modules/arc/kubernetes/capacity-monitor-rbac.yaml` — cross-namespace RBAC granting listener SAs (arc-systems) pod management in arc-runners
- Updated `modules/arc/deploy.sh` — ensures Harbor `osdc` project exists, applies priority classes + RBAC, switches Helm chart to forked ARC controller image from Harbor
- New `docs/arc-fork-build-deploy.md` — operational runbook for the fork (build, deploy, config, maintenance)

[...]

PR #511, opened 2026-04-29
Bump runner pod memory to 1Gi to prevent OOM under CRI backpressure

huydhn:
* #514
* __->__ #511

**Impact:** All ARC runner pods across OSDC clusters — affects job
reliability when nodes are under high pod-density / CRI load.
**Risk:** low

Increases memory request and limit on the `runner` container in the
ARC runner scale set template from 512Mi to 1Gi, and updates the
matching constant in the cluster simulator so capacity modeling
stays in sync.

Multiple PyTorch CI jobs failed because the runner pod (the one
hosting the GitHub Actions agent + Node.js container hooks)
OOMKilled mid-prepare_job:

- 2026-04-28 17:50: pytorch/pytorch run #25068774574 (lintrunner-
  noclang-all) — runner `mt-l-x86iamx-8-16-6kzqd-runner-gmxpk` on
  c7i.48xlarge node ip-10-4-55-181. Containerd recorded TaskOOM on
  the runner main container.
- 2026-04-28 18:16: pytorch/pytorch run #25069893231 (linux-jammy-
  py3.14t-clang18 / build-osdc) — runner
  `mt-l-x86iavx512-8-64-tk8zz-runner-fz8ns` on r7a.24xlarge node
  ip-10-4-23-113. exit 137 + TaskOOM ×2; workflow pod orphaned for
  ~10 min until GitHub's lost-runner heartbeat timeout fired.
- 2026-04-29 00:17: pytorch/pytorch runs #25084010294 and
  #25084223446 — **four** runner pods on c7i.48xlarge node
  ip-10-4-129-181 OOMed within 0.1s of each other during a
  density burst that peaked at 58 arc-runners pods on a single
  node, with kubelet housekeeping latency hitting 53.7s.

In every case the underlying trigger was kubelet/CRI overload during
a pod-density burst (~25-50 runner pods scheduled to one node within
~60s); kubelet's housekeeping loop fell 30-53 seconds behind its 1s
target, containerd reported `failed to delete task: context deadline
exceeded` on multiple sibling shims, and the runner pod's hook stdio
buffers accumulated past the 512Mi cgroup limit while waiting on
slow CRI exec calls.

The OOM is a tail consequence of CRI saturation, but the 512Mi cap
left no headroom: explicit caps allocate 200Mi (.NET managed heap
via `DOTNET_GCHeapHardLimit`) + 128Mi (V8 heap via
`--max-old-space-size`) = 328Mi out of 512Mi, leaving only ~184Mi
for native Node.js stdio buffers, OS, and process overhead. When
CRI exec hangs, native buffers grow beyond that margin.
`NODE_OPTIONS=--max-old-space-size` covers V8 heap only — not native
allocations such as tar stream and exec WebSocket buffers — and the
existing template comment already flagged this gap.

Bumping to 1Gi gives ~696Mi for native buffers and overhead — about
3.8x the previous headroom — without touching managed heap caps.
This is intended as the durable cap; an intermediate 768Mi step was
considered but, given the 04-29 four-runner OOM event with kubelet
latency >50s, the larger margin makes the fix robust against
similar bursts in the future.

- Single value change in the runner scale set template at lines 216
  and 219 (limit and request, kept equal for Guaranteed QoS):
  `memory: "512Mi"` -> `memory: "1Gi"`.
- Comment block at lines 187-193 updated with the new budget math
  (768Mi + 512Mi > 1024Mi at 75/50% claims still requires explicit
  caps) and a note pointing to native-buffer headroom as the
  motivation, with a reference to the calendar of incidents that
  drove the bump.
- `scripts/python/analyze_node_utilization.py:RUNNER_SIDECAR_MEM_MI`
  bumped 512 -> 1024 so the cluster simulator's per-pod accounting
  matches reality, plus the matching expected value updated in
  `test_analyze_node_utilization.py`.
- Existing managed heap caps unchanged: DOTNET_GCHeapHardLimit
  (200Mi) and NODE_OPTIONS=--max-old-space-size=128 still apply —
  the OOMs were not a managed-heap problem.
- Per-fleet `defs/*.yaml` and the generator script unchanged: runner
  pod memory is hardcoded in the template, so a single edit
  propagates to all 35+ fleets when `deploy.sh` regenerates
  `generated/*.yaml`.

- `modules/arc-runners/templates/runner.yaml.tpl`: runner container
  `resources.limits.memory` and `resources.requests.memory` 512Mi
  -> 1Gi; comment block updated.
- `scripts/python/analyze_node_utilization.py`:
  RUNNER_SIDECAR_MEM_MI 512 -> 1024.
- `scripts/python/test_analyze_node_utilization.py`: expected
  per-runner total memory updated to reflect the new constant
  (16384 + 1024 + 522 = 17930).

- This addresses the OOM symptom, not the root cause (kubelet/CRI
  overload from pod-density bursts). The CRI overload manifests as
  `Housekeeping took longer than expected actual=33-54s` on 20+
  nodes per 6h cluster-wide. A complementary mitigation — spreading
  runner pods across nodes via `topologySpreadConstraints` on the
  runner pod template — is being tracked separately.
- Validates clean against the existing runner-QoS validator
  (`scripts/python/validate_runner_qos.py` requires limits ==
  requests for Guaranteed QoS; 1Gi == 1Gi preserves that). All 13
  osdc linters pass; all unit tests pass.
- The generated `modules/arc-runners/generated/*.yaml` files are
  regenerated from this template by `deploy.sh` per-cluster; that
  directory is gitignored.
- Per-pod memory grows by 512Mi from the original 512Mi baseline.
  On every fleet's target NodePool there is enough margin in
  node-allocatable for the typical packing geometry. Three GPU
  fleets sized exactly to fit the 512Mi sidecar
  (`l-x86aavx2-189-704-a10g-8`, `l-x86aavx2-45-167-a10g-4`,
  `l-x86iavx512-29-115-t4`) now overshoot node-allocatable on a
  single-runner-per-node homogeneous packing by ~428Mi each. They
  still fit in mixed packings, so no jobs become unschedulable —
  but a fresh dedicated GPU node will not admit one of these
  runners on its own. Tracked as a follow-up to either trim those
  fleets' workflow memory or move them to slightly larger
  instance types.

## Test plan

Look at the simulate cluster output


malfet (approved):
LGTM, but as side note, this should be one line change, and one carries it over everywhere

jeanschmidt:
I didn't have the opportunity to tell you, but to get https://github.com/pytorch/ci-infra/pull/500 working properly, I needed to separate runners pods from workflow pods in different nodes. This was a future task planned anyways.

This means:
a) we need to update the packing computation scripts
b) you don't need to tune runner memory

runner pods now live on their own `c7i-runner` pool 

my recommendation: revert the changes to workflow memory and feel free to increase runner memory more aggressively.

yangw-dev (approved):

jeanschmidt (approved):

PR #534, opened 2026-05-06
Add A100 (p4de.24xlarge) on-demand GPU nodepool with 1/2/4/8x splits

huydhn:
**Impact:** ARC runner infrastructure on arc-cbr-production (us-east-2)
**Risk:** low

## What

Adds the p4de.24xlarge nodepool (8x NVIDIA A100 80GB) to the existing
`nodepools` module and four ARC runner sizes (1x/2x/4x/8x A100) to the
existing `arc-runners` module. Karpenter provisions p4de nodes purely
on demand — no warm floor.

## Why

PyTorch CI workloads currently submit ~1,803 jobs / 30d to
`linux.aws.a100` peaking at 38 concurrent (per
`docs/current_runner_load_distribution.md`) — a workload tier OSDC has
no on-cluster equivalent for. p4de.24xlarge offers 8x A100 80GB
on-demand at ~$40.97/hr in us-east-2 without a Capacity Block
reservation, so Karpenter can provision it directly on demand.

## How

- p4de.24xlarge goes into the existing `nodepools` module (not a
separate `nodepools-a100`) because it is plain on-demand Karpenter —
same lifecycle as g5/g6/g4dn. The `nodepools-h100` / `nodepools-b200`
modules stay separate only because they are tied to per-cluster
Capacity Block reservation IDs with their own renewal lifecycle.
- Single-instance `nodepool:` def format (matching H100/B200) since
p4de.24xlarge has fixed topology requirements
(`single-numa-node`, `topology_manager_scope: pod`) that don't fit the
multi-instance fleet pattern.
- No per-nodepool `user_data_script`: registry mirrors, CPU governor,
and GPU persistence mode are already handled cluster-wide by the base
`registry-mirror-config` and `node-performance-tuning` DaemonSets,
which tolerate every Karpenter taint. Same pattern as g5/g6/g4dn.
- `exclude_regions: [us-west-1]` keeps the nodepool out of arc-staging
(matching the g5/g6 pattern).
- No `proactive_capacity` on any A100 runner def — at ~$40.97/hr per
node, holding warm capacity is too expensive to justify by default.
Nodes scale up only when real jobs arrive; jobs absorb the ~5-10 min
provisioning latency on cold start. If queue times prove painful in
practice, raise `proactive_capacity` on the 1x A100 runner.
- Per-runner split: 88 vCPU / 1000 GiB usable across 8 GPUs (kubelet
reserves the remaining 8 vCPU / 152 GiB), proportional 11/125, 22/250,
44/500, 88/1000 splits.
- Runner names use `iavx512` (p4de.24xlarge runs Intel Cascade Lake —
AVX-512, no AMX), distinguishing them from H100/B200 `iamx`-tagged
runners.

## Changes

- **New** `modules/nodepools/defs/p4de-24xlarge.yaml` — single-instance
NodePool def, on-demand, `baremetal: true` for slow consolidation,
`exclude_regions: [us-west-1]`
- **New** four runner defs in `modules/arc-runners/defs/`:
  - `l-x86iavx512-11-125-a100.yaml` (1x A100)
  - `l-x86iavx512-22-250-a100-2.yaml` (2x A100)
  - `l-x86iavx512-44-500-a100-4.yaml` (4x A100)
  - `l-bx86iavx512-88-1000-a100-8.yaml` (8x A100, full node)
- **Modified** `scripts/python/instance_specs.py` — added
`p4de.24xlarge` (96 vCPU, 1152 GiB, 8 GPU) and ENI_MAX_PODS=250
- **Modified** `scripts/python/pytorch_workload_data.py` — added
`a.linux.a100[.{2,4,8}]` → new label mappings

## Notes

- p4de capacity is sometimes regionally constrained — Karpenter will
surface `InsufficientInstanceCapacity` if AWS has no p4de left in
us-east-2. Jobs queue until capacity returns.
- Cost: p4de.24xlarge ≈ $40.97/hr on-demand list price (~2.5x
g5.48xlarge). Pure on-demand scaling means $0 idle cost; peak (38
concurrent observed) bills 5 nodes for the duration of the burst.
- All four runner defs are deployed to both arc-staging and
arc-cbr-production by the existing `arc-runners` module wiring; only
the cbr cluster has a backing nodepool, so staging scale sets exist
without compute (matching how g5/g6 runners already behave).

## Testing

```
just test    # all tests pass, 99% coverage
just lint    # 13 linters pass
just plan arc-cbr-production
```

### DIFF
--- a/osdc/modules/arc-runners/defs/l-x86iavx512-11-125-a100.yaml
+++ b/osdc/modules/arc-runners/defs/l-x86iavx512-11-125-a100.yaml
@@ -1,0 +1,10 @@
+# ARC runner definition: l-x86iavx512-11-125-a100 (1x NVIDIA A100 80GB)
+# Scheduled on NodePool: p4de-24xlarge (from nodepools module)
+# Resource math: 88 usable vCPU / 8 GPUs = 11 vCPU/GPU, 1000 GiB / 8 = 125 GiB/GPU
+runner:
+  name: l-x86iavx512-11-125-a100
+  instance_type: p4de.24xlarge
+  disk_size: 200
+  vcpu: 11
+  memory: 125Gi
+  gpu: 1
--- a/osdc/modules/arc-runners/defs/l-bx86iavx512-88-1000-a100-8.yaml
+++ b/osdc/modules/arc-runners/defs/l-bx86iavx512-88-1000-a100-8.yaml
@@ -1,0 +1,9 @@
+# ARC runner definition: l-bx86iavx512-88-1000-a100-8 (8x NVIDIA A100 80GB, full node)
+# Scheduled on NodePool: p4de-24xlarge (from nodepools-a100 module)
+runner:
+  name: l-bx86iavx512-88-1000-a100-8
+  instance_type: p4de.24xlarge
+  disk_size: 200
+  vcpu: 88
+  memory: 1000Gi
+  gpu: 8
--- a/osdc/modules/arc-runners/defs/l-x86iavx512-22-250-a100-2.yaml
+++ b/osdc/modules/arc-runners/defs/l-x86iavx512-22-250-a100-2.yaml
@@ -1,0 +1,9 @@
+# ARC runner definition: l-x86iavx512-22-250-a100-2 (2x NVIDIA A100 80GB)
+# Scheduled on NodePool: p4de-24xlarge (from nodepools-a100 module)
+runner:
+  name: l-x86iavx512-22-250-a100-2
+  instance_type: p4de.24xlarge
+  disk_size: 200
+  vcpu: 22
+  memory: 250Gi
+  gpu: 2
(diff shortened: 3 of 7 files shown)


osdc/modules/arc-runners/defs/l-x86iavx512-11-125-a100.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: l-x86iavx512-11-125-a100
+  instance_type: p4de.24xlarge
+  disk_size: 200
+  vcpu: 11
+  memory: 125Gi
+  gpu: 1
+  # Maintain at least 1 placeholder pod so Karpenter keeps >=1 p4de.24xlarge
+  # node provisioned at all times. The placeholder consumes 1 GPU on the node;
+  # the remaining 7 GPUs are free for real runners (1x/2x/4x/8x A100 mix).
+  proactive_capacity: 1
```
jeanschmidt: this single guy, prevents to have 3 8 gpus instances running at the given time.

This is a proactive capacity for the runner, not the node.

Keep it a zero

malfet (approved):
Can you really get A100 on-demand?

huydhn:
> Can you really get A100 on-demand?

Yes, from what I see in https://github.com/meta-pytorch/pytorch-gha-infra/blob/main/runners/regions/us-east-1/pet_instances_benchmark/pet_instances.tf#L17 where all 3 multi-tenancy A100 instances there are on demand, without any capacity reservation.

However, as we have proactive capacity here set to 0 meaning there won't be any runners standing by, I want to test this out in staging on how fast we can get an A100 node up.

Issue #535, opened 2026-05-06
Base Node Migration Playbook — arc-cbr-production


jeanschmidt:
**One-time migration**: critical (base) nodepool from 5 × m7i.4xlarge → 6 × m7i.12xlarge.

This playbook is needed because:
- `aws_eks_node_group "base"` has `lifecycle.ignore_changes = [scaling_config[0].desired_size]` → tofu cannot bump the desired size; a manual AWS CLI step is required.
- An `instance_type` change triggers an EKS-managed rolling replacement of every base node — coordinate before clicking apply.

CI **does not** auto-deploy to arc-cbr-production. The production workflow is `workflow_dispatch` only, so it is safe to merge the PR and then take your time on the manual steps below.

---

### Risk mitigation
Stop the OSDC experiment as a risk reduction step: https://github.com/pytorch/test-infra/issues/5132

Wait 3-5h before proceeding

### Double-check quotas
```bash
# 1. Confirm no node group update is already in flight
NO_PROXY="$NO_PROXY,.eks.amazonaws.com" no_proxy="$no_proxy,.eks.amazonaws.com" \
aws eks list-updates \
    --cluster-name pytorch-arc-cbr-production \
    --nodegroup-name pytorch-arc-cbr-production-base-nodes \
    --region us-east-2

# 2. Check m7i family vCPU quota in us-east-2 (need at least 6×48 = 288 vCPU + headroom for the rolling overlap)
aws service-quotas get-service-quota \
    --service-code ec2 \
    --quota-code L-1216C47A \
    --region us-east-2

# 3. Review the tofu plan (this runs automatically on the PR via osdc-plan-prod.yml; double-check)
just plan arc-cbr-production
```

If quota is tight (default is often 1152), request an increase before merging.

---

## Step 1 — Merge the PR

Merge `clusters.yaml` change. **Do NOT trigger `OSDC: Deploy production` yet.**

---

## Step 2 — Set kubeconfig (laptop)

```bash
just kubeconfig arc-cbr-production
```

---

## Step 3 — Apply base infra (laptop)

```bash
OSDC_CONFIRM=ask just deploy-base arc-cbr-production
```

What this does:
- `tofu apply` for `modules/eks/terraform` → EKS starts rolling-replacing base nodes (max 33% unavailable at a time → 2 nodes drain at once).
- Applies `base/kubernetes/` manifests, including `git-cache central` StatefulSet scaled to 15 replicas.

[...]

PR #536, opened 2026-05-06
Scale up OSDC defaults for higher load

jeanschmidt:
**Impact:** All OSDC production and default cluster infrastructure (base nodes, Harbor, Karpenter, ARC, BuildKit, git cache, PyPI cache)
**Risk:** low
**Migration Complexity:** high

## What
Increases replica counts, resource allocations, and instance sizes across OSDC cluster defaults and the `arc-cbr-production` cluster to handle higher load.

## Why
Proactive scale-up to prevent service disruption under increased load. Taking a conservative "safer than sorry" approach — components can be scaled back down once we confirm the higher capacity is not needed.

## How
- Scales vertically (larger instance types, more CPU/memory) and horizontally (more replicas) in tandem
- Changes are applied at the defaults level where possible, so all clusters inherit the increase unless they override
- Production git cache and PyPI cache get significantly larger bumps than the defaults, reflecting their higher traffic
- GPU consolidation window doubled (20m -> 40m) to reduce churn under higher load, and reduce the likelyhood of giving instances back and not being able to get them again due to `InsufficientCapacity`.

## Changes

**Base infrastructure:**
- Base node count: 5 -> 6, instance type: m7i.4xlarge -> m7i.12xlarge (3x vCPUs/memory per node)

**Harbor (container registry cache):**
- Core replicas: 2 -> 4, registry replicas: 2 -> 4, nginx replicas: 3 -> 6

**Control plane components:**
- Karpenter replicas: 2 -> 4
- ARC controller replica count: 2 -> 4

**Build infrastructure:**
- BuildKit replicas per arch: 4 -> 5

**Node management:**
- GPU consolidation window: 20m -> 40m (more conservative node reclaim)

**arc-cbr-production overrides:**
- Git cache central replicas: 5 -> 15, CPU request: 6 -> 10, memory request: 8Gi -> 10Gi, memory limit: 32Gi -> 50Gi
- PyPI cache replicas: 5 -> 10

## Notes
- This is intentionally over-provisioned. Plan to evaluate actual utilization and scale back as appropriate.
- Staging cluster is unaffected — it has its own overrides.
- The m7i.12xlarge base nodes are 48 vCPU / 192 GiB each (up from 16 vCPU / 64 GiB), so total base node capacity increases roughly 3.6x (6 nodes * 3x resources vs 5 nodes * 1x).
- Base node (Critical taint) capacity stays 2x used capacity for disruption budgets without service interruption.

# Migration
The playbook to perform this migration with minimal disruption to live production traffic is outlined in the issue: https://github.com/pytorch/ci-infra/issues/535

### DIFF
--- a/osdc/clusters.yaml
+++ b/osdc/clusters.yaml
@@ -21,8 +21,8 @@
 
 defaults:
   # --- Terraform / base infra ---
-  base_node_count: 5
-  base_node_instance_type: m7i.4xlarge
+  base_node_count: 6
+  base_node_instance_type: m7i.12xlarge
   base_node_max_unavailable_percentage: 33
   # TODO(CVE-2026-31431): when bumping to a kernel 6.12.85+ AL2023 AMI,
   # remove osdc/base/kubernetes/algif-mitigation.yaml.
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM!

PR #543, opened 2026-05-07
Fix git cache slot rotation on fetch failure

jeanschmidt:
**Impact:** All CI runners consuming the git cache (pytorch/pytorch, pytorch/test-infra)
**Risk:** medium

## What
Prevents the git cache central pod from rotating its active slot when a `git fetch` fails, and adds `--force` to all fetch calls to handle mutable tag overwrites that cause non-zero exits on git 2.52.0.

## Why
A compound bug caused the git cache to silently serve stale content:

1. `git fetch origin --prune --tags` ran without `--force`, so tag clobbers (e.g. PyTorch's `ciflow/*` delete-and-recreate pattern at ~70-88/hr) caused fetch to exit non-zero.
2. A git v2.51.0 regression (batched ref-update transactions) makes any per-ref rejection abort the entire transaction — even branches that would have advanced are silently discarded. The `python:3.12-alpine` base image ships git 2.52.0 (affected), and Alpine hasn't picked up the v2.53.0 fix.
3. Despite the fetch failure, the update loop unconditionally rotated the active slot, promoting a staging slot with stale data as if it were fresh.

The net effect: the cache appeared healthy (slot swaps continued, metrics showed activity) but served progressively staler repo content until the pod restarted.

## How
- All `git fetch` calls now use a shared `FETCH_ARGS` constant that includes `--force` and `--prune-tags`, preventing tag-clobber rejections from poisoning the entire fetch transaction.
- All clone/fetch functions (`clone_full_repo`, `clone_bare_repo`, `fetch_full_repo`, `fetch_bare_repo_update`) now return `bool` indicating success/failure instead of `None`.
- The main update loop tracks a `cycle_ok` flag across all repos; slot rotation is gated on every repo succeeding. On failure, the previous active slot is retained and a counter is incremented.
- A new `git_cache_central_cycle_failed_total` Prometheus counter is registered and zero-initialized at startup so `rate()` works from the first scrape.
- Submodule update failures in `fetch_full_repo` now emit the same error metrics as fetch failures (previously they were silently swallowed with only a log warning).

## Changes
- **`central-configmap.yaml` (central.py)**:
  - Added `FETCH_ARGS = ["--prune", "--prune-tags", "--tags", "--force"]` constant
  - Changed `clone_full_repo`, `clone_bare_repo`, `fetch_full_repo`, `fetch_bare_repo_update` return types from `None` to `bool`
  - Added `cycle_ok` tracking in the main loop; slot rotation only happens when all repos update successfully
  - Added `git_cache_central_cycle_failed_total` metric definition and zero-initialization
  - Added error metrics emission for submodule update failures in `fetch_full_repo`
- **`central_lib.py`** (testable mirror):
  - Added `git_cache_central_cycle_failed_total` metric definition to `MetricsServer.METRIC_DEFS`

## Notes
- The root git bug (v2.51.0 batched ref-update regression) is fixed in git v2.53.0. Once Alpine 3.23 picks up that version, `--force` becomes a belt-and-suspenders measure rather than a hard requirement. The slot-rotation guard remains valuable regardless.
- The `--prune-tags` flag (in addition to `--prune`) ensures deleted upstream tags are cleaned up locally, preventing unbounded tag accumulation from PyTorch's ciflow churn.

## Testing
```
============================================================
  OSDC Integration Test Results
============================================================
  Cluster: arc-staging (pytorch-arc-staging)
  Date:    2026-05-07 17:53 UTC

  PR Workflow Jobs:
    ✓ test-pypi-cache-action-cpu     success
    ✓ test-pypi-cache-action-cuda    success
    ✓ test-cpu-x86-amx               success
    ✓ test-cpu-x86-avx512            success
    ✓ test-cpu-arm64                 success
    ✓ test-harbor                    success
    ✓ test-cache-enforcer            success
    ✓ test-gpu-t4-multi              success
    ✓ test-git-cache                 success
    ✓ test-release-arm64             success
    ✓ test-gpu-t4                    success
    ✓ test-pypi-cache-defaults       success
    ✓ build-arm64 / build            success
    ✓ build-amd64 / build            success

  Smoke            ⊘ SKIPPED
  Compactor        ⊘ SKIPPED

  Overall: PASSED
============================================================
```

### DIFF
--- a/osdc/base/kubernetes/git-cache/central-configmap.yaml
+++ b/osdc/base/kubernetes/git-cache/central-configmap.yaml
@@ -69,6 +69,7 @@ data:
             "git_cache_central_active_slot": ("gauge", "Current active slot (0=a, 1=b)"),
             "git_cache_central_cycle_duration_seconds": ("gauge", "Duration of last full update cycle"),
             "git_cache_central_rsyncd_restarts_total": ("counter", "rsyncd daemon restart count"),
+            "git_cache_central_cycle_failed_total": ("counter", "Update cycles that did not rotate the active slot due to failures"),
             "git_cache_central_repos_total": ("gauge", "Number of repos being cached"),
             "git_cache_central_clone_total": ("counter", "Total clone/fetch operations"),
             "git_cache_central_clone_duration_seconds": ("histogram", "Time to clone/fetch each repo"),
@@ -238,6 +239,11 @@ data:
         "pytorch/test-infra",
     ]
 
+    # Common args for `git fetch` calls. --force --prune-tags is required because
+    # PyTorch CI rewrites mutable tags (ciflow/*, viable/strict/*) — without --force,
+    # git rejects "would clobber existing tag" and the fetch exits non-zero.
+    FETCH_ARGS = ["--prune", "--prune-tags", "--tags", "--force"]
+
 
     def run(cmd: list[str], **kwargs) -> subprocess.CompletedProcess:
         """Run a command, logging it and raising on failure unless check=False."""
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM!

PR #565, opened 2026-05-13
Exclude high-end GPUs from load tests

jeanschmidt:
**Impact:** CI load tests only — no production runner changes
**Risk:** low

## What
Excludes specialized high-end GPU runner definitions (A100, H100, B200, H200, MI300, MI325) from the load test distribution so load tests only target commodity GPU types (T4, A10G, L4) and CPU/ARM runners.

## Why
High-end GPU instances have very limited capacity in staging, multi-hour queue waits, and are hard to acquire on-demand. Including them in load tests wastes capacity reservations, causes long test timeouts, and doesn't add meaningful signal — the load test validates runner orchestration (ARC workflow dispatch, pod scheduling, job completion), not GPU-specific behavior.

## How
- Pattern-based exclusion via regex applied early in `get_available_runners`, before region-based fleet filtering — keeps the logic centralized and independent of runner def file contents
- The exclusion pattern (`_LOAD_TEST_EXCLUDED_GPU_PATTERN`) is separate from the existing `_OSDC_GPU_PATTERN` used for classification, since the two serve different purposes (classification vs. filtering) and cover different GPU sets
- The `excluded_count` return value now aggregates both region-based and GPU-based exclusions; the log message was updated to reflect this

## Changes
- **`distribution.py`**: Added `_LOAD_TEST_EXCLUDED_GPU_PATTERN` regex matching high-end GPU suffixes (`-a100`, `-h100`, `-b200`, `-h200`, `-mi300`, `-mi325` with optional `-<count>`). `get_available_runners` filters matching defs before region checks, incrementing the shared `excluded_count`. Updated docstring accordingly.
- **`load_test_run.py`**: Updated log message to clarify that `excluded_count` includes both region and specialized GPU exclusions.
- **`test_distribution.py`**: Added 5 test cases covering: single A100 exclusion, all 6 denied GPU types, allowed GPU types (T4/A10G/L4) pass through, multi-GPU suffix variants, and no false positives on CPU runners with numeric suffixes.
- **`.coverage`**: Binary coverage DB updated (artifact of running tests locally).

## Notes
- The `.coverage` binary file should ideally be gitignored rather than tracked. Consider adding it to `.gitignore` in a follow-up.


osdc/integration-tests/load-test/scripts/python/distribution.py
```diff
@@ -177,7 +182,10 @@ def get_available_runners(
 ) -> tuple[set[str], int]:
     """Scan runner def YAML files and return available runner labels.
 
-    Returns (labels, excluded_count).
+    Returns (labels, excluded_count). ``excluded_count`` covers both
+    region-based exclusions (fleets unavailable in the target region) and
+    specialized-hardware exclusions (high-end GPU defs matching
+    ``_LOAD_TEST_EXCLUDED_GPU_PATTERN``).
```
malfet, line 188: This comment will run the markdown

malfet (approved):

PR #580, opened 2026-05-16
Add arc-cbr-production-uw1 entry to clusters.yaml

huydhn:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #585
* #584
* #587
* #583
* __->__ #580
* #581
* #586

Adds the new us-west-1 prod cluster definition. Mirrors
arc-cbr-production (us-east-2) for everything except:

- region: us-west-1
- vpc_cidr: 10.8.0.0/16 (non-overlapping with us-east-2's 10.4.0.0/16
  and staging's 10.0.0.0/16)
- runner_group: "arc-cbr-prod-uw1" (distinct from us-east-2's default)
- nodepools-b200 + arc-runners-b200 dropped (no B200 capacity reserved
  in us-west-1)
- nodepools-h100.capacity_reservation_ids: [] placeholder

This commit is non-functional until the prerequisites listed in
docs/prod-cluster-ha-us-west-1.md land:
1. Org runner group `arc-cbr-prod-uw1` created at pytorch.
2. Generator support for cluster-level runner_group override.
3. Generator support for cluster-level capacity_reservation_ids override.
4. Real H100 reservation IDs filled into the placeholder array.

Keeping the entry in main lets the design doc reference a concrete
example and lets the prerequisite PRs reference this entry while they
add the generator changes.

PR #583, opened 2026-05-16
Support cluster-level capacity_reservation_ids override in nodepools

huydhn:
* #585
* #584
* #587
* __->__ #583
* #580
* #581
* #586

Adds a per-cluster override for the capacity_reservation_ids that
were previously hardcoded in nodepool def files (e.g.
modules/nodepools-h100/defs/p5-48xlarge.yaml). The override key in
clusters.yaml is namespaced per module so b200 and h100 stay
independent:

    clusters:
      arc-cbr-production-uw1:
        nodepools-h100:
          capacity_reservation_ids:
            - cr-04d3d1d84e127a562
            - cr-09a53051589034fb8

This unblocks the multi-region prod design: one shared module def
file, different reservation IDs per region. The us-east-2 IDs in the
def file remain as the fallback when no cluster-level override is
set.

Implementation:

- cluster-config.py: print list values as comma-separated for shell
  consumption. Existing string/bool behaviors unchanged.
- modules/nodepools/deploy.sh: read clusters.<id>.<MODULE_NAME>.
  capacity_reservation_ids and pass as
  NODEPOOLS_CAPACITY_RESERVATION_IDS_OVERRIDE env var. MODULE_NAME
  comes from the nodepools-b200 / nodepools-h100 delegators, so the
  override key namespaces automatically.
- generate_nodepools.py: if the env var is set and non-empty, parse
  it (comma-separated) and override the def file's value. Empty
  string leaves the def value alone.

Adds 5 new tests:
- cluster-config.py list output format (comma-separated, empty list)
- Generator override wins over def value
- Override applies when def has no value set
- Empty override env var keeps def value intact

See osdc/docs/prod-cluster-ha-us-west-1.md "Phase 1 prerequisite 2"
for the broader context.

### DIFF
--- a/osdc/modules/nodepools/scripts/python/generate_nodepools.py
+++ b/osdc/modules/nodepools/scripts/python/generate_nodepools.py
@@ -138,6 +138,14 @@ def generate_nodepool_yaml(nodepool_def, module_name, defs_dir=None):
     capacity_type = nodepool_def.get("capacity_type", "on-demand")
     capacity_reservation_ids = nodepool_def.get("capacity_reservation_ids", [])
 
+    # Cluster-level override (set per cluster in clusters.yaml under e.g.
+    # `nodepools-h100.capacity_reservation_ids`). Used for multi-region prod
+    # where the same module ships shared defs but each cluster has its own
+    # capacity reservations. Wins over the def file's value when present.
+    cluster_cr_override = os.environ.get("NODEPOOLS_CAPACITY_RESERVATION_IDS_OVERRIDE", "")
+    if cluster_cr_override:
+        capacity_reservation_ids = [s for s in cluster_cr_override.split(",") if s]
+
     # ----- Node compactor opt-in -----
     # NodePools labeled osdc.io/node-compactor are managed by the compactor
     # controller, which handles consolidation via NoSchedule taints instead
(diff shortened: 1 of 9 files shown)


osdc/modules/nodepools/scripts/python/generate_nodepools.py
```diff
@@ -138,6 +138,14 @@ def generate_nodepool_yaml(nodepool_def, module_name, defs_dir=None):
     capacity_type = nodepool_def.get("capacity_type", "on-demand")
     capacity_reservation_ids = nodepool_def.get("capacity_reservation_ids", [])
 
+    # Cluster-level override (set per cluster in clusters.yaml under e.g.
```
jeanschmidt: over documenting here as well

osdc/scripts/cluster-config.py
```diff
@@ -128,6 +128,10 @@ def main():
                 sys.exit(1)
         elif isinstance(val, bool):
             print(str(val).lower())
+        elif isinstance(val, list):
+            # Comma-separated for shell consumption (used by nodepools deploy.sh
```
jeanschmidt: over commenting...

jeanschmidt (changes requested):

jeanschmidt (approved):

PR #584, opened 2026-05-16
Allow osdc-deploy-prod to target either prod cluster

huydhn:
* #585
* __->__ #584
* #587
* #583
* #580
* #581
* #586

Adds a workflow_dispatch `cluster` choice input so the prod deploy
workflow can target arc-cbr-production (us-east-2, default) OR
arc-cbr-production-uw1 (us-west-1).

### DIFF
--- a/.github/workflows/osdc-deploy-prod.yml
+++ b/.github/workflows/osdc-deploy-prod.yml
@@ -1,12 +1,23 @@
 name: "OSDC: Deploy production"
 
-# Manual rollout for arc-cbr-production. Protected by the `osdc-production` GitHub
+# Manual rollout for any prod cluster. Protected by the `osdc-production` GitHub
 # environment (configure required reviewers in repo Settings → Environments) —
 # the prod IAM role additionally trusts only tokens with
 # `environment:osdc-production` in the sub claim, so IAM enforces the gate too.
+#
+# Concurrency is scoped per-cluster so a us-east-2 deploy doesn't block a
+# us-west-1 deploy and vice versa.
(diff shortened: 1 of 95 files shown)


.github/workflows/osdc-deploy-prod.yml
```diff
@@ -24,7 +35,7 @@ on:
         default: false
 
 concurrency:
-  group: osdc-deploy-arc-cbr-production
+  group: osdc-deploy-${{ inputs.cluster }}
```
jeanschmidt: should we, for pure best practices enforcing, use the same concurrency key? In order to avoid someone deploying to two production clusters simultaneously? 

I guess this would be better.

jeanschmidt:
You know what would be cool, not sure if we should do or not, but worth discussing:

instead of deploying a single cluster per call, by default it deploys to one cluster, runs the full test battery and if all successful deploys to the next.

A selector could force only one or another when triggering the workflow.

This would make it operationally much simpler for 99% of deploys and much safer (not deploying both at the same time, only proceeding with the next if the first pass tests, etc)

jeanschmidt (changes requested):

.github/workflows/osdc-deploy-prod.yml
huydhn: This is done. Now we deploy the 2 cluster sequentially, one after another with testing. In the case of fire fighting, we can still choose the cluster to deploy

jeanschmidt (approved):

PR #585, opened 2026-05-16
Plan both prod clusters on PRs sequentially

huydhn:
* __->__ #585
* #584
* #587
* #583
* #580
* #581
* #586

Extracts the plan + PR-comment logic into a reusable _osdc-plan.yml
and has osdc-plan-prod call it twice — once per cluster, in sequence
(us-east-2, then us-west-1). Each cluster's plan posts its own PR
comment with a per-cluster marker.

The second leg runs regardless of the first's outcome (via
`!cancelled()`) so reviewers see both diffs even when one is broken.

### DIFF
--- a/.github/workflows/osdc-plan-prod.yml
+++ b/.github/workflows/osdc-plan-prod.yml
@@ -5,6 +5,7 @@ on:
     paths:
       - "osdc/**"
       - ".github/workflows/osdc-plan-prod.yml"
+      - ".github/workflows/_osdc-plan.yml"
       - ".github/workflows/_osdc-deploy.yml"
 
 concurrency:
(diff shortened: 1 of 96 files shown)


.github/workflows/osdc-plan-prod.yml
```diff
@@ -22,7 +22,15 @@ defaults:
 
 jobs:
   plan:
-    name: tofu plan (arc-cbr-production)
+    name: tofu plan (${{ matrix.cluster }})
```
jeanschmidt: parallel?

NOOOOOOO!!!!!!!!!

jeanschmidt (changes requested):

jeanschmidt (approved):

PR #588, opened 2026-05-17
Add stale scale-set detection and healing

jeanschmidt:
- Add post-deploy check (Step 6) in arc-runners deploy.sh that detects listeners crash-looping with RunnerScaleSetNotFoundException
- Add `just heal-arc <cluster> [name]` recipe to re-register AutoscalingRunnerSets by clearing stale scale-set-id annotations
- Document the stale listener recovery workflow in arc-fork-build-deploy.md

When the GitHub broker loses a scale-set ID (app permission rotation, repo rename, manual deletion, or broker key rotation), the ARC controller trusts the stale annotation and never self-heals. The heal-listeners recipe strips the annotation and deletes the old listener, forcing the controller to call CreateRunnerScaleSet and obtain a fresh ID.


huydhn (approved):
LGTM!

osdc/modules/arc-runners/deploy.sh
```diff
@@ -228,3 +228,106 @@ if ((${#STALE_RELEASES[@]} > 0)); then
     echo "No orphaned Helm history secrets found."
   fi
 fi
+
+# --- Step 6: Detect stale AutoscalingListeners ---
+LISTENER_CHECK="${ARC_LISTENER_CHECK:-enabled}"
```
huydhn, line 233: We are adding more bash scripts along the way. This would get harder to replace them with Python later :(

jeanschmidt: yeah....

PR #593, opened 2026-05-19
Add pause_runners cluster-level flag

jeanschmidt:
- Read pause_runners from clusters.yaml and propagate to runner generation
- Force maxRunners to 0 for all scale sets when pause_runners is true
- Log a warning when pause_runners is active on a cluster
- Add unit and integration tests covering pause/unpause and unset cases

Allows new clusters to be deployed with runners fully paused, preventing scale sets from accepting jobs before the cluster is validated. Set pause_runners: true in clusters.yaml, deploy, verify, then remove the flag to go live.


huydhn (approved):

huydhn:
Let's merge this so that I can rebase https://github.com/pytorch/ci-infra/pull/585 on top

PR #598, opened 2026-05-20
Make resume-runners respect region-excluded scale sets

huydhn:
## Summary

- PR #597 taught `generate_runners.py` to render `maxRunners: 0` and `proactive_capacity: 0` for runners whose nodepool excludes the cluster's region (e.g. `g5`, `g6`, `p4d`, `c7a`, `r7a` defs that carry `exclude_regions: [us-west-1]`).
- `runner_max_map.py` (used by `just resume-runners`) reads the same runner defs but did not consult the nodepool exclusions. After a `just drain-runners` cutover on `arc-cbr-production-uw1`, it would patch every excluded scale set back to `MAX_INT32`, silently undoing #597 until the next `just deploy`.
- This change imports `load_excluded_instance_types` (added in #597) and forces `max_runners=0` for any runner whose `instance_type` matches an excluded nodepool def. `parse_def_file` now also returns `instance_type` so the caller can apply the guard.

## Test plan

- [x] `just lint` — 13/13 pass
- [x] `just test` — full suite passes (78 tests in `test_runner_max_map.py`, including 3 new cases covering region-excluded behavior)
- [ ] Manual: run `uv run runner_max_map.py arc-cbr-production-uw1` after merge and confirm every region-excluded scale set (g5/g6/p4d/c7a/r7a) reports `0` in the JSON map.
- [ ] Manual: smoke-test `just drain-runners arc-cbr-production-uw1 && just resume-runners arc-cbr-production-uw1` and verify excluded scale sets remain at `maxRunners: 0`.

PR #599, opened 2026-05-20
[BE] Migrate p4d-24xlarge nodepool def to fleet format

huydhn:
## Summary

- `p4d-24xlarge.yaml` was the last def using the legacy `nodepool:` format. Two parallel formats forces every consumer to handle both shapes, and at least one consumer (load-test `distribution.py`) silently skips the legacy form — which would let region-excluded `p4d.24xlarge` show up in load-test allocations even though no NodePool exists to schedule it.
- Convert `p4d-24xlarge` to a single-instance `fleet:` so every consumer that handles `fleet:` works automatically (including `arc-runners/generate_runners.py`'s `load_excluded_instance_types`, `runner_max_map.py` from #598, and load-test `distribution.py`). The legacy `nodepool:` codepath in `generate_nodepools.py` stays for now in case external consumers still emit that shape.
- Rename the def file to `p4d.yaml` to match the directory convention (file name = fleet name, as with `g5.yaml`, `c7a.yaml`, `m7g.yaml` etc.). The generated NodePool name stays `p4d-24xlarge` because `_fleet_nodepool_name` derives it from fleet + instance, not from the def filename.
- The generated Karpenter NodePool is byte-identical to the previous output apart from a `weight: 100` field, which is inert here because the `instance-type=p4d.24xlarge` taint pins pods to this NodePool — no other NodePool advertises `p4d.24xlarge`.

### Diff against the previous generated NodePool

```diff
+  weight: 100
```

That's it — same NodePool name (`p4d-24xlarge`), same `instance-type`, same taints/labels (`node-fleet: p4d` was already auto-derived from `instance_type.split(\".\")[0]`, still `p4d` from the explicit fleet name), same `topology_manager_policy: single-numa-node`, same `exclude_regions: [us-west-1]`.

## Test plan

- [x] `just lint` — 13/13 pass
- [x] `just test` — full suite passes
- [x] Manual round-trip: ran `generate_nodepools.py` against the old and new defs and diff'd the generated YAML; only `weight: 100` differs.
- [ ] Post-merge: after the next prod deploy, verify the `p4d-24xlarge` NodePool still scales A100 jobs (`l-x86iavx512-44-500-a100-4` etc.) and that `kubectl describe np p4d-24xlarge` shows `Weight: 100`.

## Follow-up

Once this lands, the legacy `nodepool:` codepath in `generate_nodepools.py` (`_process_nodepool`) is dead code internally. Leaving it in place for now since the smoke test and tests still exercise it; can be removed in a separate cleanup PR once we're confident no external consumer emits that shape.


osdc/modules/nodepools/defs/p4d.yaml
```diff
@@ -0,0 +1,23 @@
...
+fleet:
+  name: p4d
+  arch: amd64
+  gpu: true
+  exclude_regions:
+    - us-west-1
+  instances:
+    - type: p4d.24xlarge
+      weight: 100
+      node_disk_size: 100
```
jeanschmidt: is 100Gb sufficient? I guess this is too small for this kind of instance

jeanschmidt:
NIT: When doing those changes avoid deleting and re-creating files, use git mv to keep track of git history properly

jeanschmidt (approved):

malfet (approved):

osdc/modules/nodepools/defs/p4d.yaml
huydhn: I haven't seen any issue, so I guess this is ok.  We only have small jobs running on A100 nowadays

PR #603, opened 2026-05-20
Add Renovate runner-image auto-update pipeline

jeanschmidt:
# Renovate runner-image auto-update pipeline

## What this PR does

Adds an end-to-end automation that detects new releases of the GitHub Actions runner image (`ghcr.io/actions/actions-runner`), opens a PR bumping `runner_image_tag` in `clusters.yaml`, validates the change in staging through the existing merge queue, and on merge deploys the new image to both prod clusters sequentially — all without human intervention on the happy path, but with multiple defense-in-depth gates that fail safely on anything anomalous.

## Why

Today, runner-image bumps are manual: someone notices a new release, edits `clusters.yaml`, opens a PR, gets it reviewed, merges, then manually runs the prod deploy. This is toil that scales poorly and lags behind upstream releases. Stale runner images mean missing bug fixes, security patches, and GHA feature parity.

## The flow

```
   schedule (Mon/Thu 14:00 UTC)
            │
            ▼
   ┌──────────────────────────┐
   │  osdc-renovate.yml       │  scans clusters.yaml, queries
   │  (environment:           │  ghcr.io for new image tags
   │   osdc-renovate,         │  (≥5 days old, <3.0.0)
   │   UPDATEBOT_TOKEN)       │
   └────────────┬─────────────┘
                │
                ▼ opens PR as pytorchupdatebot
                │  (renovate-runner/* branch,
                │   auto-runner-update label)
                │
                ▼ workflow run completes
                │
   ┌────────────▼─────────────────┐
   │  osdc-renovate-autoapprove   │  triggered by workflow_run
   │  (environment:               │  (no per-PR retrigger; one
   │   osdc-renovate,             │   job processes all eligible
   │   GH_PYTORCHBOT_TOKEN)       │   open PRs)
   │                              │
   │  ├─ discover                 │  lists open auto-runner-update
   │  │                           │  PRs, splits into approve vs.
   │  │                           │  close-stale (>7 days)
   │  │                           │
   │  ├─ approve (matrix)         │  author check + diff-content
   │  │                           │  gate (single-file, single-line,
   │  │                           │  semver, monotonic) + SHA-pinned
   │  │                           │  approval and auto-merge.
   │  │                           │  Deterministic validation failures
   │  │                           │  → auto-close + explanatory comment.
   │  │                           │
   │  └─ close_stale (matrix)     │  closes PRs open >7 days with
   │                              │  explanatory comment; Renovate
   │                              │  reopens on next run if still
   │                              │  eligible.
   └────────────┬─────────────────┘
                │
                ▼ gh pr merge --squash --auto --match-head-commit
                │
   ┌────────────▼─────────────────┐
   │  Merge queue                 │  runs osdc-pre-merge.yml:
   │                              │   full arc-staging deploy +
   │                              │   smoke + integration + load
   │                              │   + janitor + compactor
   └────────────┬─────────────────┘
                │
                ▼ if green, squash-merges to main
                │
   ┌────────────▼─────────────────────┐
   │  osdc-auto-update-deploy-prod    │  triggers on push to main
   │                                  │  (clusters.yaml changed)
   │  ├─ check-trigger                │  re-runs the SAME diff-
   │  │   (triple gate + diff guard)  │  content validation
   │  ├─ deploy_uw1                   │  per-module deploy +
   │  │   (arc-runners,               │  smoke
   │  │    arc-runners-h100)          │
   │  └─ deploy_ue2                   │  per-module deploy +
   │      (arc-runners,               │  smoke
   │       arc-runners-b200,          │
   │       arc-runners-h100)          │
   └──────────────────────────────────┘
```

## Safety design

The pipeline can deploy to prod without human review on the happy path, so every gate is intentionally redundant — a failure of any single layer should not be able to roll out a malicious or malformed change:

- **Two separate bot identities, both shared with pytorch/test-infra.** `UPDATEBOT_TOKEN` (owner: `pytorchupdatebot`) opens PRs; `GH_PYTORCHBOT_TOKEN` (owner: `pytorchbot`) approves them. GitHub forbids self-approval, so a compromise of either token alone cannot merge anything. Both are existing PATs maintained by the pytorch/test-infra team — reusing them avoids minting new bot accounts that would have to sign the Meta CLA before their commits could land. Both tokens likely have broader org-wide scope than this workflow strictly requires; that scope is outside our control. Both are read from a dedicated `osdc-renovate` GitHub Environment so only opted-in jobs in this repo can read them.
- **Triple metadata gate.** Both the auto-approve and the prod-deploy workflows require: PR author matches the configured Renovate bot (`pytorchupdatebot`), branch prefix `renovate-runner/`, and label `auto-runner-update`. The auto-approve workflow additionally gates on the upstream `workflow_run` having `conclusion == 'success'`, `head_branch == default_branch`, the head repo matching the base repo, and the trigger event NOT being `pull_request` — blocking fork-spoofing and PR-triggered renovate runs.
- **Diff-content gate (twice).** The auto-approve workflow refuses to approve unless the PR's diff is exactly one file (`osdc/clusters.yaml`), exactly one line added + one line removed, both lines match a strict `runner_image_tag` semver pattern, and the new version is a monotonic upgrade. The prod-deploy workflow runs the same validation again against the merge commit. A compromise of `UPDATEBOT_TOKEN` cannot push arbitrary content through, even if branch protection were misconfigured.
- **SHA pinning between validation and approval.** The auto-approve workflow records `pull_request.head.sha` at the moment it validates the diff, then passes it to GitHub's review API via `commit_id` and to auto-merge via `--match-head-commit`. If the PR head moves between validate and review (force-push race), GitHub rejects the review with 422 and auto-merge is invalidated — preventing approval of unvalidated content.
- **Auto-close on deterministic validation failure.** Wrong file, wrong line count, regex mismatch, downgrade, no version change → PR is closed with a comment naming the failure mode. Renovate will reopen on the next run when a new eligible version is available. Transient failures (e.g. `gh` API errors) propagate as job failures and leave the PR open.
- **Auto-close on staleness.** PRs open >7 days are closed with a comment. Renovate reopens on the next run if the version is still relevant. To pause without auto-closure, remove the `auto-runner-update` label. Staleness is computed from `updatedAt` (not `createdAt`) so PRs Renovate keeps rebasing aren't closed under us.
- **Author check fails loudly, not silently.** If the PR author doesn't match `OSDC_RENOVATE_BOT_LOGIN`, the job fails and the PR is left open for human investigation (likely mislabeling).
- **Merge queue does full staging validation.** The auto-approve enqueues the PR; merge queue runs the existing `osdc-pre-merge.yml` which deploys + smokes + integrates + load-tests against arc-staging. The PR only merges if staging is green.
- **Prod deploy uses the reusable workflow.** Same setup, same auth, same environment gating (`osdc-production`) as the manual `osdc-deploy-prod.yml`. The auto-deploy passes a `modules:` input so only the runner modules (the only consumers of `runner_image_tag`) re-deploy, and uw1 / ue2 deploy sequentially with smoke between them.
- **Concurrency group shared with manual prod deploy** — auto and manual prod deploys serialize, never overlap.
- **Constant concurrency group on auto-approve** — re-runs of `osdc-renovate` serialize through autoapprove with `cancel-in-progress: false`, so the second run waits for the first to finish instead of issuing a duplicate `gh pr merge --auto`.

## What this PR touches

| File | Purpose |
|------|---------|
| `osdc/renovate.json` | Renovate config: scans `clusters.yaml` for `runner_image_tag`, queries GHCR via `docker` datasource, opens PRs with the runner-update label. |
| `.github/workflows/osdc-renovate.yml` | Scheduled (Mon/Thu) Renovate run. Pinned to the `osdc-renovate` Environment so it can read `UPDATEBOT_TOKEN`. Also supports manual dispatch with dry-run option. |
| `.github/workflows/osdc-renovate-autoapprove.yml` | Triggered by `workflow_run` of `osdc-renovate`. `discover` job lists eligible PRs and splits into approve / close-stale matrices; `approve` job runs author + diff-content gate + SHA-pinned approval/auto-merge; `close_stale` closes PRs open >7 days. Reads `GH_PYTORCHBOT_TOKEN` from the `osdc-renovate` Environment. |
| `.github/workflows/osdc-auto-update-deploy-prod.yml` | Post-merge prod deploy of the bumped runner image to both clusters sequentially. |
| `.github/workflows/_osdc-deploy.yml` | Extended with optional `modules:` input (CSV) for per-module deploys, and now derives `aws-region` from `clusters.yaml` instead of hardcoding. Existing callers unchanged. |
| `.github/workflows/osdc-pre-merge.yml` | Path filter updated to include the new workflows so edits to them trigger staging validation. |
| `osdc/justfile` | New `just region <cluster>` recipe (used by `_osdc-deploy.yml` to resolve AWS region). |
| `CODEOWNERS` | Adds owners for `osdc/renovate.json`. |

## What happens if Renovate misbehaves

If `UPDATEBOT_TOKEN` is compromised and the attacker pushes a `renovate-runner/*` branch with an arbitrary diff: the auto-approve refuses (diff guard) and auto-closes the PR with a comment naming the failure mode; even if the guard somehow let it through, the merge queue's staging deploy + smoke would catch a broken image; even if that passed, the prod-deploy `check-trigger` re-validates the diff and refuses to deploy.

If `UPDATEBOT_TOKEN` is compromised and the attacker force-pushes new content between validation and approval: GitHub rejects the SHA-pinned review with 422, and `--match-head-commit` invalidates the auto-merge.

If `GH_PYTORCHBOT_TOKEN` is compromised: it can approve PRs, but it cannot open `renovate-runner/*` PRs (author check uses `OSDC_RENOVATE_BOT_LOGIN`), and any merge it enqueues still has to pass the merge queue's full staging validation, and the post-merge prod-deploy re-validates the diff.

If a runner image regression slips through staging smoke: the uw1 deploy + smoke runs first. A bad image fails uw1 smoke and `deploy_ue2` does not run (uw1 is degraded; ue2 stays on the previous image; human investigation required).

If Renovate proposes a downgrade (e.g., attacker rolling back to a known-vulnerable runner): the monotonicity check in both diff-validation guards rejects it and the auto-approve closes the PR.

## What's not automated

- Approving the **opening** of the pipeline itself (this PR): requires human review and merge as normal.
- Changes to non-runner-image fields in `clusters.yaml`: out of scope for Renovate; still manual.
- Runner-controller image bumps (the ARC controller itself, separate from the runner image): out of scope.
- Cross-cluster rollback: if ue2 fails after uw1 succeeded, rollback is manual.

## Required setup before this can run

- GitHub Environment `osdc-renovate` created on this repo. **MUST NOT** have required reviewers or a wait timer (the pipeline runs unattended on the happy path). Deployment-branch policy should restrict to `main`.
- Environment secrets in `osdc-renovate`:
  - `UPDATEBOT_TOKEN` — the shared `pytorchupdatebot` PAT maintained by the pytorch/test-infra team. Used to push `renovate-runner/*` branches and open PRs.
  - `GH_PYTORCHBOT_TOKEN` — the shared `pytorchbot` PAT maintained by the pytorch/test-infra team. Used to approve PRs, auto-merge, comment, and close. **Must be a different identity than `UPDATEBOT_TOKEN`** — GitHub forbids self-approval.
- Repo variable `OSDC_RENOVATE_BOT_LOGIN` set to `pytorchupdatebot` (the GitHub login that owns `UPDATEBOT_TOKEN`). Any mismatch and every auto-update PR fails the author check.
- Merge queue enabled on `main` with `osdc-pre-merge` listed as a required check.

If `UPDATEBOT_TOKEN` or `GH_PYTORCHBOT_TOKEN` is ever revoked or replaced with a dedicated PAT for this workflow, the minimum required scopes are documented at the top of each workflow file.


.github/workflows/_osdc-deploy.yml
```diff
@@ -52,6 +52,11 @@ on:
         required: false
         type: boolean
         default: false
+      modules:
```
huydhn, line 40: Nit: The change in `_osdc-deploy.yml` seems like a BE change that could go into another PR

.github/workflows/osdc-renovate.yml
```diff
@@ -0,0 +1,69 @@
...
+
+permissions:
+  contents: read
+
+concurrency:
+  group: osdc-renovate
+  cancel-in-progress: false
+
+jobs:
+  renovate:
```
huydhn, line 60: We would need an environment here to protect the `RENOVATE_TOKEN`

.github/workflows/osdc-renovate.yml
```diff
@@ -0,0 +1,69 @@
...
+          RENOVATE_DRY_RUN: ${{ (inputs.dryRun && 'full') || '' }}
+          RENOVATE_REPOSITORIES: ${{ github.repository }}
+          RENOVATE_PLATFORM: github
+          # NOTE: The downstream auto-approve and auto-deploy workflows compare
+          # pull_request.user.login against vars.OSDC_RENOVATE_BOT_LOGIN. That
+          # login is the GitHub identity of RENOVATE_TOKEN (the PAT owner's
+          # username) — NOT the local part of the email below. If you set
+          # OSDC_RENOVATE_BOT_LOGIN to the email's local part, every auto-update
+          # PR will fail the author check.
+          RENOVATE_GIT_AUTHOR: "OSDC Renovate <osdc-renovate@users.noreply.github.com>"
```
huydhn: Should we use the same credential of PyTorch update bot here to avoid the complication of signing CLA for a new bot?  Here is the code reference https://github.com/pytorch/test-infra/blob/main/.github/actions/update-commit-hash/action.yml#L83-L84

.github/workflows/osdc-renovate-autoapprove.yml
```diff
@@ -0,0 +1,165 @@
...
+  group: osdc-renovate-approve-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  approve-and-queue:
+    if: |
+      github.event.pull_request.head.repo.full_name == github.repository &&
+      startsWith(github.event.pull_request.head.ref, 'renovate-runner/') &&
+      contains(github.event.pull_request.labels.*.name, 'auto-runner-update')
+    runs-on: ubuntu-latest
```
huydhn, line 88: Same request to use an environment here to protect `OSDC_APPROVER_TOKEN` secret

huydhn: On PyTorch, we are using PyTorch bot token to approve pin commit update PR https://github.com/pytorch/pytorch/blob/main/.github/workflows/nightly.yml#L178

.github/workflows/osdc-renovate-autoapprove.yml
```diff
@@ -0,0 +1,165 @@
...
+#       - Contents: Read                 (read PR diff)
+#       - Pull requests: Read and write  (submit approving review, enqueue)
+#       - Metadata: Read                 (always required)
+#   - DO NOT grant: Workflows, Administration, Actions, Secrets, or any
+#     organization permissions.
+#   - MUST be a DIFFERENT identity than RENOVATE_TOKEN — GitHub forbids
+#     self-approval, so the approver cannot be the PR author.
+#   - Owner: a dedicated bot account, NOT a human's PAT.
+on:
+  pull_request_target:
```
huydhn: I wonder if you need `pull_request_target` here.  It has a wider blast radius.  Using a workflow_run trigger like https://github.com/pytorch/pytorch/blob/main/.github/workflows/upload-docs-preview.yml#L11-L15 might be an alternative to run the approval workflow after `osdc-renovate` workflow finishes.  The step would roughly be:

1. `osdc-renovate` runs, and create an update PR if there is a new version of runner-image.  Ideally, the workflow knows the PR number here, so that it can store it as an artifact. No new change means no PR and an empty artifact
2. `osdc-renovate-autoapprove` runs once (1) completes.  Note that it runs as a workflow in `ci-infra`, so it has access to `OSDC_APPROVER_TOKEN` secret.  It can check the artifact from (1) to get the PR number and exit early if there is none

This feels safer as we don't need to run `osdc-renovate-autoapprove` on every PRs, also `pull_request_target` is a magnet for security scanners

.github/workflows/osdc-renovate.yml
```diff
@@ -0,0 +1,69 @@
...
+  cancel-in-progress: false
+
+jobs:
+  renovate:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - uses: renovatebot/github-action@693b9ef15eec82123529a37c782242f091365961 # v46.1.14
```
huydhn, line 67: Let's see if this works. Maybe we can modernize our custom pin commit update script https://github.com/pytorch/test-infra/blob/main/.github/scripts/update_commit_hashes.py to use the same  tool.  Although, the difference in that script is that it creates the PR and stamp it in one go using 2 different tokens

.github/workflows/osdc-renovate.yml
jeanschmidt: done :)

.github/workflows/osdc-renovate.yml
jeanschmidt: done :)

.github/workflows/osdc-renovate-autoapprove.yml
jeanschmidt: Done :)

.github/workflows/osdc-renovate-autoapprove.yml
jeanschmidt: yeah, after reconsidering I believe that pull_request_target is not ideal, claude still disagreed with me, but it feel cleaner and more correct with workflow_run

.github/workflows/osdc-renovate.yml
jeanschmidt: something to consider, for sure

.github/workflows/_osdc-deploy.yml
jeanschmidt: Bundled because the auto-deploy workflow is the only consumer and reviewing it in isolation would hide the per-module scoping. Happy to split if you'd prefer.

.github/workflows/osdc-renovate-autoapprove.yml
malfet: Let's not reuse the same token across repos, for every repository create a different one.
And may be in 2026, we can do some sort of OpenID instead of a hardcoded token?

malfet (changes requested):
Why "adding auto-update pipeline" should take 900 lines of code? 
Feels like an amalgamation of 3-4 commits into one PR

jeanschmidt:
> Why "adding auto-update pipeline" should take 900 lines of code? Feels like an amalgamation of 3-4 commits into one PR

makes my life of iterating on this much easier, but I guess you got a valid point.

Let me try to break this down into a stack (would have to be landed at once anyways)

PR #607, opened 2026-05-21
Isolate g5.48xlarge into its own fleet to protect 8-GPU runner

huydhn:
## Summary

Every g5 NodePool (g5.8xlarge, g5.12xlarge, g5.48xlarge, …) currently shares the same `node-fleet=g5` label and taint, so 1-GPU or 4-GPU workflow pods can schedule onto a g5.48xlarge node. Under sustained 1-GPU bursts the 8-GPU SKU becomes GPU-fragmented and starves its runner (`mt-l-x86aavx2-189-704-a10g-8` dropped to 3 active runners with 29 placeholder-workflows Pending on 2026-05-20).

This PR splits `g5.48xlarge` out of the shared fleet so it can only host the 8-GPU runner. 1-GPU and 4-GPU runners stay on the unified `g5` fleet — any fragmentation between those two is **out of scope** for this fix.

## Changes

- `modules/nodepools/defs/g5.yaml` — drop `g5.48xlarge` from the `g5` fleet (other SKUs unchanged).
- `modules/nodepools/defs/g5-48xlarge.yaml` — **new file**, single-instance fleet `g5-48xlarge`. Mirrors the `p4d-24xlarge.yaml` precedent of an isolated GPU SKU getting its own def file.
- `modules/nodepools/scripts/python/generate_nodepools.py` — when a fleet name already encodes family+size (e.g. `g5-48xlarge`), emit the NodePool with that name verbatim instead of `g5-48xlarge-48xlarge`.
- `modules/arc-runners/scripts/python/generate_runners.py` — special-case `g5.48xlarge` in `derive_fleet_name` so the 8-GPU runner's `CAPACITY_AWARE_NODE_FLEET`, node affinity, and toleration values all point at the isolated fleet.
- `test_generate_runners.py` — adjust fleet-name tests.

## Effect

| Runner | Instance | `node-fleet` before | `node-fleet` after |
|---|---|---|---|
| `l-x86aavx2-29-113-a10g`     | g5.8xlarge  | `g5` | `g5` *(unchanged)* |
| `l-x86aavx2-45-167-a10g-4`   | g5.12xlarge | `g5` | `g5` *(unchanged)* |
| `l-x86aavx2-189-704-a10g-8`  | g5.48xlarge | `g5` | `g5-48xlarge` |

g5.48xlarge nodes become reserved for 8-GPU jobs. 1-GPU and 4-GPU pods can no longer land on them.

NodePool resource names stay the same (`g5-48xlarge.yaml`, `g5-8xlarge.yaml`, etc.); only the `node-fleet` label/taint *value* changes — and only for `g5.48xlarge`.

## Migration is no-restart

Existing g5.48xlarge nodes keep their old `node-fleet=g5:NoSchedule` taint. New 8-GPU workflow pods only tolerate `node-fleet=g5-48xlarge`, so:

- In-flight 8-GPU jobs run to completion on the existing nodes.
- New 8-GPU scheduling decisions go to newly-provisioned g5.48xlarge nodes that carry the new label.
- Old g5.48xlarge nodes drain naturally and Karpenter consolidates them via `WhenEmpty` once idle.

No pod restarts, no scale-set bouncing.

## Test plan

- [x] `just lint` — all 13 linters pass.
- [x] `just test` — all unit tests pass (98.81% coverage).
- [x] Manual: ran `generate_nodepools.py` and confirmed the new `g5-48xlarge` NodePool is produced with `node-fleet=g5-48xlarge` label/taint, while `g5-8xlarge`, `g5-12xlarge`, `g5-16xlarge`, `g5-24xlarge` keep `node-fleet=g5`.
- [x] Manual: ran `just generate-arc-runners arc-cbr-production` and verified `l-x86aavx2-189-704-a10g-8` emits `CAPACITY_AWARE_NODE_FLEET=g5-48xlarge` while the other two a10g runners still emit `g5`.
- [ ] Staging soak before prod.
- [ ] Monitor 8-GPU scale set after rollout — expect `Pending` placeholder-workflow count on `mt-l-x86aavx2-189-704-a10g-8` to drop to ~0 once the queue rolls over.

## Follow-ups (out of scope)

- Same isolation for 4-GPU A10G (`g5.12xlarge`) if 1-GPU pressure starts hurting that runner.
- Same pattern for g6 (l4) and g4dn (t4) families if fragmentation surfaces there.
- Longer-term bin-packing scheduler profile (`MostAllocated` scoring) — separate PR.

### DIFF
--- a/osdc/modules/arc-runners/scripts/python/generate_runners.py
+++ b/osdc/modules/arc-runners/scripts/python/generate_runners.py
@@ -54,14 +54,21 @@ def normalize_name(name):
     return name.replace(".", "-").replace("_", "-")
 
 
-def derive_fleet_name(instance_type):
-    """Derive the node-fleet name from an instance type.
+def derive_fleet_name(instance_type, fleet_map=None):
+    """Derive the node-fleet name for an instance type.
 
-    Returns the instance family (everything before the dot).  GPU scheduling
-    is handled by nvidia.com/gpu resource requests, not fleet-level isolation,
-    so GPU and CPU instances use the same derivation: family name only.
+    Consults ``fleet_map`` (instance_type -> fleet_name, populated from
+    modules/nodepools/defs/*.yaml) first, so a runner whose instance_type
+    sits in a split-out fleet (e.g. g5.48xlarge in g5-48xlarge.yaml) picks
+    up that fleet's label automatically — no per-instance hardcoding here.
+
+    Falls back to the instance family (everything before the dot) when the
+    map is empty or missing the instance, matching the historical default
+    used before per-SKU fleets existed (r7a, g5, c7i, p6-b200, …).
     """
-    return instance_type.split(".")[0]  # r7a, g5, c7i, p6-b200, etc.
+    if fleet_map and instance_type in fleet_map:
+        return fleet_map[instance_type]
+    return instance_type.split(".")[0]
 
 
 # Kubernetes resource quantity suffixes → multiplier (bytes)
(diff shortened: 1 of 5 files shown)


jeanschmidt:
As this PR is from a forked repo, I'll open a new one in ci-infra to make those changes.

considerations:
* Larger instances are more efficient for pod placement while maintaining the cost/resources ratio. So we don't want to exclude them from smaller pods placement, and we can cause similar problems on a 2nd level if we do so.

huydhn:
> As this PR is from a forked repo, I'll open a new one in ci-infra to make those changes.

Please go ahead.

>     * Larger instances are more efficient for pod placement while maintaining the cost/resources ratio. So we don't want to exclude them from smaller pods placement, and we can cause similar problems on a 2nd level if we do so.

Make sense, I was debating with myself yesterday about this and decided to go with a minimal mitigation PR first and just moving 48xlarge in its own pool looks simpler, not necessarily the most efficient one

PR #610, opened 2026-05-21
Exclude Graviton-2 (arm64g2) runners from load tests

jeanschmidt:
**Impact:** Load test distribution only — no production runner changes **Risk:** low

## What
Removes Graviton-2 (`arm64g2`) ARM runners from load test job distribution due to persistent capacity issues on those instance types.

## Why
Graviton-2 instances (`linux.arm64.2xlarge` / `linux.arm64.2xlarge.ephemeral`) constantly hit capacity limits, causing load test failures. Newer Graviton generations (g3, g4) remain available and cover the ARM64 workload.

## How
- Rather than just removing the g2 entries from the label map, adds a regex-based exclusion pattern (`_LOAD_TEST_EXCLUDED_ARM_PATTERN`) that filters g2 runners at scan time — same approach used for high-end GPU exclusions
- This catches any g2 runner definition (including hypothetical baremetal variants) without needing to maintain an explicit denylist

## Changes
- Remove `linux.arm64.2xlarge` and `linux.arm64.2xlarge.ephemeral` from `OLD_TO_OSDC_LABEL` mapping and `PRODUCTION_JOB_COUNTS`
- Add `_LOAD_TEST_EXCLUDED_ARM_PATTERN` regex matching `arm64g2` runner names
- Apply the new pattern alongside the existing GPU exclusion in `get_available_runners()`
- Add 3 tests: g2 exclusion, baremetal g2 exclusion, and a negative test confirming g3/g4 are not affected

## Testing
- `pytest test_distribution.py` — all existing and new tests pass

### DIFF
--- a/osdc/integration-tests/load-test/scripts/python/distribution.py
+++ b/osdc/integration-tests/load-test/scripts/python/distribution.py
@@ -63,8 +63,6 @@ OLD_TO_OSDC_LABEL: dict[str, str] = {
     "linux.g6.4xlarge.experimental.nvidia.gpu": "l-x86aavx2-29-113-l4",
     "linux.g6.12xlarge.nvidia.gpu": "l-x86aavx2-45-172-l4-4",
     # ARM64 (Graviton)
-    "linux.arm64.2xlarge": "l-arm64g2-6-32",
-    "linux.arm64.2xlarge.ephemeral": "l-arm64g2-6-32",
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM! @malfet mentions that we don't care about g2 nowadays, it looks like there are some places still referring to it

PR #611, opened 2026-05-22
Add node_fleet override to decouple large-instance runners from shared node fleets

jeanschmidt:
* #618
* #617
* #612
* __->__ #611

**Impact:** Runner generation, load-test distribution, node utilization analysis
**Risk:** low

## What
Introduces an optional `node_fleet` field on runner defs that overrides the default instance-family-based fleet derivation, and adds a `runner_overhead` loader that reads actual pod resource requests from rendered YAMLs instead of hardcoded constants.

## Why
Large workflow pods (e.g. g5.48xlarge) sharing a node fleet with smaller pods of the same instance family (e.g. g5.8xlarge) compete for node resources. The big pods queue for extended periods because Karpenter packs the fleet with smaller instances first. Giving large runners their own dedicated node fleet lets Karpenter provision appropriately-sized nodes without contention.

## How
- Extracted `derive_fleet_name` and `validate_node_fleet` into a shared `fleet_naming` module so both `generate_runners.py` and `distribution.py` use the same derivation logic (single source of truth)
- Fleet name derivation accepts an optional `override` parameter; when set, it bypasses the instance-family split and uses the override verbatim
- Validation enforces DNS-1123 label rules and blocks reserved names (`c7i-runner`) to prevent workflow pods from landing on the runner control-plane pool
- `runner_overhead.py` parses rendered Helm values + ConfigMap YAMLs to extract runner/listener/workflow resource requests, replacing hardcoded overhead constants

## Changes
- `scripts/python/fleet_naming.py` — new module: `derive_fleet_name(instance_type, override=None)` and `validate_node_fleet(value)` with DNS-1123 + reserved-name checks
- `scripts/python/runner_overhead.py` — new module: `RunnerPodOverhead` dataclass and `load_runner_pod_overhead()` loader that reads rendered runner YAMLs
- `scripts/python/conftest.py` — shared pytest fixtures (lru_cache clearing, `make_test_overhead` helper)
- `modules/arc-runners/scripts/python/generate_runners.py` — reads `node_fleet` from runner defs, validates it, passes override to `derive_fleet_name`; removes the now-extracted local `derive_fleet_name` function
- `integration-tests/load-test/scripts/python/distribution.py` — imports shared `fleet_naming`, honors `node_fleet` override for region-exclusion checks
- `scripts/python/test_fleet_naming.py` — 187-line test suite covering valid labels, invalid formats, type checking, and reserved names
- `scripts/python/test_runner_overhead.py` — 370-line test suite covering YAML parsing, workflow-extra clamping, disagreement warnings, and error-message context
- `modules/arc-runners/scripts/python/test_generate_runners.py` — added node_fleet validation tests (empty, whitespace, newline, quote, uppercase, too-long, slash, reserved) and an override variant to the real-template render suite
- `integration-tests/load-test/scripts/python/test_distribution.py` — added node_fleet override/exclusion/fallback/validation tests

## Testing
- `just test` — all unit tests pass including the new fleet_naming, runner_overhead, generate_runners, and distribution test suites
- `just lint` — all 13 linters pass


osdc/scripts/python/fleet_naming.py
```diff
@@ -0,0 +1,52 @@
...
+# pod to tolerate it would let user code preempt the orchestrator and break the
+# priority-class ladder.
+RESERVED_NODE_FLEET_NAMES = frozenset({"c7i-runner"})
+
+# K8s DNS-1123 label: lowercase alphanumeric and dashes, start/end alphanumeric,
+# 1-63 chars. Matches the constraints applied at apply time by the API server.
+_NODE_FLEET_PATTERN = re.compile(r"^[a-z0-9]([-a-z0-9]{0,61}[a-z0-9])?$")
+
+
+def validate_node_fleet(value):
```
huydhn, line 24: Is there a away to guard against the name here to match only the fleet names from `nodepools/defs`?

osdc/modules/arc-runners/scripts/python/generate_runners.py
```diff
@@ -221,14 +216,20 @@ def generate_runner(def_file, template_content, cluster_config, output_dir, modu
...
+
     if max_burst_capacity > 0 and proactive_capacity > 0 and max_burst_capacity < proactive_capacity:
         log_warning(
             f"In {def_file}: max_burst_capacity ({max_burst_capacity}) < proactive_capacity ({proactive_capacity}); "
             f"the cap will prevent the listener from reaching its proactive baseline. "
             f"Consider raising max_burst_capacity."
         )
 
-    node_fleet = derive_fleet_name(instance_type)
+    node_fleet = derive_fleet_name(instance_type, override=node_fleet_override)
```
huydhn: Do we need to import `validate_node_fleet` here for `validate_node_fleet` → `derive_fleet_name` call pattern? It looks easier to just let `derive_fleet_name` calls `validate_node_fleet` and check for the `node_fleet_override` there.  Then both the generate_runners script and distribution test will just need to call `derive_fleet_name`

osdc/modules/arc-runners/scripts/python/generate_runners.py
```diff
@@ -23,13 +23,17 @@
...
+# instance_specs and fleet_naming live in scripts/python/ at the repo root.
+# Add it to sys.path so the import works both when run directly (deploy.sh)
+# and when run via pytest (pyproject.toml testpaths also adds it).
 _scripts_python = str(Path(__file__).resolve().parents[4] / "scripts" / "python")
 if _scripts_python not in sys.path:
     sys.path.insert(0, _scripts_python)
 
+from fleet_naming import derive_fleet_name, validate_node_fleet  # noqa: E402
+
+__all__ = ["derive_fleet_name"]
```
huydhn: Just curious, what is this line for? My read here is that generate_runners re-export `derive_fleet_name`, but it looks weird

osdc/scripts/python/fleet_naming.py
huydhn: Basically, having a typo in `node_fleet: g5-xlarge` feels worst than not setting it because the latter will fallback to the g5 family while having the wrong fleet name would just pending forever

osdc/scripts/python/runner_overhead.py
```diff
@@ -0,0 +1,262 @@
+#!/usr/bin/env python3
```
huydhn: As you're already using ghstack, moving the changes for `runner_overhead` and `conftest` to a different PR would be nice.  AFAICT, we don't need them here to support `node_fleet`

osdc/scripts/python/fleet_naming.py
jeanschmidt: no, because modules are independent, unless we do some sophisticated check cluster-by-cluster loading all nodepools and all runner defined in all modules and cross-check.

This would be a complex script that we would have to maintain.

At best this would be a separate task IMO

osdc/modules/arc-runners/scripts/python/generate_runners.py
jeanschmidt: done

osdc/modules/arc-runners/scripts/python/generate_runners.py
jeanschmidt: yeah, mild nicer, less explicit, but I can make the change

osdc/scripts/python/runner_overhead.py
jeanschmidt: ok, creating a pr in this stack...

osdc/scripts/python/fleet_naming.py
jeanschmidt: nah, changed my mind, I'll add this validation on another PR.

huydhn (approved):
LGTM!

PR #612, opened 2026-05-22
Add dedicated nodepool fleets for large-instance runners

jeanschmidt:
* #618
* #617
* __->__ #612
* #611

**Impact:** Large-instance ARC runners (bare-metal and 48xlarge) across all clusters
**Risk:** medium

## What
Creates 8 dedicated Karpenter NodePool definitions for large instances and wires each large-instance runner to its own fleet via the `node_fleet` override.

## Why
Large runners that consume an entire node (bare-metal, 48xlarge) were sharing Karpenter NodePools with small runners that pack multiple pods per node. This coupling causes scheduling and scaling interference — Karpenter consolidation and disruption decisions for densely-packed small runners affect the whole-node large runners and vice versa. Dedicated fleets isolate the two scaling patterns so each can be tuned independently.

## How
- Each new `-large` nodepool is a name-only subset of its parent fleet, restricted to only the largest instance types
- Isolation is enforced by the existing `node-fleet=<name>` taint mechanism — only runners with a matching `node_fleet` override tolerate the taint
- GPU pools (g4dn-large, g5-large, p4d-large) carry forward NVME, baremetal, and NUMA topology settings from their parent definitions
- Builds on the `node_fleet` override field added in b4b8fef

## Changes
- Added `node_fleet` override to 8 large-instance runner definitions:
  - `l-barm64g3-62-226` → `m7g-large`
  - `l-bx86iavx512-88-1000-a100-8` → `p4d-large`
  - `l-bx86iavx512-94-344-t4-8` → `g4dn-large`
  - `l-x86aavx2-189-704-a10g-8` → `g5-large`
  - `l-x86iavx512-94-192` → `c7a-large`
  - `l-x86iavx512-94-768` → `r7a-large`
- Created 8 new nodepool definitions in `modules/nodepools/defs/`:
  - `c7a-large` (c7a.48xlarge)
  - `g4dn-large` (g4dn.metal, baremetal)
  - `g5-large` (g5.48xlarge, NVME)
  - `m7g-large` (m7g.metal, arm64, baremetal)
  - `p4d-large` (p4d.24xlarge, single-numa-node topology, baremetal)
  - `r7a-large` (r7a.48xlarge)

## Notes
- Completes the work started in b4b8fef which added the `node_fleet` override mechanism
- The `c7i-large` and `m8g-large` pools include multiple instance sizes as fallback options; other pools are single-instance-type
- `p4d-large` is the only pool with explicit `topology_manager_policy: single-numa-node` to prevent NUMA fragmentation on multi-NUMA GPU nodes

## Testing
- Run `just lint` and `just test` to validate YAML structure and naming conventions
- After deploy: verify new NodePools appear in `kubectl get nodepools` and carry the correct `node-fleet` taint
- Confirm large runners schedule onto their dedicated pools (check pod node affinity and tolerations)
- Monitor that small runners on the parent fleets are unaffected


osdc/modules/nodepools/defs/m7g-large.yaml
```diff
@@ -0,0 +1,11 @@
+# Karpenter NodePool fleet: m7g-large — dedicated pool for large m7g runners.
+# Name-only subset of m7g; separation is enforced by the node-fleet=m7g-large taint.
+fleet:
+  name: m7g-large
```
huydhn: I would prefer to call this `m7g-metal` to make it clearer that it's a meta runner.  There are `m7g.16xlarge` from https://aws.amazon.com/ec2/instance-types/m7g/ what has a "similar" spec with the same number of CPU and memory, but it's not a metal box.  The runner name `l-barm64g3-62-226` also means metal runner with `b`

osdc/modules/nodepools/defs/g4dn-large.yaml
```diff
@@ -0,0 +1,13 @@
+# NUMA note: restricted topology policy may reject mixed GPU packing; monitor for TopologyAffinityError
+# Karpenter NodePool fleet: g4dn-large — dedicated pool for the 8-GPU T4 runner.
+# Name-only subset of g4dn; separation is enforced by the node-fleet=g4dn-large taint.
+fleet:
+  name: g4dn-large
```
huydhn: Same commend goes here, should we call this `g4dn-metal` to make it clearer that this is a metal one

huydhn (approved):
Mostly LGTM! But I suggest we call metal nodepools as they are, i.e. `m7g-metal` looks clearer than `m7g-large` to me, less need to do mental mapping of different runner types.

osdc/modules/nodepools/defs/m7g-large.yaml
jeanschmidt: agree

osdc/modules/nodepools/defs/g4dn-large.yaml
jeanschmidt: agree :)

PR #616, opened 2026-05-22
Serialize tofu operations per cluster and add lock-timeout to prevent DynamoDB lock races

jeanschmidt:
**Impact:** CI pipelines (plan + deploy workflows) and local just recipes **Risk:** low

## What
Adds GitHub Actions concurrency groups to serialize tofu plan and deploy jobs per cluster, and sets a 15-minute lock-timeout on every tofu plan/apply/destroy invocation to tolerate queued operations instead of failing immediately on a held DynamoDB state lock.

## Why
Parallel CI runs (e.g. concurrent PRs triggering plan, or a plan racing a deploy) against the same cluster contend on the shared DynamoDB state lock. Without serialization, the second job hits the lock and fails immediately. This creates flaky CI failures and requires manual reruns.

## How
- Used a shared concurrency group key (`osdc-tofu-${{ inputs.cluster }}`) across both reusable workflows so plan and deploy against the same cluster are mutually exclusive
- Set `cancel-in-progress: false` to queue rather than cancel waiting jobs
- Bumped plan workflow timeout from 30m to 45m to accommodate queuing delays
- Applied `-lock-timeout=15m` uniformly to all tofu plan, apply, and destroy calls rather than relying on instant lock acquisition

## Changes
- `.github/workflows/_osdc-deploy.yml`: add per-cluster concurrency group to the deploy job
- `.github/workflows/_osdc-plan.yml`: add matching concurrency group to the plan job; increase timeout from 30m to 45m
- `osdc/justfile`: add `-lock-timeout=15m` to all 6 tofu plan/apply invocations across `deploy-base`, `deploy-module`, and `plan` recipes
- `osdc/scripts/destroy-cluster.sh`: add `-lock-timeout=15m` to both `destroy_module` and `destroy_base` tofu destroy calls

## Testing
- Open two PRs targeting the same cluster and verify the plan jobs queue instead of racing
- Trigger a deploy while a plan is in progress and confirm the deploy waits for the plan to release the lock
- Verify `just plan <cluster>` and `just deploy <cluster>` pass locally with the new flag

### DIFF
--- a/.github/workflows/_osdc-deploy.yml
+++ b/.github/workflows/_osdc-deploy.yml
@@ -63,6 +63,12 @@ jobs:
     runs-on: ubuntu-latest
     environment: ${{ inputs.environment }}
     timeout-minutes: 60
+    # Serialize all tofu state-mutating work per cluster. Shared with
+    # _osdc-plan.yml so plan and deploy against the same cluster never
+    # race on the DynamoDB state lock.
+    concurrency:
+      group: osdc-tofu-${{ inputs.cluster }}
+      cancel-in-progress: false
     defaults:
       run:
         working-directory: osdc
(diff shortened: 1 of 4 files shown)


huydhn (approved):
Ty for the fix!

PR #617, opened 2026-05-22
Add runner_overhead loader for rendered runner YAMLs

jeanschmidt:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #618
* __->__ #617
* #612
* #611

**Impact:** scripts/python/ test infrastructure (no production callers yet)
**Risk:** low

## What
Adds a `runner_overhead` module that parses rendered Helm values + ConfigMap YAMLs to extract runner / listener / workflow-pod resource requests, plus a shared pytest `conftest.py` that clears the loader's `lru_cache` between tests and exposes a `make_test_overhead` helper.

## Why
`analyze_node_utilization.py` currently hardcodes `HOOKS_OVERHEAD_CPU_M` / `HOOKS_OVERHEAD_MEM_MI` constants that drift from the actual values shipped in the generated runner YAMLs. The loader replaces those constants with values read from the source of truth (the rendered YAMLs that get applied to the cluster). Lands as its own change so the node_fleet override PR ahead of it does not have to carry an unrelated module.

## Changes
- `scripts/python/runner_overhead.py` - new module: `RunnerPodOverhead` dataclass + `load_runner_pod_overhead()` loader that walks `modules/arc-runners*/generated/*.yaml`, picks the first file as source of truth, and warns on disagreements
- `scripts/python/conftest.py` - shared pytest fixtures (autouse `lru_cache` clearing, `make_test_overhead` helper)
- `scripts/python/test_runner_overhead.py` - test suite covering YAML parsing, workflow-extra clamping, disagreement warnings, and error-message context

## Testing
- `just lint` - all 13 linters pass
- Unit tests for `runner_overhead` pass (two pre-existing `TestParseCpu` failures live in `daemonset_overhead.parse_cpu_millicores` and are unrelated to this change)

## Follow-up
A subsequent change will wire `load_runner_pod_overhead()` into `analyze_node_utilization.py` and delete the hardcoded `HOOKS_OVERHEAD_*` constants.

Signed-off-by: Jean Schmidt <contato@jschmidt.me>

PR #618, opened 2026-05-22
Validate node_fleet references against enabled NodePool definitions

jeanschmidt:
* __->__ #618
* #617
* #612
* #611

**Impact:** PR-time validation gate + render-time generator hook
**Risk:** low — additive checks, hard-fails on misconfiguration

## What
Adds a validator that, for every cluster in clusters.yaml, asserts each arc-runner def's effective node_fleet (override or instance-family fallback) maps to a fleet defined by that cluster's enabled nodepool* modules. Runs as part of `just test` AND as a render-time hook in generate_runners.py that aborts generation on validation failure.

## Why
A typo like `node_fleet: g5-xlarge` (real fleet is `g5-large`) silently passes today. The runner def renders, the workflow pod's nodeAffinity targets a non-existent fleet, and jobs pend forever waiting for Karpenter to provision a node that will never appear. Catching this at PR time avoids the "runner pending in production" debug loop.

## How
- Extracted `is_excluded_for_region`, `load_excluded_instance_types`, and `iter_fleet_names` into a new shared module `scripts/python/nodepool_defs.py` (eliminates duplication between generate_nodepools.py, generate_runners.py, and the validator)
- New `runner_fleet_validator.py` walks every nodepools* module the cluster enables to build the available-fleets set, filters region-excluded fleets, and validates each arc-runners* def against it
- Runners whose instance_type is in `excluded_instance_types` for the region are skipped (max_runners=0 by design — not orphans)
- Reserved names (c7i-runner) are excluded from the available set so workflow pods can't override into the runner-pod control-plane fleet
- Generator hook runs BEFORE `shutil.rmtree(output_dir)` so a validation failure preserves the last-known-good rendered YAMLs
- Clusters with arc-runners but no nodepools modules hard-fail with an explicit error

## Changes
- `scripts/python/nodepool_defs.py` - new shared module (122 lines): is_excluded_for_region, load_excluded_instance_types, iter_fleet_names
- `scripts/python/runner_fleet_validator.py` - new validator (200 lines): validate_cluster_runner_fleets, collision detection, fleet collection
- `scripts/python/test_runner_fleet_mapping.py` - 19 unit tests + parametrized live check over clusters.yaml; uses pytest_generate_tests so malformed clusters.yaml skips gracefully without blocking unit tests
- `scripts/python/test_nodepool_defs.py` - 27 tests covering all three shapes (fleet, fleets, legacy nodepool) and the combined-shape case
- `modules/arc-runners/scripts/python/generate_runners.py` - calls validator before rmtree; sys.exit(1) on failure; consumer_root via OSDC_ROOT env var only (no path walk-up)
- `modules/arc-runners/scripts/python/test_generate_runners.py` - fixtures updated to include matching nodepool defs so main() exercises the new validator hook
- `modules/nodepools/scripts/python/generate_nodepools.py` - imports is_excluded_for_region from nodepool_defs

## Testing
- `just test` - all unit tests pass (two pre-existing test_runner_overhead.py::TestParseCpu failures unrelated to this change)
- `just lint` - all 13 linters pass
- Validator against live clusters.yaml: 0 errors for all 3 clusters
- Injected typo verified: generator exits non-zero, output dir preserved

### DIFF
--- a/osdc/scripts/python/nodepool_defs.py
+++ b/osdc/scripts/python/nodepool_defs.py
@@ -1,0 +1,122 @@
+"""Shared parsers for nodepool def YAMLs.
+
+Used by:
+- ``modules/nodepools/scripts/python/generate_nodepools.py`` (the canonical
+  consumer — generates NodePool manifests from these defs)
+- ``modules/arc-runners/scripts/python/generate_runners.py`` (filters runners
+  whose backing nodepool is excluded for the current region)
+- ``scripts/python/runner_fleet_validator.py`` (cross-checks runner defs
+  against the fleet names produced by these defs)
+
+A def YAML uses exactly one of three shapes — ``fleet`` (single fleet),
+``fleets`` (list of fleets), or legacy ``nodepool`` (single instance type).
+The helpers here parse each shape uniformly.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import yaml
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+
+def is_excluded_for_region(def_data: dict, region: str) -> bool:
+    """Return True if ``region`` appears in the def's ``exclude_regions`` list.
+
+    Returns False when ``region`` is empty/None or the def has no
+    ``exclude_regions`` key — keeps callers backward-compatible when region is
+    not supplied.
+    """
+    if not region:
+        return False
+    if not isinstance(def_data, dict):
+        return False
+    return region in (def_data.get("exclude_regions") or [])
+
+
+def load_excluded_instance_types(defs_dir: Path, region: str) -> set[str]:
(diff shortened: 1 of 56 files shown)


osdc/scripts/python/nodepool_defs.py
```diff
@@ -0,0 +1,122 @@
...
+    empty set when ``region`` is falsy or the directory is missing.
+    """
+    excluded: set[str] = set()
+    if not region or not defs_dir.is_dir():
+        return excluded
+    for def_file in sorted(defs_dir.glob("*.yaml")):
+        try:
+            data = yaml.safe_load(def_file.read_text()) or {}
+        except yaml.YAMLError:
+            continue
```
huydhn, line 57: Nit: No warning here for broken YAML files?

huydhn (approved):
LGTM!

PR #620, opened 2026-05-23
Create a common Grafana dashboard

jathu:
* Creates a common area for all Grafana dashboards shared by the team. This will be published to: [fburl.com/pytorch-metrics](https://fburl.com/pytorch-metrics)
   * Allows a way to test changes in a custom folder
   * Workflow on main will push to the real folder
* Include the initial dashboard to track key developer experience metrics

## Test Plan

Pushed to my test folder: https://pytorchci.grafana.net/dashboards/f/flrhzs

```sh
$ cd grafana
$ GRAFANA_TOKEN=... mise run publish --folder flrhzs
```

### DIFF
--- a/.github/workflows/grafana-publish.yml
+++ b/.github/workflows/grafana-publish.yml
@@ -1,0 +1,37 @@
+name: "Grafana: Publish dashboards"
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+    paths:
+      - grafana/**
+      - .github/workflows/grafana-publish.yml
+
+permissions:
+  contents: read
+
+concurrency:
+  group: grafana-publish
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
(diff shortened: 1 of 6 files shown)


zxiiro (approved):
> Creates a common area for all Grafana dashboards shared by the team. This will be published to: [fburl.com/pytorch-metrics](https://fburl.com/pytorch-metrics)

How do non-meta employees get access to this? is the https://pytorchci.grafana.net the method? if so how would folks get access to it?

grafana/generator.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
```
huydhn: We are trying to move away from shell script in CI, so let's rewrite this in Python

huydhn (approved):
LGTM! Let's use Python for the generator script

jathu:
@zxiiro that url is just an alias for https://pytorchci.grafana.net/dashboards/f/fvnzj9. If you don't have access to this, let's chat this week to see what we can do

PR #622, opened 2026-05-26
Add us-east-1 OSDC cluster and workflows

georgehong:
Adds `us-east-1` and associated workflows. For initial bootstrapping, we need to initialize the bucket and rerun the workflows.
```
just bootstrap arc-cbr-production-ue1
```
- Checked cidr can match current prod.
- Checked EC2 instance types are launchable. 

For deployment, from local machine, we'll be using the following setting to avoid diverting workflows while the region is initially provisioning. 
```
    region: us-east-1
    + pause_runners: true
    state_bucket: ciforge-tfstate-arc-cbr-prod-ue1
```

### DIFF
--- a/osdc/clusters.yaml
+++ b/osdc/clusters.yaml
@@ -227,6 +227,43 @@ clusters:
       - monitoring
       - logging
 
+  meta-prod-aws-ue1:
+    cluster_name: meta-prod-aws-ue1
+    region: us-east-1
+    state_bucket: ciforge-tfstate-arc-cbr-prod-ue1
+    access_config:
+      authentication_mode: API_AND_CONFIG_MAP
+      cluster_admin_role_names:
+        - osdc_gha_prod
+    base:
+      vpc_cidr: "10.4.0.0/16"
+    git_cache:
+      central_replicas: 15
+      central_cpu_request: "10"
+      central_cpu_limit: "15"
+      central_memory_request: "10Gi"
+      central_memory_limit: "50Gi"
+    node_compactor:
+      min_node_age_seconds: 900
+    arc-runners:
+      github_config_url: "https://github.com/pytorch"
+      github_secret_name: meta-prod-aws-ue1
+      runner_name_prefix: "mt-"
+      runner_group: "meta-prod-aws-ue1"
+    pypi_cache:
+      replicas: 10
+    modules:
+      - karpenter
+      - arc
+      - nodepools
+      - arc-runners
+      - pypi-cache
+      - cache-enforcer
+      - zombie-cleanup
+      - harbor-cache-recovery
+      - monitoring
+      - logging
+
   arc-cbr-production:
     cluster_name: pytorch-arc-cbr-production
     region: us-east-2
--- a/.github/workflows/osdc-deploy-prod.yml
+++ b/.github/workflows/osdc-deploy-prod.yml
@@ -6,10 +6,9 @@ name: "OSDC: Deploy production"
 # `environment:osdc-production` in the sub claim, so IAM enforces the gate too.
 #
 # Default behavior (target = all): sequential rollout.
-#   1. Deploy arc-cbr-production-uw1.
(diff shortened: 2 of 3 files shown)


osdc/clusters.yaml
```diff
@@ -227,6 +227,49 @@ clusters:
...
+      min_node_age_seconds: 900
+    buildkit:
+      arm64_instance_type: c7gd.16xlarge
+      amd64_instance_type: m6id.24xlarge
+      pods_per_node: 2
+    arc-runners:
+      github_config_url: "https://github.com/pytorch"
+      github_secret_name: pytorch-arc-cbr-production
+      runner_name_prefix: "mt-"
+      runner_group: "arc-cbr-prod-ue1"
```
georgehong: I've added this to match the pattern used for uw1. I think we'll need to add this to the repo settings?

georgehong:
Remove `pause_runner: True` setting that entered the clusters.yaml file. This should be local-only for the first manual deployment.

georgehong:
Update names to use `meta-prod-aws-ue1` wherever possible. Left unchanged are:
- `runner_name_prefix` due to requiring an exact match, which would be involved
- `state_bucket` as discussed offline.

zxiiro (approved):

jeanschmidt (approved):

huydhn (approved):

huydhn:
Btw, here is the `pause_runners` flag @jeanschmidt introduces to run the deployment but not accept any traffic https://github.com/pytorch/ci-infra/blob/main/osdc/modules/arc-runners/scripts/python/generate_runners.py#L173

When I did the us-west-1 deployment, I did it in 2 phase:

0. Merge this change, update local checkout to latest main
1. Run a local deployment with `pause_runners` set manually to true, which brought the cluster online. Check that everything looks good and wait for an hour to warm up all the cache
2. Trigger CI deployment without the parameter to start getting traffic

osdc/clusters.yaml
```diff
@@ -227,6 +227,48 @@ clusters:
...
+      runner_name_prefix: "mt-"
+      runner_group: "meta-prod-aws-ue1"
+    pypi_cache:
+      replicas: 10
+    modules:
+      - karpenter
+      - arc
+      - nodepools
+      - arc-runners
+      - buildkit
```
huydhn, line 264: Let's remove Buildkit here because:

1. It's not used with the `runner_group` override set to `meta-prod-aws-ue1`.  For the context, to use Buildkit, we need to use a dedicated `runner_group` called `release-runners` for [x86](https://github.com/pytorch/ci-infra/blob/main/osdc/modules/arc-runners/defs/rel-l-x86iavx512-8-64.yaml#L11) and [aarch64](https://github.com/pytorch/ci-infra/blob/main/osdc/modules/arc-runners/defs/rel-l-arm64g4-16-62.yaml#L11).  And with the above override, there will not be any `release-runners` group in ue1, thus nothing can call the Buildkit cluster on that regions anyway.  https://github.com/pytorch/ci-infra/pull/609 removed it on uw1 accordingly.
2. $$$ consideration.  The cluster has fixed capacity and costs 1K daily.  Issue #72 from the issue tracker doc is to find out a way to auto-scale it.

So, removing Buildkit is sufficient to unblock the deployment here.  On the other hand, I think we have an issue to fix when we rename the current default cluster to `meta-prod-aws-ue2` because the current logic says that those release runners will go to `meta-prod-aws-ue2`, which is wrong.  Let me add this note to the rename task.

georgehong: Thanks for that context! Both the listed buildkit item and its configuration block have now been removed in the most recent update to the PR.

georgehong:
Proceeding to merge. The `tofu plan` warning is expected since an initial manual deployment is necessary, and as Huy's above instructions mention, we'll be overriding with:
```
pause_runners: true
```

PR #624, opened 2026-05-26
Add GitHub Actions secrets inventory

zxiiro:
Documents all environment and repository secrets with who added them, who requested them, which workflows consume them, and PoC contacts.


jordanconway (approved):
Great resource!

PR #630, opened 2026-05-26
Allow per-cluster max_runners overrides for arc-runners-X scale sets

huydhn:
`arc-runners-X/defs/*.yaml` now own per-cluster capacity directly. `max_runners` accepts either an int (baseline applied to every cluster — backwards-compatible with the ~30 existing CPU/GPU defs and the B200 defs) or a mapping:

```yaml
max_runners:
  default: 8                  # required baseline
  arc-cbr-production-uw1: 48  # per-cluster override
```

Each `arc-runners-h100` def adopts the mapping form. `arc-cbr-production-uw1` inherits the 6 H100 nodes freed by pytorch-gha-infra#1180, so its scale sets render at def×6 (48 / 24 / 12 / 6). us-east-2 keeps the def baselines (unchanged).

Capacity stays co-located with the runner shape rather than split between `clusters.yaml` and the def — addresses [@jeanschmidt's review feedback](https://github.com/pytorch/ci-infra/pull/630#discussion_r3306485322) on the earlier `clusters.yaml`-based approach.

## Test plan
- [x] `pytest modules/arc-runners/scripts/python/test_generate_runners.py` — 5 new tests covering the mapping form, plus all existing tests green.
- [x] End-to-end render against both clusters confirms `maxRunners`: 48 / 24 / 12 / 6 (uw1) and 8 / 4 / 2 / 1 (us-east-2 baseline).
- [ ] Deploy `arc-runners-h100` to `arc-cbr-production-uw1`; confirm `maxRunners` = 48 / 24 / 12 / 6.
- [ ] Deploy `arc-runners-h100`/`-b200` to `arc-cbr-production`; confirm `maxRunners` is unchanged.


osdc/clusters.yaml
```diff
@@ -211,6 +211,15 @@ clusters:
       capacity_reservation_ids:
         - cr-04d3d1d84e127a562
         - cr-09a53051589034fb8
+    arc-runners-h100:
+      # arc-runners-h100/defs/*.yaml are sized for 1 reserved 8-GPU node, but
+      # this cluster holds 6 (see nodepools-h100.capacity_reservation_ids above).
+      # Override each scale set's max_runners to def×6.
+      max_runners:
```
jeanschmidt: IMO max runners should be defined in the nodepool, 

Can't we do this in the nodepool and duplicate them? Maybe it makes more sense

huydhn: Let me try that, and see how it looks like. I think I know what you mean

huydhn: @jeanschmidt Wdyt of this new syntax?

```
max_runners:
  default: 1                  # required baseline
  arc-cbr-production-uw1: 2  # per-cluster override
```

osdc/modules/arc-runners/scripts/python/generate_runners.py
```diff
@@ -166,9 +166,30 @@ def generate_runner(def_file, template_content, cluster_config, output_dir, modu
...
-        return False
+    if max_runners is not None:
+        if isinstance(max_runners, dict):
+            # Mapping form: one positive int per cluster, with `default` as the baseline.
+            if "default" not in max_runners:
+                log_error(
+                    f"Invalid definition file {def_file}: max_runners mapping must include a "
+                    f"`default` key for the baseline value, got keys {sorted(max_runners)}"
+                )
+                return False
```
malfet: Just curious, why not `throw ValueError("Invalid definition file {def_file}:...")`?

huydhn: Good point, throwing here is easier indeed.  Return false here will skip the invalid definition to move on to the next one, which is not what I want

malfet (approved):

PR #634, opened 2026-05-27
cluster-config: add runner-image-modules subcommand and just recipes

jeanschmidt:
* #644
* #643
* #642
* #641
* #640
* #639
* #638
* #637
* #636
* #635
* __->__ #634

**Impact:** OSDC tooling for cluster config introspection
**Risk:** low

## What
Adds a new `runner-image-modules` subcommand to `cluster-config.py` that
returns the comma-separated list of enabled modules consuming
`runner_image_tag`. Exposes `region` and `runner-image-modules` as `just`
recipes so workflows and operators can call them uniformly.

## Why
Upcoming Renovate-driven runner image auto-update needs a single source of
truth for which modules to redeploy when `runner_image_tag` bumps. CI
workflows also need a uniform way to derive the AWS region for a cluster
without hardcoding values.

## How
- New constant `RUNNER_IMAGE_CONSUMER_MODULES` lists the modules whose
  generated manifests embed `runner_image_tag` (arc-runners and its
  b200/h100 delegates).
- `runner-image-modules` filters the cluster's enabled modules through
  that list and prints comma-joined output.
- `just region` and `just runner-image-modules` thin-wrap the existing
  cluster-config commands so downstream tooling has one entrypoint.

## Changes
- `osdc/scripts/cluster-config.py`: new subcommand + consumer list.
- `osdc/scripts/test_cluster_config.py`: unit tests for the new path.
- `osdc/justfile`: two new recipes.

## Testing
- `cd osdc && just test` to run unit tests.
- `just region arc-staging` and `just runner-image-modules arc-staging`
  for manual sanity.

### DIFF
--- a/osdc/justfile
+++ b/osdc/justfile
@@ -168,6 +168,14 @@ deploy-status cluster name='':
         -o json \
     | python3 "{{SCRIPTS}}/deploy-status.py" "${ARGS[@]}"
 
+# Print the AWS region for a cluster (single source of truth: clusters.yaml)
+region cluster:
+    @export CLUSTERS_YAML="{{CLUSTERS_YAML}}"; uv run {{CFG}} {{cluster}} region
+
+# Print the enabled modules that consume runner_image_tag (comma-separated)
+runner-image-modules cluster:
+    @export CLUSTERS_YAML="{{CLUSTERS_YAML}}"; uv run {{CFG}} {{cluster}} runner-image-modules
+
 # Update kubeconfig for a cluster (aws eks update-kubeconfig)
 kubeconfig cluster:
     @export CLUSTERS_YAML="{{CLUSTERS_YAML}}"; \
--- a/osdc/scripts/cluster-config.py
+++ b/osdc/scripts/cluster-config.py
@@ -18,6 +18,10 @@ Usage:
     # Check if a module is enabled
     cluster-config.py arc-staging has-module arc      -> exits 0 (true) or 1 (false)
 
+    # List the enabled modules that consume runner_image_tag (comma-separated)
+    cluster-config.py arc-cbr-production runner-image-modules
+        -> arc-runners,arc-runners-b200,arc-runners-h100
+
     # List all cluster IDs
     cluster-config.py --list
 """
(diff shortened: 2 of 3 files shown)


osdc/scripts/cluster-config.py
```diff
@@ -30,6 +34,13 @@
 
 CONFIG_PATH = Path(os.environ.get("CLUSTERS_YAML", Path(__file__).resolve().parent.parent / "clusters.yaml"))
 
+# Modules whose generated manifests embed `runner_image_tag` from clusters.yaml.
+# Authoritative source: modules/arc-runners/scripts/python/generate_runners.py is
```
malfet, line 38: If this is an authoritative source, why this file does not try to import them? Otherwise what will keep them in sync?

osdc/justfile
```diff
@@ -168,6 +168,14 @@ deploy-status cluster name='':
         -o json \
     | python3 "{{SCRIPTS}}/deploy-status.py" "${ARGS[@]}"
 
+# Print the AWS region for a cluster (single source of truth: clusters.yaml)
+region cluster:
+    @export CLUSTERS_YAML="{{CLUSTERS_YAML}}"; uv run {{CFG}} {{cluster}} region
```
malfet, line 173: Seems unrelated

osdc/scripts/test_cluster_config.py
```diff
@@ -328,6 +328,56 @@ def test_modules_production(self, capsys):
         assert "monitoring" in lines
         assert "buildkit" in lines
 
+    def test_runner_image_modules_staging(self, capsys):
```
malfet, line 331: I've tried to look at the logs in https://github.com/pytorch/ci-infra/actions/runs/26486182516/job/77994041549?pr=634 and didn't find those tests to be executed. If they are not run in CI, what's the point of adding new tests?

malfet (commented):
Please explain if new tests are actually validated in CI
And I'm confused about `RUNNER_IMAGE_CONSUMER_MODULES` concept. What is it? Why you want to filter out some modules but not others?

osdc/scripts/test_cluster_config.py
jeanschmidt: Hi, we don't print every test that we run on CI, only red ones and coverage per file (the gate is all tests green and at least 97% coverage on all python files).

so for `osdc/scripts/cluster_config.py` the coverage should be on 98%

https://github.com/pytorch/ci-infra/actions/runs/26486182516/job/77994041549?pr=634#step:6:96

Do you believe we should print all tests we run?

osdc/justfile
jeanschmidt: its used on https://github.com/pytorch/ci-infra/pull/635/changes#diff-654d299495d094e327ed7508bdbf92d26be6e78c742bfab0929f53ceb7c7c022R30

osdc/scripts/cluster-config.py
jeanschmidt: I just don't know a reliable way to check if a module will be deploying ARC scalesets, the best I can come up with is have a list of modules and check them on the cluster.

We can, maybe, force modules that create a ARC scaleset to define a configuration, or have a file, but then the problem just move from one place to another...

Any ideas here?

PR #636, opened 2026-05-27
scripts: add validate-runner-bump.py for Renovate diff validation

jeanschmidt:
* #644
* #643
* #642
* #641
* #640
* #639
* #638
* #637
* __->__ #636
* #635
* #634

**Impact:** OSDC scripts — new standalone validator
**Risk:** low

## What
Introduces a shared Python validator for Renovate-style runner-image bump
PRs. Confirms the diff touches exactly one file at exactly one hunk, that
the bumped value is strict semver, and that the new version is strictly
greater than the old.

## Why
Both the pre-merge auto-approve gate and the post-merge auto-deploy
workflow need the same correctness guarantees on a Renovate diff before
they trust it. Centralizing the logic in one Python script keeps the YAML
workflows minimal and gives us proper unit-test coverage.

## How
- Pure-Python, no third-party deps; parses unified-diff text and the
  target file path from CLI args.
- Strict checks: single file, single hunk, single changed line, strict
  semver on both sides, new > old.
- Exit non-zero with a precise reason on any failure so callers can fail
  fast and the surfaced error tells reviewers exactly what is off.
- Ships with a comprehensive unit-test file covering happy path and every
  rejection branch.

## Changes
- `osdc/scripts/validate-runner-bump.py`: new validator.
- `osdc/scripts/test_validate_runner_bump.py`: unit tests.

## Notes
No callers yet; wired up in later commits.

## Testing
- `cd osdc && just test` runs the new test suite.

### DIFF
--- a/osdc/scripts/validate-runner-bump.py
+++ b/osdc/scripts/validate-runner-bump.py
@@ -1,0 +1,135 @@
+#!/usr/bin/env python3
+# /// script
+# requires-python = ">=3.12"
+# ///
+"""Validate a Renovate-style runner_image_tag bump patch.
+
+Defense-in-depth gate shared by osdc-renovate-autoapprove.yml (pre-merge,
+operates on PR file list) and osdc-auto-update-deploy-prod.yml (post-merge,
+operates on the merge commit). Keeping bit-for-bit identical regex,
+line-count, and monotonicity rules in one place avoids drift between the
+two enforcement points.
+
+Inputs (env vars):
+    PATCH_JSON  JSON from `gh api repos/X/pulls/N/files` (top-level array)
+                OR `gh api repos/X/commits/SHA` (object with .files).
+                Shape is auto-detected: array => PR form; object => commit form.
+
+Outputs (stdout, KEY=VALUE lines):
+    decision=approve | close-wrong-file-count | close-wrong-file |
+             close-no-patch | close-multi-line | close-bad-pattern |
+             close-no-change | close-downgrade
+    reason=<human-readable explanation>
+    old_ver=<X.Y.Z>  (only when decision=approve)
+    new_ver=<X.Y.Z>  (only when decision=approve)
+
+Exit code: 0 on every deterministic decision (callers branch on `decision`).
+Non-zero only for unrecoverable errors (missing PATCH_JSON, malformed JSON).
+"""
+
+import json
+import os
+import re
+import sys
+
+EXPECTED_FILE = "osdc/clusters.yaml"
+
+# MUST remain bit-for-bit equivalent to the prior bash regex so the pre-
+# and post-merge gates accept/reject the same set of patches.
+LINE_PATTERN = re.compile(r'^    runner_image_tag:[ \t]*"(?P<ver>\d+\.\d+\.\d+)"[ \t]*(#.*)?$')
+
+
+def _emit(decision: str, reason: str, **extra: str) -> None:
+    print(f"decision={decision}")
+    print(f"reason={reason}")
+    for k, v in extra.items():
+        print(f"{k}={v}")
+
+
+def _semver_tuple(version: str) -> tuple[int, int, int]:
+    parts = version.split(".")
+    return (int(parts[0]), int(parts[1]), int(parts[2]))
+
+
+def main() -> int:
+    raw = os.environ.get("PATCH_JSON")
+    if not raw:
+        print("validate-runner-bump.py: PATCH_JSON env var is required", file=sys.stderr)
+        return 2
+
+    try:
+        payload = json.loads(raw)
+    except json.JSONDecodeError as e:
+        print(f"validate-runner-bump.py: PATCH_JSON is not valid JSON: {e}", file=sys.stderr)
+        return 2
+
+    # Auto-detect shape: top-level array = PR-files form; object = commit form.
+    if isinstance(payload, list):
+        files = payload
+    elif isinstance(payload, dict):
+        files = payload.get("files") or []
+    else:
+        print(
+            f"validate-runner-bump.py: PATCH_JSON must be array or object, got {type(payload).__name__}",
+            file=sys.stderr,
+        )
+        return 2
+
+    if len(files) != 1:
+        _emit("close-wrong-file-count", f"expected exactly 1 file changed; got {len(files)}")
+        return 0
+
+    file_path = files[0].get("filename", "")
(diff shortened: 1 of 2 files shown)


osdc/scripts/validate-runner-bump.py
```diff
@@ -0,0 +1,135 @@
...
+    else:
+        print(
+            f"validate-runner-bump.py: PATCH_JSON must be array or object, got {type(payload).__name__}",
+            file=sys.stderr,
+        )
+        return 2
+
+    if len(files) != 1:
+        _emit("close-wrong-file-count", f"expected exactly 1 file changed; got {len(files)}")
+        return 0
```
malfet, line 75: Why wrong file count is not an error?

osdc/scripts/validate-runner-bump.py
```diff
@@ -0,0 +1,135 @@
...
+"""
+
+import json
+import os
+import re
+import sys
+
+EXPECTED_FILE = "osdc/clusters.yaml"
+
+# MUST remain bit-for-bit equivalent to the prior bash regex so the pre-
```
malfet: > MUST remain bit-for-bit equivalent to the prior bash regex

There are no prior, not sure what you are referring to...

osdc/scripts/validate-runner-bump.py
```diff
@@ -0,0 +1,135 @@
...
+# and post-merge gates accept/reject the same set of patches.
+LINE_PATTERN = re.compile(r'^    runner_image_tag:[ \t]*"(?P<ver>\d+\.\d+\.\d+)"[ \t]*(#.*)?$')
+
+
+def _emit(decision: str, reason: str, **extra: str) -> None:
```
malfet: Extra is not an str, is it? Also traditionally this is called kwargs
```suggestion
def _emit(decision: str, reason: str, **extra: dict) -> None:
```

osdc/scripts/validate-runner-bump.py
```diff
@@ -0,0 +1,135 @@
...
+    print(f"decision={decision}")
+    print(f"reason={reason}")
+    for k, v in extra.items():
+        print(f"{k}={v}")
```
malfet: Nit
```suggestion
    print(*(f"{k}={v}" for k, v in {"decision": decision, "reason": reason, **extra}.items()), sep="\n")
```

osdc/scripts/validate-runner-bump.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+def _semver_tuple(version: str) -> tuple[int, int, int]:
+    parts = version.split(".")
+    return (int(parts[0]), int(parts[1]), int(parts[2]))
```
malfet: Nit
```suggestion
    return tuple(int(x) for x in version.split("."))
```

malfet (changes requested):
Why it needs to be so complex? All this 400 line change does is checks whether some property in `osdc/clusters.yaml` contains positive version bump right? Why not do something like
`current, old = json.read("osdc/clusters.yaml"),  json.read_string(subprocess.check_output("git show HEAD~1 osdc/clusters.yml"))` and just compare the appropriate key? And you can make a very simple `gh api call` to figure out whether diff is 1 line, can't you?

malfet (commented):
Asked claude to write succint diff for validating that something is positive semver and this is what it comeup with
```python
#!/usr/bin/env python3
"""Exit 0 iff `git diff` is a single property line with a positive semver bump."""
import re
import subprocess
import sys

SEMVER = re.compile(r"(\d+)\.(\d+)\.(\d+)")
PROP = re.compile(r"^[-+]\s*([\w.-]+)\s*[:=]\s*['\"]?v?(\d+\.\d+\.\d+)['\"]?\s*,?\s*$")


def main() -> int:
    diff = subprocess.check_output(["git", "diff", "--unified=0", "--no-color"], text=True)
    hunks = [l for l in diff.splitlines() if l.startswith(("+", "-")) and not l.startswith(("+++", "---"))]
    if len(hunks) != 2 or hunks[0][0] == hunks[1][0]:
        return 1
    old, new = sorted(hunks)  # '-' < '+'
    om, nm = PROP.match(old), PROP.match(new)
    if not (om and nm) or om.group(1) != nm.group(1):
        return 1
    ov, nv = tuple(map(int, om.group(2).split("."))), tuple(map(int, nm.group(2).split(".")))
    return 0 if nv > ov else 1


if __name__ == "__main__":
    sys.exit(main())
```

jeanschmidt:
Thanks — taking the nits at lines 37/42/46/51 in the next revision (kwargs rename, dict-merge `_emit`, comprehension `_semver_tuple`, stale "prior bash regex" comment deleted).

On replacing the script with the ~25-line version: pushing back, three reasons.

**1. Two callers, two diff shapes.** `osdc-renovate-autoapprove.yml` (pre-merge) has no checkout and consumes `gh api repos/X/pulls/N/files` (array). `osdc-auto-update-deploy-prod.yml` (post-merge) consumes `gh api repos/X/commits/SHA` (object). `git diff --unified=0` needs a fetch+checkout in both — more YAML, not less.

**2. The regex is intentionally over-specified.** `^[-+]\s*([\w.-]+)\s*[:=]\s*['\"]?v?(\d+\.\d+\.\d+)` matches *any* `key: semver` on *any* file. This gate is defense-in-depth against a compromised `RENOVATE_TOKEN`: it must refuse anything that isn't literally `^    runner_image_tag:[ \t]*"X.Y.Z"...$` inside `osdc/clusters.yaml`. A loose regex turns the gate into a rubber stamp.

**3. Structured `decision=` drives the close-comment UX.** On reject the workflow posts `Auto-closed: close-downgrade — 1.3.0 -> 1.2.99` and Renovate retries on the next eligible version. A boolean exit collapses seven distinct reasons into "validation failed" — harder to triage stale closures.

> Why wrong file count is not an error?

Non-zero exit aborts the job before the close-comment step runs. Exit codes are reserved for unrecoverable input errors (missing/malformed `PATCH_JSON`); deterministic outcomes go through `decision=` so one YAML handler covers all rejection branches.

The script is ~130 LOC, another 300ish lines are tests to meet the 97% coverage and test things like pin every branch — downgrade, `1.10.0 > 1.9.0` lex-vs-numeric, unquoted version, wrong indent, wrong key, inline-comment tolerance, hunk-header handling, both API shapes. For the security boundary on prod runner rollout, I'd rather over-test.

PR #647, opened 2026-05-27
[Grafana][PyTorch DevX] Introduce historical % Commits Red on Main (broken trunk)

jathu:
Like HUD, let's introduce a time series of "% Commits Red on Main (broken trunk)". Each data point has the daily redness. HUD by default uses a 7-day avg, so the blue line should represent that.

# Test Plan

Pushed to my test folder: https://pytorchci.grafana.net/dashboards/f/flrhzs. See **"% Commits Red on Main (broken trunk)"**

```
$ mise run generate --folder flrhzs
```

### DIFF
--- a/grafana/pytorch_devx.json
+++ b/grafana/pytorch_devx.json
@@ -20,7 +20,7 @@
       }
     }
   ],
-  "cursorSync": "Off",
+  "cursorSync": "Crosshair",
   "editable": true,
   "elements": {
     "panel-4": {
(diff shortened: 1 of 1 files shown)


huydhn:
@claude Review this PR, focus on the query to make sure that it's consistent with the same query on hud.pytorch.org/metrics from https://github.com/pytorch/test-infra/tree/main/torchci/clickhouse_queries

huydhn (approved):

PR #655, opened 2026-05-29
[Grafana][OSDC] Use palette-classic-by-name colors

jeanschmidt:
* #659
* #657
* #656
* __->__ #655

----

- Switch all 7 panels from palette-classic to
  palette-classic-by-name color mode

Notes:
palette-modern-by-name does not work as expected.
fieldConfig.overrides would require deployment-time
tooling to query clusters and assign colors dynamically,
which adds unwanted complexity. palette-classic-by-name
is a best-effort approach but consistent per-series
colors across time ranges remain an open problem.

### DIFF
--- a/grafana/osdc.json
+++ b/grafana/osdc.json
@@ -69,7 +69,7 @@
             "fieldConfig": {
               "defaults": {
                 "color": {
-                  "mode": "palette-classic",
+                  "mode": "palette-classic-by-name",
                   "seriesBy": "max"
                 },
                 "custom": {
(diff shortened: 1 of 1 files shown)


huydhn (approved):

jathu (changes requested):
Kinda hard to differentiate the colors right now (between dark green and black)

Image

jeanschmidt:
> fieldConfig.overrides would require deployment-time tooling to query clusters and assign colors dynamically, which adds unwanted complexity. palette-classic-by-name is a best-effort approach but consistent per-series colors across time ranges remain an open problem.

There are 3 options AFAIK
* keep as is: the colors are different between graphs
* [this solution]: let grafana pick the colors based on hash (but you can't change them, it is what it is)
* write a deployment-time script to hardcode colors (BAD IMO, removes the dynamic nature)

jeanschmidt:
@claude review these changes

jathu (approved):

PR #656, opened 2026-05-29
[Grafana][OSDC] Add queue time heatmaps

jeanschmidt:
* #659
* #657
* __->__ #656
* #655

----

- Add panel-10: queue time heatmap excluding capacity-constrained GPU instances (h100, b200, gb200, a100)
- Add panel-11: queue time heatmap for capacity-constrained GPU instances only
- Both panels use ClickHouse workflow_job data with log-scale Y-axis in minutes
- Layout places the two heatmaps side-by-side (12 cols each) in a new row at y=40

Notes:
The two panels split queue time visibility by GPU scarcity.
Non-constrained runners (panel-10, green) show general queue health,
while constrained runners (panel-11, blue) isolate wait times caused
by limited GPU capacity. Both use exponential opacity color scaling
with 30 log-10 Y-buckets for resolution across sub-minute to
multi-hour queue times.

### DIFF
--- a/grafana/AGENTS.md
+++ b/grafana/AGENTS.md
@@ -28,6 +28,14 @@ mise run push --folder "..."
   * NEVER make edits to any folders other than the folder UID provided with `--folder`
   * NEVER use curl directly to interact with the API
 * Panel titles must end with the list of dashboard variables referenced in the panel's query, formatted as `[var1, var2]` without the `$` prefix (e.g. `Running Jobs [cluster, scale_set]`). The `$` is omitted so Grafana doesn't interpolate the variable value into the title. Include every `$var`/`${var}` the query uses; omit the suffix entirely if the query uses no variables.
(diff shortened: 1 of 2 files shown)


huydhn (approved):
LGTM!

jeanschmidt:
@claude review this changes

jathu:
@claude look at each new panel individually using a subagent. 

1) Based on the panel's title and description, does the query comply with the intent? 
2) I also noticed that the query is hardcoding runner_group_name, is it possible to make this dynamic?
3) I noticed the user is repeating the list of excluded labels in both panels — can we share them?
4) What is the value in adding a filter for `'2020-01-01`, can we remove this for simplicity without significantly affecting performance?
5) Is this check necessary `AND started_at > created_at`? Would this always be true?

jathu:
@jeanschmidt let's address theses

1) Get claude to fix the titles for the variables being used
2) Perhaps we should keep track of the runner group to cluster in AGENTS.md and tell to keep them in sync within the dashboards
4) Let's drop the check

PR #657, opened 2026-05-29
[Grafana][OSDC] Add runner_name_prefix variable

jeanschmidt:
* #659
* __->__ #657
* #656
* #655

----

- Add new runner_name_prefix dashboard variable that extracts
  the prefix from runner_name via ClickHouse query
- Filter both heatmap queries by runner_name_prefix to allow
  drilling down to specific runner configurations
- Cascade runner_name_prefix into scale_set variable so the
  scale set dropdown only shows matching entries

### DIFF
--- a/grafana/osdc.json
+++ b/grafana/osdc.json
@@ -1215,7 +1215,7 @@
                       },
                       "pluginVersion": "4.17.0",
                       "queryType": "timeseries",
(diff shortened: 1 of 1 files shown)


huydhn:
@claude Check those new SQL queries for correctness and performance

jathu (changes requested):
What is the value in having scale set and runner name? Can we use one over the other? (I am not familiar with the differences)

huydhn (approved):

jeanschmidt:
> What is the value in having scale set and runner name? Can we use one over the other? (I am not familiar with the differences)

Sure, I believe we do, but adds complexity into the graph we need to decide.
* Runner name: the label jobs request, those are your ml-l-... like `mt-l-arm64g3-16-62`.
* Scale Set name: each cluster deploys, per runner label, a scale set name. It takes the format `mt-l-arm64g3-16-62-<hash>`. It can change as we do deployments.

so my goal here is: I want to see how `mt-l-arm64g3-16-62` are behaving, instead of searching all `mt-l-arm64g3-16-62-<hash>` in the scale set name I can see in a single place to filter them all (if you do that, the scale set also updates to match only the ones you care)
BUT, it I want to drill down to a specific deployment/version of the scale set, or a region, or something, I can do so by filtering on the scale set name

jathu (approved):

PR #688, opened 2026-06-02
remove nodepools and arc-runners from uw1 region

jeanschmidt:
* __->__ #688
* #687

uw1 is about 20% more expensive. We already have full duplication from ue1 and ue2. So we don't need to run those expensive GPU jobs on this region.


huydhn (approved):
LGTM! Probably the easiest way to not run anything on uw1 instead of manually exclude different runner type

PR #694, opened 2026-06-03
Add HUD_API_URL to secrets creation

zxiiro:
Make HUD_API_URL be created in our crcr-secrets for use by our lambda functions.

### DIFF
--- a/.github/workflows/crcr-deploy-prod.yml
+++ b/.github/workflows/crcr-deploy-prod.yml
@@ -46,5 +46,5 @@ jobs:
         TF_VAR_github_app_id: ${{ secrets.CRCR_GITHUB_APP_ID }}
         TF_VAR_github_app_secret: ${{ secrets.CRCR_GITHUB_APP_SECRET }}
         TF_VAR_github_app_privatekey: ${{ secrets.CRCR_GITHUB_APP_PRIVATEKEY }}
-        TF_VAR_hud_api_url: ${{ secrets.CRCR_HUD_API_URL }}
+        TF_VAR_hud_api_url: ${{ vars.CRCR_HUD_API_URL }}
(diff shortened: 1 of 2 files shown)


crcr/aws/secrets.tf
```diff
@@ -9,6 +9,7 @@ resource "aws_secretsmanager_secret_version" "main" {
     GITHUB_APP_SECRET      = var.github_app_secret
     GITHUB_APP_PRIVATE_KEY = var.github_app_privatekey
     REDIS_LOGIN            = random_password.redis_password.result
+    HUD_API_URL            = var.hud_api_url
```
fffrog: Hey @zxiiro，thank you for updates.

I am so sorry for confusion, from my perspective, it is not necessary to move `HUD_API_URL` to AWG Security Manager, I think placing it in environment of callback lambda is enough :D

can-gaa-hou: https://github.com/pytorch/ci-infra/blob/6ece301aeefedb6ae3f6f0d954923e405b5417c9/crcr/aws/callback.tf#L27

We already have this one in the callback lambda environment, so we don't need to have one more in AWS Secret Manager :)

Update: the only place you need to configure this value is in the GitHub secrets of this repo, since I add the following code in the workflow

https://github.com/pytorch/ci-infra/blob/6ece301aeefedb6ae3f6f0d954923e405b5417c9/.github/workflows/crcr-on-pr.yml#L68

zxiiro: Thanks for the clarification. I've removed it as a secret and switched it over to a environment variable.

I believe secrets.CRCR_HUD_API_URL is the the right call to handle variables though. I think it should be `${{ vars.CRCR_HUD_API_URL }}`. I'll update this PR to fix that.

fffrog (approved):
SGTM.

Thank you very much for your help.

PR #699, opened 2026-06-04
Add .env to .gitignore

zxiiro:
This is a local environment file that should not be committed.


zxiiro:
I'm using this to override the helm install that's not working on macos. Since I have it installed as a system tool I'm using `MISE_DISABLE_TOOLS=helm`

jordanconway (approved):

PR #700, opened 2026-06-04
OSDC: Add lf-prod-aws-ue1 and lf-prod-aws-ue2 clusters

zxiiro:
This adds 2 new clusters to OSDC: lf-prod-aws-ue1 and lf-prod-aws-ue2

### DIFF
--- a/.github/workflows/_osdc-deploy.yml
+++ b/.github/workflows/_osdc-deploy.yml
@@ -12,6 +12,11 @@ on:
         required: false
         type: string
         default: ""
+      use_lf_credentials:
+        description: "Use LF_AWS_DEPLOY_ROLE_ARN secret instead of Meta AWS credentials"
+        required: false
+        type: boolean
+        default: false
       taint_nodes:
         description: "Taint ARC runner nodes with NoSchedule before deploy for graceful refresh"
         required: false
@@ -84,7 +89,7 @@ jobs:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@b47578312673ae6fa5b5096b330d9fbac3d116df # v4.2.1
         with:
-          role-to-assume: arn:aws:iam::${{ secrets.META_AWS_ACC_ID }}:role/${{ secrets.META_AWS_DEPLOY_ROLE }}
+          role-to-assume: ${{ inputs.use_lf_credentials && secrets.LF_AWS_DEPLOY_ROLE_ARN || format('arn:aws:iam::{0}:role/{1}', secrets.META_AWS_ACC_ID, secrets.META_AWS_DEPLOY_ROLE) }}
           aws-region: us-west-1
           role-duration-seconds: 7200
 
@@ -134,7 +139,7 @@ jobs:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@b47578312673ae6fa5b5096b330d9fbac3d116df # v4.2.1
         with:
-          role-to-assume: arn:aws:iam::${{ secrets.META_AWS_ACC_ID }}:role/${{ secrets.META_AWS_DEPLOY_ROLE }}
+          role-to-assume: ${{ inputs.use_lf_credentials && secrets.LF_AWS_DEPLOY_ROLE_ARN || format('arn:aws:iam::{0}:role/{1}', secrets.META_AWS_ACC_ID, secrets.META_AWS_DEPLOY_ROLE) }}
           aws-region: us-west-1
           role-duration-seconds: 7200
 
@@ -175,7 +180,7 @@ jobs:
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@b47578312673ae6fa5b5096b330d9fbac3d116df # v4.2.1
         with:
-          role-to-assume: arn:aws:iam::${{ secrets.META_AWS_ACC_ID }}:role/${{ secrets.META_AWS_DEPLOY_ROLE }}
+          role-to-assume: ${{ inputs.use_lf_credentials && secrets.LF_AWS_DEPLOY_ROLE_ARN || format('arn:aws:iam::{0}:role/{1}', secrets.META_AWS_ACC_ID, secrets.META_AWS_DEPLOY_ROLE) }}
           aws-region: us-west-1
           role-duration-seconds: 7200
 
--- a/.github/workflows/osdc-deploy-prod.yml
+++ b/.github/workflows/osdc-deploy-prod.yml
@@ -1,14 +1,16 @@
 name: "OSDC: Deploy production"
 
-# Manual rollout for prod. Protected by the `osdc-production` GitHub
-# environment (configure required reviewers in repo Settings → Environments) —
-# the prod IAM role additionally trusts only tokens with
-# `environment:osdc-production` in the sub claim, so IAM enforces the gate too.
+# Manual rollout for prod. Protected by per-account GitHub environments:
+#   - Meta clusters: `osdc-production` environment
+#   - LF clusters:   `osdc-lf-production` environment
+# Each environment's IAM role trusts only tokens carrying that environment
+# name in the sub claim, so IAM enforces the gate independently.
 #
-# Default behavior (target = all): sequential rollout.
-#   1. Deploy arc-cbr-production-uw1 + smoke.
-#   2. Deploy arc-cbr-production (ue2) + smoke.
-#   3. Deploy meta-prod-aws-ue1 + smoke.
+# Default behavior (target = all): sequential rollout of all clusters.
+#   Meta: arc-cbr-production-uw1 → arc-cbr-production → meta-prod-aws-ue1
+#   LF:   lf-prod-aws-ue1 → lf-prod-aws-ue2 (runs in parallel with Meta)
+#
+# Use `all-meta` or `all-lf` to deploy only one account's clusters.
 #
 # Picking a specific cluster from the dropdown bypasses the sequencing and
 # deploys only that one — useful for hotfixes or to recover from a partial
@@ -27,9 +29,13 @@ on:
         default: all
         options:
           - all
-          - meta-prod-aws-ue1
+          - all-meta
+          - all-lf
           - arc-cbr-production-uw1
           - arc-cbr-production
+          - meta-prod-aws-ue1
+          - lf-prod-aws-ue1
+          - lf-prod-aws-ue2
       taint_nodes:
         description: "Taint ARC runner nodes before deploy (graceful refresh)"
         required: false
@@ -55,8 +61,10 @@ permissions:
   contents: read
 
 jobs:
+  # --- Meta clusters ---
+
   deploy_uw1:
-    if: ${{ inputs.target == 'all' || inputs.target == 'arc-cbr-production-uw1' }}
+    if: ${{ inputs.target == 'all' || inputs.target == 'all-meta' || inputs.target == 'arc-cbr-production-uw1' }}
     uses: ./.github/workflows/_osdc-deploy.yml
     with:
       cluster: arc-cbr-production-uw1
@@ -72,7 +80,7 @@ jobs:
     if: |
       !cancelled() && (
         inputs.target == 'arc-cbr-production' ||
-        (inputs.target == 'all' && needs.deploy_uw1.result == 'success')
+        ((inputs.target == 'all' || inputs.target == 'all-meta') && needs.deploy_uw1.result == 'success')
       )
     uses: ./.github/workflows/_osdc-deploy.yml
     with:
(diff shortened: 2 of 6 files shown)


zxiiro:
Just saving my progress. This is not deployed yet and I have not run any validation steps yet. It is a work in progress.

osdc/clusters.yaml
```diff
@@ -318,3 +318,39 @@ clusters:
       - harbor-cache-recovery
       - monitoring
       - logging
+
+  lf-prod-aws-ue1:
+    region: us-east-1
+    cluster_name: lf-prod-aws-ue1
+    pause_runners: true
```
zxiiro: Reminder to remove this before this is considered for merging.

jeanschmidt:
we might want full duplication on two regions, don't we?

zxiiro:
> we might want full duplication on two regions, don't we?

Yeah we could definitely do that for redundancy. Maybe we can do it as a followup once this cluster is actively working?

osdc/clusters.yaml
```diff
@@ -319,3 +319,39 @@ clusters:
...
+    git_cache:
+      central_replicas: 2
+      central_cpu_request: "2"
+      central_cpu_limit: "4"
+      central_memory_request: "4Gi"
+      central_memory_limit: "8Gi"
+    node_compactor:
+      min_node_age_seconds: 900
+    arc-runners:
+      github_config_url: "https://github.com/pytorch/pytorch"
```
huydhn: IIRC, this line might be problematic. I understand that we only want LF runners on `pytorch/pytorch` but this implies that LF fleet can't use `runner_group` because GitHub only supports `runner_group` option for org-wide runners.  Repo-specific runner will go to the default runner group.  This implies that we can't have multi-region on LF fleet because each region must be in its own runner group to avoid runner name collision

huydhn: Maybe we just need to ack the fact and field only one cluster on LF

zxiiro: I'm fine with 1 cluster config. I think at some point the pytorch github org is supposed to move over to LF management at which point we could probably move the LF cluster to org level.

huydhn: It looks like we could have this register org-wide with `github_config_url: "https://github.com/pytorch` and limit the runners to only selected repos, `pytorch/pytorch`.  We should try it out, I haven't done it before

jeanschmidt: We should absolutely do this, as it is a known issue that a single region causes InsufficientCapacity errors, and it would be bad to have queuing on LF side if we can avoid this

jeanschmidt: I created a runner group `[lf-prod-aws-ue1](https://github.com/organizations/pytorch/settings/actions/runner-groups/89)`

seems that it is a configuration we already use for other runners/runner groups, so it should be safe

zxiiro: Okay I've updated my config to use org level runners now by pointing to github.com/pytorch. I need to wait for a quota increase for EIPs since Jean's change on Friday adds additional secondary EIPs. I will try a deploy once I have the quota increased.

zxiiro: Alright based on @jeanschmidt 's comment this is now updated to deploy org level runners.

huydhn: > Alright based on @jeanschmidt 's comment this is now updated to deploy org level runners.

Oh nice, you would need to add a runner_group too, pointing to the one that Jean has already setup `lf-prod-aws-ue1`

zxiiro:
@jeanschmidt @huydhn The latest code is deployed and smoke tests are passing. This was deployed manually. I think this is ready to go. I will work on wiring up the CI/CD workflow in another PR. 

The monitoring and logging modules have also been added in this iteration since @jeanschmidt gave me the grafana keys. I don't yet have access to that system so cannot check but I think the cluster should be sending logs over to there now.

osdc/clusters.yaml
```diff
@@ -319,3 +319,40 @@ clusters:
...
+      github_secret_name: lf-prod-aws-ue1
+      runner_name_prefix: "lf-"
+    pypi_cache:
+      replicas: 2
+    modules:
+      - karpenter
+      - arc
+      - nodepools
+      - arc-runners
+      - pypi-cache
```
jeanschmidt, line 408: you need to add the `buildkit` module, otherwise image builds will fail

zxiiro: Updated.

osdc/clusters.yaml
```diff
@@ -319,3 +319,41 @@ clusters:
...
+      central_cpu_request: "2"
+      central_cpu_limit: "4"
+      central_memory_request: "4Gi"
+      central_memory_limit: "8Gi"
+    node_compactor:
+      min_node_age_seconds: 900
+    arc-runners:
+      github_config_url: "https://github.com/pytorch"
+      github_secret_name: lf-prod-aws-ue1
+      runner_name_prefix: "lf-"
```
huydhn, line 338:
```suggestion
      runner_name_prefix: "lf-"
      runner_group: lf-prod-aws-ue1
```

zxiiro:
@huydhn I deployed your suggested change into the cluster. 

Just ran integration tests but its failing but I don't think it's due to anything with the cluster. Seems like the integration test is failing due to permissions issue with the GitHub App I'm using. The GitHub App I'm using `pytorch-fdn-gha-arc` does not have permissions to see pytorch-canary and run its workflows. If I recall it only has permissions to pytorch/pytorch and pytorch/test-infra.

zxiiro:
> @huydhn I deployed your suggested change into the cluster.
> 
> Just ran integration tests but its failing but I don't think it's due to anything with the cluster. Seems like the integration test is failing due to permissions issue with the GitHub App I'm using. The GitHub App I'm using `pytorch-fdn-gha-arc` does not have permissions to see pytorch-canary and run its workflows. If I recall it only has permissions to pytorch/pytorch and pytorch/test-infra.

Fixed. Didn't realize it needed a local PAT for integration tests so it was failing due to me not having a local PAT with workflow access.

Was able to get the integration-tests to pass.

https://github.com/pytorch/pytorch-canary/actions/runs/27170990930

osdc/clusters.yaml
```diff
@@ -319,3 +319,42 @@ clusters:
...
+    cluster_name: lf-prod-aws-ue1
+    state_bucket: lf-osdc-tfstate-prod-ue1
+    access_config:
+      authentication_mode: API_AND_CONFIG_MAP
+      cluster_admin_role_names:
+        - lf_osdc_admin
+    base:
+      base_node_count: 4
+      vpc_cidr: "10.12.0.0/16"
+    git_cache:
```
huydhn: Just FYI, Jean is going to remote this soon https://github.com/pytorch/ci-infra/pull/703, just note it here to avoid a land race later

osdc/clusters.yaml
```diff
@@ -319,3 +319,42 @@ clusters:
...
+      central_memory_limit: "8Gi"
+    node_compactor:
+      min_node_age_seconds: 900
+    arc-runners:
+      github_config_url: "https://github.com/pytorch"
+      github_secret_name: lf-prod-aws-ue1
+      runner_name_prefix: "lf-"
+      runner_group: lf-prod-aws-ue1
+    pypi_cache:
+      replicas: 2
```
huydhn: This looks low, other prod clusters use 10 here. We had an incident before where a burst of traffics kills 2 replicas

huydhn (approved):
LGTM! Plz resolve the remaining comments before merging

osdc/clusters.yaml
zxiiro: I went ahead and removed the module from my latest iteration.

zxiiro:
> LGTM! Plz resolve the remaining comments before merging

@jeanschmidt wanted me to also tackle adding a 2nd cluster in this PR so I updated this PR to handle that. The 2nd cluster is not yet fully deployed pending AWS Quata increase request so hold off on merging this one until I can confirm deployment.

osdc/clusters.yaml
zxiiro: I don't expect LF clusters to handle as much traffic as Meta's one since we have limited budget. So I set it to 6 replicas for now. We can tune it more later once we have more data on real load on the system.

huydhn: @jeanschmidt Did we do any analysis on the number of replicas there or did we just pick a reasonably high number and yolo go with it?

zxiiro:
I'm also reducing the buildkit replicas to 8 for the LF clusters.

osdc/clusters.yaml
huydhn: Per the chat with @jeanschmidt earlier today, let's look out for multiple buildkit cluster to see if they are working as expected
* Meta fleet only fields 1 Buildkit cluster to save cost
* Release runners current use fixed runner group https://github.com/pytorch/ci-infra/blob/main/osdc/modules/arc-runners/defs/rel-l-x86iavx512-8-64.yaml#L11

jeanschmidt:
> I'm also reducing the buildkit replicas to 8 for the LF clusters.

Note that @huydhn is working on a change to make buildkit dynamic scalable AND available on all regions. So maybe you would be good to just keep the standard config and leverage those changes when we deploy them

osdc/clusters.yaml
jeanschmidt: no, I started with 2, and increased every time we had issues :) 

So, I'll recommend stick with those numbers.

jeanschmidt:
I believe we're missing CI changes for this, @zxiiro are you planning to add it as a separate PR?

zxiiro:
> I believe we're missing CI changes for this, @zxiiro are you planning to add it as a separate PR?

Yes I plan to add CI in a followup PR once this is live.

Also I'm still waiting on AWS to increase our Quota in us-east-2, it's been taking awhile to get a response so I escalated it with our account team.

.github/workflows/osdc-deploy-prod.yml
```diff
@@ -100,3 +108,36 @@ jobs:
...
     secrets: inherit
+
+  # --- LF clusters ---
+
+  deploy_lf_ue1:
+    if: ${{ inputs.target == 'all' || inputs.target == 'all-lf' || inputs.target == 'lf-prod-aws-ue1' }}
+    uses: ./.github/workflows/_osdc-deploy.yml
+    with:
+      cluster: lf-prod-aws-ue1
+      environment: osdc-production
```
jeanschmidt, line 117: on .github/workflows/osdc-deploy-prod.yml you are mentioning that you use `osdc-lf-production` worth checking what is the right one

.github/workflows/_osdc-deploy.yml
```diff
@@ -84,7 +89,7 @@ jobs:
       - name: Configure AWS credentials
```
jeanschmidt, line 89: should we add a guard step in case `use_lf_credentials` is set but the secret is empty (and vice-versa)?

something like:
```
  - name: Verify LF secret present
    if: ${{ inputs.use_lf_credentials }}
    run: |
      test -n "${{ secrets.LF_AWS_DEPLOY_ROLE_ARN }}" || { echo "LF role ARN not configured"; exit 1; }
  - name: Verify Meta secret present
    if: ${{ inputs.use_lf_credentials }}
    run: |
      test -n "${{ secrets.META_AWS_DEPLOY_ROLE }}" || { echo "Meta role not configured"; exit 1; }
      test -n "${{ secrets.META_AWS_ACC_ID }}" || { echo "Meta role account id not configured"; exit 1; }
```

.github/workflows/osdc-deploy-prod.yml
zxiiro: I changed my mind on that and decided it would be cleaner to use the existing osdc-production and osdc-staging environments since we already had them configured. I'll check and remove any place I forgot to remove the `osdc-lf-production` in docs.

osdc/clusters.yaml
```diff
@@ -317,3 +317,75 @@ clusters:
...
+      base_node_count: 4
+      vpc_cidr: "10.12.0.0/16"
+    node_compactor:
+      min_node_age_seconds: 900
+    arc-runners:
+      github_config_url: "https://github.com/pytorch"
+      github_secret_name: lf-prod-aws-ue1
+      runner_name_prefix: "lf-"
+      runner_group: lf-prod-aws-ue1
+    buildkit:
```
jeanschmidt, line 340: worth looking: https://github.com/pytorch/ci-infra/commit/c4bf41db010cd18a054b8e8f7a330401ca637a79#diff-82b61cda543427ce26ac5658181b0f3a5173fef119b5e44974333d81e4486983R275-R278

.github/workflows/_osdc-deploy.yml
jeanschmidt: if we use the same environment, probably not

.github/workflows/osdc-deploy-prod.yml
zxiiro: I've updated the code to remove mention of the osdc-lf-production environment since I removed it.

osdc/clusters.yaml
zxiiro: I added the autoscaling config. I set the buildkit numbers lower though since 360 systems seems like a lot to allow. Do we really need that much?

jeanschmidt (approved):

osdc/clusters.yaml
jeanschmidt: yes we do :( 

This is the peak of concurrent running build jobs from what we queried.

huydhn: The max number currently set to 2x the peak number of concurrent Docker build jobs on CI. It's just an upper ceiling.
* Given that we are deploying 2 clusters here, it might be reasonable to cut this by half assuming a balance load between the 2 regions.
* Then we could lower if even further by another 20% if we follow the lf dial on https://github.com/pytorch/test-infra/issues/5132, but this seems hard to maintain IMO as the value from the issue can change.

On the other end, the minimum is 1 node per arch as the HA proxy needs to have something to connect to.  We are using `24xlarge` for x86 Docker build so it translates to 2 x86 replicas. Same goes for the `16xlarge` aarch64 = 4x aarch64 replicas minimum.  You can set a lower minimum than that, but it doesn't lower the bill further.

zxiiro: Okay let's just keep it at 4 min and 64 max for now. I will tune it further once we see it in action.

zxiiro:
Since everyone approved. I'm going to merge this now. What are the next steps to put actual load on the system?

jeanschmidt:
after deployed, I'll merge my PT PR and traffic should start flowing, I can do this next morning

jeanschmidt:
https://github.com/pytorch/pytorch/pull/186663

PR #703, opened 2026-06-05
Remove git-cache module and all dependent wiring

jeanschmidt:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #706
* #707
* #705
* #704
* __->__ #703

**Impact:** All OSDC clusters — runner pods, BuildKit pods, NodePools, monitoring, deploy pipeline, integration tests
**Risk:** high

## What
Removes the two-tier git clone cache (central StatefulSet + rsync DaemonSet) and every reference to it across the platform: startup taints, tolerations, volume mounts, env vars, monitoring, cluster config, docs, and tests.

## Why
The git-cache module is being decommissioned. Runner pods no longer need a local git object cache for `actions/checkout` acceleration.

Note on blast radius:
- Lots of tests are being removed;
- Lots of documentations are being removed as well;

Signed-off-by: Jean Schmidt <contato@jschmidt.me>

PR #704, opened 2026-06-05
Add module-aware startup taints framework for Karpenter NodePools

jeanschmidt:
* #706
* #707
* #705
* __->__ #704
* #703

**Impact:** All OSDC clusters using the nodepools module
**Risk:** low

## What
Introduces a registry-driven framework for emitting Karpenter `startupTaints` on NodePools, gated by which modules are enabled on each cluster. This prevents pods from scheduling on nodes before module-specific DaemonSets have finished initializing.

## Why
After removing the git-cache module, its taint no longer prevented early pod scheduling on nodes that weren't fully initialized. This adds a general-purpose mechanism so any module can declare startup taints that block scheduling until its init DaemonSet removes them — without risking stuck taints on clusters where the module isn't deployed.

## Notes
The `MODULE_STARTUP_TAINTS` registry is currently empty — this commit adds the framework only. Subsequent changes will populate it with specific module taints (e.g. for the init DaemonSet that replaces git-cache's scheduling gate).


osdc/modules/nodepools/scripts/python/generate_nodepools.py
```diff
@@ -245,6 +262,26 @@ def generate_nodepool_yaml(nodepool_def, module_name, defs_dir=None):
...
+    # Emit a startupTaint only when:
+    #   1. its owning module is in NODEPOOLS_ENABLED_MODULES for this cluster, AND
+    #   2. its optional applies_when(nodepool_def) predicate returns True.
+    # This prevents stuck taints on clusters/nodepools where no DaemonSet exists
+    # to remove the taint at end-of-init.
+    enabled_modules = set(os.environ.get("NODEPOOLS_ENABLED_MODULES", "").split())
+    startup_taint_lines = []
+    for module, taints in MODULE_STARTUP_TAINTS.items():
+        if module not in enabled_modules:
+            continue
```
huydhn, line 294: ~~I was thinking if using `continue` here is appropriate. But I guess we have modules like `nodepool-b200` which are only available in some regions, so this makes sense then~~

Wait a sec, I read this wrong.  Looking at the example from https://github.com/pytorch/ci-infra/pull/706, then we should be able to throw an error here when a module is listed in `MODULE_STARTUP_TAINTS`, but is not enabled, right? It looks like a mistake when a cluster config does that

huydhn (approved):

PR #705, opened 2026-06-05
Add shared taint-remover library and wire DaemonSet startup taint removal

jeanschmidt:
* #706
* #707
* __->__ #705
* #704
* #703

**Impact:** All Karpenter-managed runner nodes — scheduling behavior changes
**Risk:** medium

## What
Introduces a shared Python library (`taint_remover.py`) that DaemonSets call at end-of-init to remove their per-DS startup taint, preventing workload pods from scheduling on nodes where critical init (cache-enforcer, registry-mirror, perf-tuning) hasn't completed yet.

## Why
After removing the git-cache module (969faee), no mechanism remained to prevent early pod scheduling on nodes that haven't finished initialization. Pods could land on nodes before iptables cache-enforcement rules, registry mirror config, or performance tuning were applied — causing cache misses, pull failures, or degraded performance.


huydhn:
@claude Review osdc/base/kubernetes/node-taint-remover/lib/taint_remover.py script
* Are there any condition where _wait_for_instance_type_taint or remove_taint_forever might be stuck in the while loop?
* Is the way to query k8s API ok?
* What's about the TOKEN_PATH and CA_PATH? Do we need them

jeanschmidt:
* 1 - Fixed accepting 401/409/422 as retriable
* 2 - not a real issue, given in production the token lasts 1h, and if the taint is not removed in 1h (what should take seconds) it won't be solved, as virtually almost certainly there is an unrecoverable error at this point. Still it is basically 1LOC change, so why not...? Fixed.
* 3 - the wait forever semantics is intentional. focus on what is actually broken than just adding noise.

huydhn (approved):
LGTM!

PR #709, opened 2026-06-07
[alerts] add per-cluster Mimir rule namespaces + alert routing labels

georgehong:
**Context:**
After updating the key, we're now able to see the alerts in Grafana, with examples of the following form: 
```
alloy/monitoring/arc-alerts/<SOME_HASH>
...
```
I'm noticing the following issues:
1. these alerts seem to be constantly created/destroyed: *Failed to load rules from group arc in alloy...* after clicking on such a rule in the UI.
2. there are multiple copies. Each cluster seems to emit their own version: `alloy/monitoring/gpu-alerts/<SOME_HASH_1, SOME_HASH_2>`, and this seems to contest.

To address this:
- set mimir_namespace_prefix to the cluster name so clusters stop contesting/recreating each other's rules in the shared Mimir tenant. New name should be `<CLUSTER_NAME>/monitoring/arc-alerts/<SOME_HASH>`
- alerts: add team=pytorch-dev-infra-osdc to all rules and priority=P1 to critical rules, for the alerting-infra issue pipeline. The label pytorch-dev-infra was not directly chosen since that would automatically flag it with treehugger. 

After this change, I'm going to add the delivery system wiring for one of the clusters and a subset of the alerts. Afterwards, the label can be updated if all succeeds.

```
just lint && just test
```

### DIFF
--- a/osdc/modules/monitoring/kubernetes/alerts/arc-alerts.yaml
+++ b/osdc/modules/monitoring/kubernetes/alerts/arc-alerts.yaml
@@ -14,6 +14,8 @@ spec:
           for: 15m
           labels:
             severity: critical
+            team: pytorch-dev-infra
+            priority: P3
           annotations:
             summary: "ARC listener stalled — jobs queued but no progress"
             description: "ARC listener has {{ $value }} assigned jobs but the assignment rate has been zero for 15 minutes. Runners may not be picking up new jobs."
--- a/osdc/modules/monitoring/helm/alloy-values.yaml
+++ b/osdc/modules/monitoring/helm/alloy-values.yaml
@@ -114,6 +114,7 @@ alloy:
       mimir.rules.kubernetes "grafana_cloud" {
         address  = env("GCLOUD_BASE_URL")
         tenant_id = env("GCLOUD_RW_API_USER")
+        mimir_namespace_prefix = env("CLUSTER_NAME")
 
         basic_auth {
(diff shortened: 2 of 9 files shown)


huydhn:
> alerts: add team=pytorch-dev-infra-osdc to all rules and priority=P1 to critical rules, for the alerting-infra issue pipeline. The label pytorch-dev-infra was not directly chosen since that would automatically flag it with treehugger.

Would it be a good thing to get it flags with TreeHugger?  It seems easier to just use the right team name `pytorch-dev-infra` here, but set the priority to a a very low one at the bottom of the barrel as only high-prio alerts are looked into anyway.  Then, we can bump the priority later

huydhn:
@claude How does mimir_namespace_prefix work? I see that it's set to the cluster name, but it doesn't seem to be referred to anywhere in the PR. Is that a correct change?

osdc/modules/monitoring/kubernetes/alerts/arc-alerts.yaml
```diff
@@ -14,6 +14,8 @@ spec:
           for: 15m
           labels:
             severity: critical
+            team: pytorch-dev-infra-osdc
```
huydhn: Setting the correct team name `pytorch-dev-infra`, but lower the priority to P2 or P3 here makes more sense to me.  We don't look into low priority alerts anyway AFAIK

huydhn (approved):

PR #710, opened 2026-06-08
Add proactive_capacity_max clusters.yaml option

zxiiro:
Adds a per-cluster ceiling for ARC runner proactive capacity.
Setting proactive_capacity_max: N in clusters.yaml causes each scale
set to render min(def_proactive_capacity, N). Setting it to 0 disables
warm-pool pre-provisioning entirely. Leaving it unset (default) leaves
every runner def value intact. The cap never raises capacity.

This allows gradual dial-down (e.g. 10 → 5 → 0) rather than a cliff
edge, which is needed to safely reduce warm-pool cost over multiple
deploys.

The staging cluster sets proactive_capacity_max: 0 so no placeholders
are pre-provisioned there — only on-demand pairs created for in-flight
jobs.

### DIFF
--- a/osdc/modules/arc-runners/scripts/python/generate_runners.py
+++ b/osdc/modules/arc-runners/scripts/python/generate_runners.py
@@ -159,8 +159,9 @@ def generate_runner(def_file, template_content, cluster_config, output_dir, modu
     proactive_capacity = runner.get("proactive_capacity", 0)
     max_burst_capacity = runner.get("max_burst_capacity", 0)
     node_fleet_override = runner.get("node_fleet")
-    if cluster_config.get("force_proactive_capacity_zero"):
-        proactive_capacity = 0
+    proactive_cap = cluster_config.get("proactive_capacity_max")
+    if proactive_cap is not None:
+        proactive_capacity = min(proactive_capacity, proactive_cap)
 
     if not runner_name or not instance_type:
         log_error(f"Invalid definition file: {def_file}")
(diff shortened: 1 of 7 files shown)


zxiiro:
We want to use this on the LF cluster to allow it to scale down to 0 when there is low activity in PyTorch CI.

osdc/modules/arc-runners/scripts/python/generate_runners.py
```diff
@@ -390,8 +390,15 @@ def main():
     runner_image_tag = resolve_value(cluster_cfg, defaults, "arc.runner_image_tag") or "2.333.1"
     cluster_config["runner_image"] = f"ghcr.io/actions/actions-runner:{runner_image_tag}"
 
-    # Staging clusters: force proactive_capacity to 0 regardless of runner def value
-    cluster_config["force_proactive_capacity_zero"] = "staging" in cluster_id
+    # Staging clusters always get proactive_capacity=0; production clusters can
+    # opt in via force_proactive_capacity_zero: true in clusters.yaml.
+    cluster_config["force_proactive_capacity_zero"] = "staging" in cluster_id or bool(
```
huydhn: IMO, treating staging differently here looks weird and it's a gotcha. Why not set `force_proactive_capacity_zero=true` in cluster.yaml on staging cluster.

Also, setting a default to false makes sense, but I guess `bool(cluster_cfg.get("force_proactive_capacity_zero"))` has already done that?

huydhn:
The change looks ok, but I think the bigger discussion here is that, by allowing LF cluster to have 0 proactive capacity, we would have different queueing expectation / behaviour between different prod clusters, which is hard to explain to our users.

Another questions is about the existing scale up lambda. It supposes to have similar proactive capacity IIRC, do we do the same step there on LF fleet? If no, I wonder why we need this now.

jeanschmidt:
I believe we should have proactive capacity to 0. This seems a free cost savings both for LF fleet and Meta fleet. This should work given we leverage hud for scaling AND in the case of its failure we have a hardcoded 'hud offline' proactive capacity as a recovery mechanism.

It is important to stress out that ARC, as currently designed, does not work without spare capacity. But HUD should give a decent proxy for this.

Another point, we probably don't want to go from 10 -> 0 in one deployment.

So I believe this change should be broken into two parts:
- Allow to configure, per cluster, proactive capacity size, so we can safely scale down to 0;
- Have a emergency mechanism, independent of GH, where we can quickly set back capacity to higher levels if HUD is working, but not as expected;

zxiiro:
> Another questions is about the existing scale up lambda. It supposes to have similar proactive capacity IIRC, do we do the same step there on LF fleet? If no, I wonder why we need this now.

The LF Scale Up Lambda today is configured to scale down to 0.

osdc/modules/arc-runners/scripts/python/generate_runners.py
zxiiro: Yeah I was thinking about this when I made the PR too. I wasn't sure if we should keep the staging code here or explicitly setting this setting in clusters.yml.

I do think making the cluster set it up is more clean so let me update this PR to update the cluster.yml side and remove this condition in code.

zxiiro: I removed the conditional statement. In code so now all this is configured explicitly in clusters.yml

zxiiro:
I took this PR into another direction. Introducting a different parameter `proactive_capacity_max` which I believe will satisfy @jeanschmidt 's request allow to ramp down the capacity our a set amount.

Setting it to 0 will effectively lower capacity to 0 like the previous parameter. However we can now also set any other number to set a max proactive capacity allowing us to scale down more cautiously. This one setting applies to all runner definitions setting the same max proactive value for all runners.

Minimum proactive capacity is still managed by the runner definition if the max is set to a higher value so the max setting only sets the ceiling for proactive provisioning.

jeanschmidt (approved):

huydhn (approved):

Issue #696, opened 2026-06-03
H100 4-GPU jobs accepted then fail with TopologyAffinityError (NUMA-blind acceptance path)


huydhn:
## Summary

H100 (4-GPU) jobs on the `arc-cbr-production-uw1` cluster can be **accepted by GitHub and then fail immediately** with `TopologyAffinityError` because the job-acceptance path is NUMA-blind. We should gate acceptance on NUMA feasibility so these jobs queue instead of failing.

## Incident

- Job: `pytorch/helion` run `26856066015`, runner `mt-l-x86iamx-88-900-h100-4-2pfzt-runner-26mj5`, cluster `pytorch-arc-cbr-production-uw1`, node `ip-10-8-98-255` (p5.48xlarge).
- The runner pod started fine; the `-workflow` pod (the 4-GPU job container) was rejected by the kubelet: `TopologyAffinityError: Resources cannot be allocated with Topology locality`.
- Not a one-off: the same node rejected **4 consecutive 4-GPU workflow pods in ~90s** (including a capacity placeholder) — a fragmentation livelock.

## Root cause

The `p5-48xlarge` pool pins `topology_manager_policy: single-numa-node` (`scope: pod`). A p5.48xlarge has **2 NUMA sockets x 4 H100 each**, so a 4-GPU pod needs an **entirely empty socket**. The pool is shared by 1/2/4-GPU runners; the NUMA-blind scheduler sprinkles small runners across both sockets, and once >=1 GPU is used on each socket, **no 4-GPU job can be admitted** even though 4 GPUs are free in aggregate.

## Why current mechanisms don't catch it

- **GitHub assignment** is label-based; **ARC scaling** and the **capacity-aware placeholder gate** all reason in *scalar* `nvidia.com/gpu` counts + pod phase — none are NUMA-aware.
- The capacity gate (`runningRunners + runningPairs`) is correct in aggregate, but a fragmented node is a **trap**: scalar-free but NUMA-split. The NUMA-blind kube-scheduler *prefers* it (no preemption needed) and routes the real workflow pod there, bypassing the validated placeholder slot. The kubelet — the only NUMA-aware actor — rejects it last, after the job is already accepted.

## Options to discuss

[...]

PR #716, opened 2026-06-09
Add NFD topology-updater and numa-scheduler modules (#696)

georgehong:
* #737
* __->__ #716

Adds two new OSDC modules for NUMA-aware GPU scheduling:

1. nfd — deploys NFD topology-updater as a DaemonSet on GPU nodes.
   Publishes NodeResourceTopology CRDs showing per-NUMA-zone resource
   availability (GPU/CPU/memory per socket). Polls every 15s to stay
   ahead of the ARC capacity provisioner's 30s interval.

2. numa-scheduler — deploys a secondary kube-scheduler with the
   NodeResourceTopologyMatch plugin. Reads NRT objects and only places
   pods on nodes where a single NUMA zone can satisfy the full resource
   request. Uses MostAllocated scoring to pack small pods together and
   keep free sockets available for large (4-GPU) jobs.

The scheduler-plugins chart is downloaded via curl from the GitHub
release (kubernetes-sigs/scheduler-plugins) at deploy time — the
project does not publish to a Helm registry or OCI.

Both modules are inert until added to a cluster's `modules:` list in
clusters.yaml (matching the standard module enablement pattern — no
separate `enabled` flag). Deploying them has no effect on existing
scheduling until runner definitions set scheduler_name: numa-scheduler
on workflow pods.

Includes smoke tests for both modules (namespace, Helm release,
DaemonSet/Deployment health, NRT CRD existence, scheduler pod
placement on base-infrastructure nodes).

Part of https://github.com/pytorch/ci-infra/issues/696

**Dry-Run** both targets for `helm upgrade`:
```
# pointed at arc-staging
helm upgrade --install nfd nfd/node-feature-discovery --version 0.17.1 --namespace nfd --create-namespace -f modules/nfd/helm/values.yaml --dry-run=server
...
STATUS: pending-install
...
```
```
curl -fsSL "https://github.com/kubernetes-sigs/scheduler-plugins/releases/download/v0.34.7/scheduler-plugins-0.34.7.tgz" -o /tmp/scheduler-plugins-0.34.7.tgz

helm upgrade --install numa-scheduler /tmp/scheduler-plugins-0.34.7.tgz --namespace numa-scheduler --create-namespace -f modules/numa-scheduler/helm/values.yaml --set scheduler.replicaCount=2 --dry-run=server
...
STATUS: pending-install
...
```
**Follow-Ups**
1. ARC fork (jeanschmidt/actions-runner-controller): Add SchedulerName field to capacity config + placeholder pod builder. New env var CAPACITY_AWARE_SCHEDULER_NAME, defaults to empty (no behavior change). Publish as new chart version.

2. Deploy updated ARC controller:
    - Update arc.chart_version in clusters.yaml
    - just deploy-module <cluster> arc

3. Wire schedulerName into runner template (runner.yaml.tpl):
    - Add CAPACITY_AWARE_SCHEDULER_NAME to listener env
    - Add schedulerName: {{SCHEDULER_NAME}} to job-pod.yaml

4. Activate for H100 runners.


georgehong:
planning some additional changes to remove the gh dependency to extract the release (i.e. use curl instead), and add some additional module tests.

georgehong:
Add additional smoke tests, and similar to the cluster deployment, if this is validated, I can try running this on arc-staging to ensure deployment works as expected before continuing.

osdc/modules/nfd/helm/values.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - key: nvidia.com/gpu
+      operator: Exists
+      effect: NoSchedule
+    - key: node-fleet
+      operator: Exists
+      effect: NoSchedule
+    - key: instance-type
+      operator: Exists
+      effect: NoSchedule
+    - key: git-cache-not-ready
```
huydhn: This has been removed, so you will need to do a rebase. See this here seems to indicate that you want to add this into this new list too https://github.com/pytorch/ci-infra/blob/main/osdc/modules/nodepools/scripts/python/generate_nodepools.py#L55

huydhn:
@claude This PR introduces a new osdc/modules/numa-scheduler/helm/values.yaml, how does that work? Is it always enforced or can it be optional?

A couple more things that I have in mind:

* 8xH100 or 8xB200 jobs use all the NUMA cores from the server (2). Do they still work with the new schedule?
* Some CI jobs doesn't care about NUMA core, only the number of GPUs. So, they are ok if CPU are scattered as long as the requested number of GPUs are there. It's better to run them right away instead of waiting, or worst failing the jobs

huydhn:
@claude 

> 3. Jobs that only care about GPU count (CPU may be scattered)
Same mechanism solves it: don't route those jobs through numa-scheduler. Leave schedulerName unset and they go through the default scheduler — placed immediately on any node with enough free GPUs, CPUs allowed to scatter across sockets. No waiting, no NUMA filtering.
The only failure mode to avoid is opting a NUMA-agnostic job into this scheduler on a single-numa-node node — then a request that spans sockets would stay Pending. As long as opt-in stays selective (only the latency-sensitive single-socket jobs that actually benefit from socket-local GPU+CPU+memory), the "scatter is fine, run now" jobs are untouched.

For this point, what is the impact of enabling numa scheduler on https://github.com/pytorch/ci-infra/blob/main/osdc/modules/arc-runners-h100/defs/l-x86iamx-88-900-h100-4.yaml scale set that is using https://github.com/pytorch/ci-infra/blob/main/osdc/modules/nodepools-h100/defs/p5.yaml nodepool. The nodepool has `topology_manager_policy: single-numa-node`

georgehong:
Updating to break down changes into components in stack available here: https://github.com/pytorch/ci-infra/pull/738.

PR #717, opened 2026-06-09
Document ARC stale scale set recovery runbook

zxiiro:
Add a step-by-step recovery procedure to operations.md for when AutoscalingListener pods crash-loop after config changes that alter the GitHub-side scale set registration (github_config_url or runner_group changes).

### DIFF
--- a/osdc/docs/operations.md
+++ b/osdc/docs/operations.md
@@ -126,6 +126,60 @@ just resume-runners arc-staging     # restore maxRunners from def files, remove 
 
 `recycle-nodes` deletes all Karpenter `NodeClaims`, forcing Karpenter to reprovision on demand. Use after AMI/userData changes when you do not need to preserve in-flight jobs. Staging runs this automatically at the end of `just deploy` (driven by `recycle_karpenter_nodes: true` in `clusters.yaml`).
 
+### Recovering from stale ARC scale sets
+
+Certain `arc-runners` config changes (e.g. changing `github_config_url` from
+repo-scoped to org-scoped, or changing `runner_group`) cause the ARC controller to
+register new GitHub-side scale sets while old `AutoscalingListener` objects remain
+in the cluster pointing at the now-invalid scale set IDs. Symptoms: listener pods
+crash-loop with `RunnerScaleSetNotFoundException` (404 from GitHub API), or two
+listener pods exist for the same runner set with different spec hashes.
+
+**Step 1 — Check for duplicate or erroring listeners:**
+
+```bash
+kubectl get autoscalinglisteners -n arc-systems
+kubectl get pods -n arc-systems | grep -v Running
+```
+
+If you see two listeners for the same runner set (different hash suffixes), or pods
+cycling through `Error` / `ContainerCreating`, proceed with the cleanup.
+
+**Step 2 — Delete all `AutoscalingRunnerSet` objects to force re-registration:**
+
+```bash
+kubectl delete autoscalingrunnersets --all -n arc-runners
+```
+
+This removes the in-cluster ARS objects. The ARC Helm releases are untouched — the
+controller recreates ARS objects (and fresh GitHub scale set IDs) on the next deploy.
+
+**Step 3 — Force-redeploy arc-runners:**
+
+```bash
+HELM_FORCE_UPGRADE=1 just deploy-module <cluster> arc-runners
+```
+
+`HELM_FORCE_UPGRADE=1` bypasses the skip-if-no-diff logic so the upgrade runs even
+when the rendered templates haven't changed.
+
+**Step 4 — Clean up any remaining stale listeners:**
+
+After the redeploy, new listeners get a fresh spec hash. Old listeners (prior hash)
+may linger briefly. Delete them explicitly:
+
+```bash
+OLD_HASH=<old-hash>   # e.g. 58d9767d — visible in the listener name
+kubectl delete autoscalinglistener -n arc-systems \
+  $(kubectl get autoscalinglisteners -n arc-systems --no-headers | grep "$OLD_HASH" | awk '{print $1}')
+```
+
+All listener pods should reach `Running` within ~30 seconds.
+
+> **Note:** For `runner_group`-only changes (no URL change), you typically do **not**
+> need to delete the ARS objects — a `HELM_FORCE_UPGRADE=1` redeploy is sufficient.
+> Only stale listeners need manual cleanup in that case.
+
 ### Read-only debugging
 
 ```bash


huydhn:
@claude Check if there is already an OSDC just command to clean up stale scale set in OSDC. The problem described here looks familiar

huydhn:
I think we should just refer to `just heal-arc` here if this is meant to be a runbook doc. More details of what happens behind the scene is good though

osdc/docs/operations.md
```diff
@@ -156,6 +156,60 @@ just resume-runners meta-staging-aws-uw1     # restore maxRunners from def files
...
+kubectl get pods -n arc-systems | grep -v Running
+```
+
+If you see two listeners for the same runner set (different hash suffixes), or pods
+cycling through `Error` / `ContainerCreating`, proceed with the cleanup.
+
+**Step 2 — Delete all `AutoscalingRunnerSet` objects to force re-registration:**
+
+```bash
+kubectl delete autoscalingrunnersets --all -n arc-runners
```
jeanschmidt, line 181: better not instruct agents to do that

---

<details>
<summary>🔴 `kubectl delete autoscalingrunnersets --all -n arc-runners` wipes the h100/b200 modules' scale sets, which Step 3's single-module redeploy never recreates (ai-generated section)</summary>

Three modules — `arc-runners`, `arc-runners-h100`, `arc-runners-b200` — all install their Helm releases into the single hardcoded `arc-runners` namespace, so `--all` deletes every ARS from all three. Step 3 (line 190) redeploys only the `arc-runners` module, leaving the GPU scale sets deleted with nothing recreating them. `osdc/modules/arc-runners/deploy.sh:105` hardcodes the namespace: `helm_upgrade_if_changed "arc-${runner_name}" arc-runners`. `osdc/modules/arc-runners-h100/deploy.sh:13` and `osdc/modules/arc-runners-b200/deploy.sh:13` both `exec` that same script, overriding only `ARC_RUNNERS_DEFS_DIR`/`ARC_RUNNERS_OUTPUT_DIR`/`ARC_RUNNERS_MODULE_NAME` — never the namespace. `osdc/clusters.yaml:593-595` enables all three modules on the production cluster `meta-prod-aws-ue2`; the h100 and b200 defs directories hold 8 and 4 runner definitions respectively. Unlike the ConfigMaps (`osdc/modules/arc-runners/templates/runner.yaml.tpl:390` carries `osdc.io/module`), the ARS objects themselves carry no module label, so `--all` has no way to distinguish them. On `meta-prod-aws-ue2`, an operator following this runbook deletes all H100 and B200 GPU runner scale sets along with the CPU ones. Step 3 restores only the CPU scale sets and reports success; every GPU CI job on the cluster queues indefinitely until someone independently notices and runs `just deploy-module meta-prod-aws-ue2 arc-runners-h100` and `... arc-runners-b200`. Narrow the delete to the module being recovered, using the same ConfigMap→release mapping `remove-module` uses at `osdc/justfile:638-652`: `kubectl get cm -n arc-runners -l osdc.io/module=arc-runners -o name | sed 's|configmap/arc-runner-hook-|autoscalingrunnerset/arc-|' | xargs -r kubectl delete -n arc-runners`. Broader and preferable: drop the manual delete entirely and point at `just heal-arc &lt;cluster&gt;`, which operates per-ARS (see the second finding on line 159).

<sub>Reviewed by claude-opus-5[1m] at max effort, against 4f807d0a1.</sub>

</details>

jeanschmidt, line 181: yeah, red flag, only on very critical scenarios, all this should be stated

---

<details>
<summary>🔴 Step 2 tells operators to delete every AutoscalingRunnerSet in the cluster, aborting all in-flight CI jobs, instead of using the repo's `just heal-arc` (ai-generated section)</summary>

The new runbook's Step 2 is an unscoped `--all` delete of every `AutoscalingRunnerSet` in `arc-runners`. Each ARS owns an EphemeralRunnerSet which owns the EphemeralRunners, so the delete cascades to every runner pod on the cluster and kills every in-flight job. The runbook has no drain step, no warning, and no reference to `just heal-arc`, the purpose-built recovery this repository already ships for exactly this symptom. The following sentence at line 184 ("The ARC Helm releases are untouched") actively reassures the operator that the step is benign. `osdc/modules/arc-runners/deploy.sh:306-337` detects this exact symptom (`RunnerScaleSetNotFoundException`, line 293) and prints a banner telling the operator to run `just heal-arc ${CLUSTER} &lt;ars-name&gt;` (line 330). `heal-arc` (`osdc/justfile:1239-1449`) never deletes an ARS — it clears the `runner-scale-set-id` annotation in place (line 1303), deletes only label-matched listeners (lines 1308-1310), paces deletes, and gates on `OSDC_CONFIRM` (lines 1263-1277). `osdc/docs/arc-fork-build-deploy.md:282-342` already documents this recovery and recommends "Heal a single ARS (smallest blast radius — start here to validate)" (line 318). The repo's only other runbook with a destructive `--all` (`osdc/docs/cluster-recreation.md:56-74`) runs `just drain-runners ${CLUSTER}` first, verifies `kubectl get pods -n arc-runners ... | wc -l # must be 0`, and warns that proceeding "will cause canceled jobs". `osdc/justfile:534` states the convention: "Operator is expected to have run `just drain-runners` first for runner modules.". An on-call engineer following this page during a listener crash-loop deletes every runner scale set on the cluster. Every PyTorch CI job currently executing on that cluster — up to the 1h typical suite duration noted at operations.md:153 — is killed mid-run and must be re-queued, when the actual fault affected only the listeners of the scale sets whose config changed. Replace Step 2 with the existing tooling: `just heal-arc-check &lt;cluster&gt;` to inspect, then `just heal-arc &lt;cluster&gt; &lt;ars-name&gt;` for the affected set (or `just heal-arc &lt;cluster&gt;` for a broad event), and link to `arc-fork-build-deploy.md#stale-listener-recovery`. If a manual ARS delete is genuinely still needed for the `github_config_url` case that `heal-arc` cannot cover, scope it to the named sets (`kubectl delete autoscalingrunnerset -n arc-runners &lt;name&gt; ...`), precede it with `just drain-runners &lt;cluster&gt;`, and state plainly that it cancels in-flight jobs. I would pick the first: it reuses the reviewed, confirmation-gated path.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 4f807d0a1.</sub>

</details>

osdc/docs/operations.md
```diff
@@ -156,6 +156,60 @@ just resume-runners meta-staging-aws-uw1     # restore maxRunners from def files
 
 `recycle-nodes` deletes all Karpenter `NodeClaims`, forcing Karpenter to reprovision on demand. Use after AMI/userData changes when you do not need to preserve in-flight jobs. Staging runs this automatically at the end of `just deploy` (driven by `recycle_karpenter_nodes: true` in `clusters.yaml`).
 
+### Recovering from stale ARC scale sets
```
jeanschmidt, line 159: yeahp, we now have just heal-arc, more effective and simple

---

<details>
<summary>🟠 The new runbook contradicts the `just heal-arc` recovery path that arc-runners/deploy.sh prints and arc-fork-build-deploy.md documents for the identical symptom (ai-generated section)</summary>

The repository already ships automated recovery for exactly this symptom, and the deploy script tells operators to use it. This section documents an incompatible manual procedure and never mentions `just heal-arc`, `just heal-arc-check`, or the existing runbook, so the tooling and the docs now point operators in two different directions from the same failure. `osdc/modules/arc-runners/deploy.sh:293` greps listener logs for `RunnerScaleSetNotFoundException` — the same string this section names at operations.md:165 — and lines 326-334 print a banner ending in `just heal-arc ${CLUSTER} &lt;ars-name&gt;` / `just heal-arc ${CLUSTER}`. `osdc/docs/arc-fork-build-deploy.md:282-342` is the existing "Stale Listener Recovery" runbook for the same symptom, listing the five drift categories `heal-arc` detects and recovers. `osdc/justfile:1239-1449` implements it: it clears the `runner-scale-set-id` annotation rather than deleting the ARS, polls for the controller to stamp a fresh ID before cascading, and verifies convergence. Nothing in this diff references any of it. An operator who reaches operations.md first — the day-to-day doc — runs the destructive `--all` delete instead of `just heal-arc`, which recovers the same condition without deleting ARS objects. The two docs will also drift independently as ARC behaviour changes. Make this section a pointer rather than a parallel procedure: document `just heal-arc-check &lt;cluster&gt;` / `just heal-arc &lt;cluster&gt; [&lt;ars-name&gt;]` as the first step, link to `docs/arc-fork-build-deploy.md#stale-listener-recovery`, and keep the manual steps only as a clearly-labelled fallback for the case `heal-arc` does not cover (a `github_config_url` change, where the annotation is present and syntactically valid).

<sub>Reviewed by claude-opus-5[1m] at max effort, against 4f807d0a1.</sub>

</details>

osdc/docs/operations.md
```diff
@@ -156,6 +156,60 @@ just resume-runners meta-staging-aws-uw1     # restore maxRunners from def files
...
+kubectl delete autoscalingrunnersets --all -n arc-runners
+```
+
+This removes the in-cluster ARS objects. The ARC Helm releases are untouched — the
+controller recreates ARS objects (and fresh GitHub scale set IDs) on the next deploy.
+
+**Step 3 — Force-redeploy arc-runners:**
+
+```bash
+HELM_FORCE_UPGRADE=1 just deploy-module <cluster> arc-runners
```
jeanschmidt, line 190: <details>
<summary>🟡 Runbook documents the internal `HELM_FORCE_UPGRADE` env var instead of the supported `force` positional argument of `just deploy-module` (ai-generated section)</summary>

The repository's documented interface for forcing a Helm upgrade is the third positional argument to `deploy-module`. `HELM_FORCE_UPGRADE` is the internal variable the recipe exports on the operator's behalf, and `osdc/docs/modules.md:43` is the one place that documents that mapping. Lines 190, 193 and 210 copy the internal variable name into a third document as the operator-facing command. operations.md:104-105, 55 lines above this hunk, already documents the supported form: "# Force a Helm upgrade even when the release looks unchanged" / "just deploy-module meta-staging-aws-uw1 arc force". modules.md:43: "`deploy-module` accepts an optional third positional `force` argument; setting it to any non-empty value ... exports `HELM_FORCE_UPGRADE=1` for the duration of the run, which the shared `helm-upgrade.sh` helper honours". The implementation is osdc/justfile:438-439 (`if [[ -n "{{force}}" ]]; then export HELM_FORCE_UPGRADE=1`) and osdc/scripts/helm-upgrade.sh:40. The variable name is now duplicated across justfile:439, helm-upgrade.sh:40, modules.md:43 and three lines here. If the recipe stops exporting that exact name — renaming it, or passing a flag to helm-upgrade.sh instead — `just deploy-module &lt;cluster&gt; arc-runners force` keeps working while this runbook silently stops forcing anything: the deploy prints its skip-if-no-diff message and the operator, mid-incident, concludes the recovery failed for some other reason. Use the documented interface at line 190: `just deploy-module &lt;cluster&gt; arc-runners force`. Then reword line 193 to "the third positional `force` argument bypasses the skip-if-no-diff logic (see [`modules.md`]​(modules.md))" and line 210 to "a `force` redeploy is sufficient", so the internal variable name appears nowhere in this file.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 4f807d0a1.</sub>

</details>

osdc/docs/operations.md
```diff
@@ -156,6 +156,60 @@ just resume-runners meta-staging-aws-uw1     # restore maxRunners from def files
...
+After the redeploy, new listeners get a fresh spec hash. Old listeners (prior hash)
+may linger briefly. Delete them explicitly:
+
+```bash
+OLD_HASH=<old-hash>   # e.g. 58d9767d — visible in the listener name
+kubectl delete autoscalinglistener -n arc-systems \
+  $(kubectl get autoscalinglisteners -n arc-systems --no-headers | grep "$OLD_HASH" | awk '{print $1}')
+```
+
+All listener pods should reach `Running` within ~30 seconds.
```
jeanschmidt, line 207: <details>
<summary>🟡 The stated ~30s recovery window ignores the per-scale-set cost; the repo's own tooling budgets 60s to settle plus a 30s sleep before it even checks (ai-generated section)</summary>

The sentence gives a fixed recovery time for an operation whose cost scales with the number of scale sets. After Step 2 deleted every ARS, the controller has to re-register each scale set against the GitHub broker and schedule one listener pod per set; nothing in the procedure makes that constant-time. `osdc/modules/arc-runners/deploy.sh:247` sets `LISTENER_TIMEOUT="${ARC_LISTENER_CHECK_TIMEOUT:-60}"` and polls up to 60s just for listeners to settle after a normal deploy. `osdc/justfile:1425-1426` sleeps 30s before it even begins verifying a heal, and justfile:1317-1324 polls up to 60s (30 × 2s) for the controller to stamp a new scale-set ID on a single ARS. With ~61 ARS on meta-prod-aws-ue2 the paced heal path alone spends ~30s just issuing the listener deletes (61 × 0.5s at justfile:1398). An operator following the runbook on a prod cluster sees listeners still not Running at the 30s mark and reads a normal, still-converging recovery as a failed one — at which point the next thing the page offers is more deletion. Replace the sentence with a scale-aware expectation, e.g. "Listener pods come back one per scale set; allow at least `ARC_LISTENER_CHECK_TIMEOUT` (default 60s, see `modules/arc-runners/deploy.sh`) after the last delete, and longer on clusters with dozens of runner defs. Re-check with `just heal-arc-check &lt;cluster&gt;` before taking further action.".

<sub>Reviewed by claude-opus-5[1m] at max effort, against 4f807d0a1.</sub>

</details>

jeanschmidt (changes requested):
## Review summary

🔴 2 blocker · 🟠 1 major · 🟡 2 minor

- 🔴 `kubectl delete autoscalingrunnersets --all -n arc-runners` wipes the h100/b200 modules' scale sets, which Step 3's single-module redeploy never recreates — `osdc/docs/operations.md:181`
- 🔴 Step 2 tells operators to delete every AutoscalingRunnerSet in the cluster, aborting all in-flight CI jobs, instead of using the repo's `just heal-arc` — `osdc/docs/operations.md:181`
- 🟠 The new runbook contradicts the `just heal-arc` recovery path that arc-runners/deploy.sh prints and arc-fork-build-deploy.md documents for the identical symptom — `osdc/docs/operations.md:159`
- 🟡 Runbook documents the internal `HELM_FORCE_UPGRADE` env var instead of the supported `force` positional argument of `just deploy-module` — `osdc/docs/operations.md:190`
- 🟡 The stated ~30s recovery window ignores the per-scale-set cost; the repo's own tooling budgets 60s to settle plus a 30s sleep before it even checks — `osdc/docs/operations.md:207`

<details>

**What this changes**

The change adds one 54-line section, "Recovering from stale ARC scale sets", to osdc/docs/operations.md. Nothing else moves; there is no code in the diff. The section documents a four-step manual procedure for AutoscalingListener pods crash-looping after a `github_config_url` or `runner_group` change: inspect listeners in `arc-systems`, delete every AutoscalingRunnerSet in `arc-runners` with `--all`, force-redeploy the `arc-runners` module with `HELM_FORCE_UPGRADE=1`, then hand-delete leftover listeners selected by substring-matching a hash against `kubectl get` table output. A closing note carves out `runner_group`-only changes as not needing the ARS delete.

**What it gets right**

The symptom description is accurate and matches what the code already detects — `modules/arc-runners/deploy.sh` greps listener logs for the same `RunnerScaleSetNotFoundException` string, and duplicate listeners with differing hash suffixes is a real signature. The trigger it names is not in the "Common triggers" list in `arc-fork-build-deploy.md`, so the scenario itself is new content rather than a restatement. The explanation of `HELM_FORCE_UPGRADE` bypassing skip-if-no-diff is correct, and the section is placed next to the other runner-lifecycle procedures, where an operator looking for it would land.

**Where the risk is**

Everything concentrates in one block of prose, and the findings fall into four tight clusters that several independent reviewers reached separately — convergence worth weighing. The largest is a scope mismatch between Step 2 and Step 3: all three `arc-runners*` modules install into the single `arc-runners` namespace, so `--all` deletes the H100 and B200 scale sets that the single-module redeploy never recreates, and on `meta-prod-aws-uw1`, where `arc-runners` is not in the module list at all, Step 3 exits 0 with "Skipping". A second cluster covers the missing drain: the ARS delete cascades to EphemeralRunnerSets and runner pods, while the sentence beside it says only that Helm releases are untouched. A third covers Step 4's shell block, where `OLD_HASH=&lt;old-hash&gt;` is a bash parse error that leaves the variable empty, `grep ""` then matches everything, and the resulting unpaced bulk delete is the pattern a justfile comment records as having exhausted CNI IPs. A fourth is that the whole section is a second, broader path for a symptom `just heal-arc` already handles and `deploy.sh` already points operators at.

The recurring class is a command written wider than the fault it repairs, described by prose written narrower than the command. The scope and drain clusters compound: an unscoped delete, no drain, and a restore step narrower than the destruction all land in one procedure.

**Before merge, and after**

The destructive mechanics are the merge-blocking part: the `--all` scope against a single-module Step 3, the absent drain or job-abort warning, the Step 4 `OLD_HASH` placeholder and unguarded expansion, and the missing `just kubeconfig &lt;cluster&gt;` before the delete steps. Several one-line edits belong here too because they are cheap: `HELM_FORCE_UPGRADE=1` versus the documented `force` argument, "~30 seconds" versus the repo's 30-60s figure, and the sentence crediting the controller with recreating Helm-owned ARS objects.

Reasonable as follow-up: whether this section should exist at all beside `arc-fork-build-deploy.md`'s Stale Listener Recovery, or fold into it — that is a structural decision, not a patch. So is the open question on the `runner_group`-only note, which turns on ARC controller behaviour that cannot be settled from this repository.

</details>

PR #718, opened 2026-06-09
Add HUD failure base capacity to runners

jeanschmidt:
- Add `hud_failure_base_capacity` runner def field, template placeholder, and env var propagation to the ARC listener
- Zero `hud_failure_base_capacity` on staging clusters, paused runners, and region-excluded instance types
- Warn when `max_burst_capacity` would cap below the HUD fallback baseline
- Bump ARC fork chart to 0.14.1-jeanschmidt.12
- Add unit tests covering default, nonzero, staging-forced, and misconfiguration-warning scenarios

Complements jeanschmidt/actions-runner-controller#8 which adds `HUDFailureBaseCapacity` to the capacity monitor's fallback formula: `desiredPairs = ProactiveCapacity * Multiplier + HUDFailureBaseCapacity`. This lets operators set
`proactive_capacity: 0` (no steady-state pre-warming) while still surging a fixed floor of placeholder pairs when the HUD API becomes unreachable — decoupling the failure-mode safety net from the normal proactive scaling budget.


zxiiro (approved):

osdc/modules/arc-runners/scripts/python/generate_runners.py
```diff
@@ -218,6 +221,13 @@ def generate_runner(def_file, template_content, cluster_config, output_dir, modu
             f"Consider raising max_burst_capacity."
         )
 
+    if max_burst_capacity > 0 and hud_failure_base_capacity > 0 and max_burst_capacity < hud_failure_base_capacity:
+        log_warning(
```
huydhn: log_error?

huydhn (approved):
LGTM! But consider treating `max_burst_capacity < hud_failure_base_capacity` as an error for clarity as I don't think we will read the warnings that carefully

malfet (approved):

PR #720, opened 2026-06-09
grafana/osdc: add K8s Warning Events table + time series

georgehong:
While working on https://github.com/pytorch/ci-infra/issues/696, we wanted to determine frequency of "TopologyAffinityError: Resources cannot be allocated with Topology locality." We can use the Loki logs to do a rough breakdown by *Events* by introducing a panel of warning volume and time series of occurrences. 
- Exact/Aggregated distinction - this exists due to Kubernetes deduplication, with full explanation: 
```
Type — Exact: Events is a true occurrence count (one event per occurrence). Aggregated: k8s merged repeats onto one object (count>1), so Events is relative volume, not an exact tally.
```
For the aggregated type of Event, the time series exists to track changes in behavior over time.

Rendered at https://pytorchci.grafana.net/dashboards/f/fk7zwm/ (under OSDC)

Authored by Claude.

### DIFF
--- a/grafana/osdc.json
+++ b/grafana/osdc.json
@@ -1436,6 +1436,337 @@
           }
         }
       }
+    },
+    "panel-12": {
+      "kind": "Panel",
+      "spec": {
+        "data": {
+          "kind": "QueryGroup",
+          "spec": {
+            "queries": [
+              {
+                "kind": "PanelQuery",
+                "spec": {
+                  "hidden": false,
+                  "refId": "A",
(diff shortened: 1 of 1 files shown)


georgehong:
Closing PR for now to address feedback that time series Loki warnings are difficult to parse. There's also some separation between exact/aggregated events that could require additional clarification, since the most frequently occurring (by far) warning (FailedScheduling) is also part of regular OSDC operation and would support further drill down.

PR #723, opened 2026-06-10
BuildKit autoscaling on staging: in-cluster KEDA + LB queue + warm baseline

huydhn:
* #727
* #728
* #726
* #725
* #724
* __->__ #723

**Impact:** OSDC arc-staging buildkit only (autoscaling is opt-in; other
clusters unchanged).
**Risk:** low

Absorb ciflow/docker bursts without overloading existing pods, and scale back to
a small warm per-arch baseline when idle.

- HAProxy `server maxconn 1` + `timeout queue`: one build per pod; excess builds
  queue and flow onto new pods as they register, instead of stacking on busy
  pods (so scaled-up pods don't sit idle).
- KEDA ScaledObject per arch via `metrics-api` scraping the LB's own metrics
  (haproxy_backend_current_sessions) — no Grafana / external metrics backend.
- Warm baseline: amd64_min=2 / arm64_min=4 (1 physical node each); *_max caps
  the burst and sizes the NodePool limits.
- preStop drain + PDB + long terminationGracePeriodSeconds for kill-free
  scale-down.

staging: amd64 m6id.24xlarge @ 2/node (min 2), arm64 m7gd.16xlarge @ 4/node
(min 4). Runner-side connect retry (separate pytorch/pytorch change) lets a build
tolerate waiting for a pod from a cold/queued pool.

**Testing**: Validation https://github.com/pytorch/ci-infra/actions/runs/27302743769 on pytorch-canary https://github.com/pytorch/pytorch-canary/actions/runs/27303571628 (comment below)

Signed-off-by: Huy Do <huydo@meta.com>


huydhn:
### Integration-test validation (updated)

**Correction to the earlier note below:** the `maxconn 1` + HAProxy `timeout queue` design does **not** let a client wait for a pod. Both `docker buildx` and `buildctl` use the `moby/buildkit` gRPC client, which dials with the default **~20 s `MinConnectTimeout`** and fail-fast RPCs (not configurable). A connection sitting in HAProxy's queue never receives the HTTP/2 server preface, so the client gives up at ~20 s regardless of `timeout queue`. So `timeout queue` was **removed**, and clients now **retry the connect** until a pod frees / the pool scales up. (Connectivity test uses `buildctl`; the scale test and prod `.ci/docker/build.sh` use `docker buildx` — same client, same retry.)

**Latest run** — [27303571628](https://github.com/pytorch/pytorch-canary/actions/runs/27303571628), deliberately over-subscribed: per arch, **1 connectivity build + an 8-wide scale burst = 9 builds against a max-8 pool**. **18/18 jobs succeeded.**

| job | result | notes |
|---|---|---|
| amd64 `build` (connectivity) | ✅ 3.0 min | won the race for a pod |
| arm64 `build` (connectivity) | ✅ 11.1 min | **odd-one-out: retried 30× (~10 min)** until a scale peer freed a pod, then built on attempt 30 |
| amd64 `scale (1..8)` | ✅ ~11–13 min | connect (after scale-up) + ~10 min hold |
| arm64 `scale (1..8)` | ✅ ~11–13 min | same |

**Validates:**
- **Connect-retry rides out scale-up _and_ over-subscription** — the 9th build waited ~10 min for a peer to finish, reconnected, and succeeded; no false failures.
- **KEDA scale-up** brought both arches to `max` off the in-cluster HAProxy session metric (the scale builds landed once pods were up).
- The gate still holds: with no scale-up, the back of the burst exhausts its retry budget / the 30-min job timeout and fails.

(Earlier run [27247502866](https://github.com/pytorch/pytorch-canary/actions/runs/27247502866) also covered kill-free scale-down to the 2/4 baseline + `WhenEmpty` node-consolidation lag — still applies.)

---

<details><summary>Original note (superseded — "queue worked / no connect timeouts" was wrong)</summary>

### Staging validation run

Drove a balanced burst of 8 amd64 + 8 arm64 builds against the staging pool (each held a `maxconn=1` slot ~5m), the same shape as the new integration-test scale test: https://github.com/pytorch/pytorch-canary/actions/runs/27247502866 — **16/16 builds succeeded.**

**BuildKit nodes / pods during the run:**

| Phase | amd64 pods | arm64 pods | buildkit nodes |
|-------|-----------|-----------|----------------|
| Baseline (before) | 2 | 4 | 2 (1 amd64 + 1 arm64) |
| Mid-burst (~T+9m) | scaling 2→8 | scaling 4→8 | climbing; 1 pod already draining |
| Peak | 8 | 8 | 6 (4× `m6id.24xlarge` + 2× `m7gd.16xlarge`) |
| After (~T+18m) | back to 2 | back to 4 | trailing down |

Observations:
- **Queue worked as intended** — every queued `buildctl` connected and rode the queue onto new pods as they registered; no connect timeouts, so no runner-side wait was needed for this burst.
- **Scale-up** — KEDA brought both arches to `max` (8/8) off the in-cluster HAProxy session metric.
- **Kill-free scale-down** — pods drained (`preStop` waited for `:1234` to go idle) and returned to the 2/4 baseline with zero failed builds.
- **Node consolidation lag (expected)** — with `consolidationPolicy: WhenEmpty`, survivor pods left some nodes half-full, so the node count trails the pod count back down rather than dropping immediately.

</details>

huydhn:
Full stack testing on staging https://github.com/pytorch/ci-infra/actions/runs/27302743769

osdc/clusters.yaml
```diff
@@ -162,15 +166,22 @@ clusters:
       github_secret_name: pytorch-arc-staging
       runner_name_prefix: "c-mt-"
     buildkit:
-      replicas_per_arch: 2
       arm64_instance_type: m7gd.16xlarge
+      arm64_pods_per_node: 4
```
jeanschmidt, line 170: why the asymmetry here? given m7gd.16xlarge is about 67% the size of m6id.24xlarge we're dedicating about 30% of the space for each?

jeanschmidt:
@claude Please review this PR. Biggest concerns:

* make sure the autoscaling is implemented in a race condition free, 
* Is there a risk of not scaling properly or failing to deploy
* are edge cases properly handled? 
* how is monitoring set up for keda, and do we have metrics to troubleshoot keda misbehave? And for haproxy?
* is the handling of disrruption on karpenter properly wired?

contex: this is how CI clients are intended to use the buildkit: https://github.com/pytorch/pytorch/pull/186955/changes, ie: connect-fail-retry.

osdc/clusters.yaml
huydhn: I copy the previous spec we use to build Docker for x86 and aarch64 here. We fielded 12xlarge for x86 but only needed 4xlarge for aarch64 image build.  So, we can fit 2 12xlarge pods in a m6id.24xlarge and 4 4xlarge pods into m7gd.16xlarge. Those are the current nodes used by Buildkit.

jeanschmidt:
@claude 

following up on https://github.com/pytorch/ci-infra/pull/723#issuecomment-4674353283 and https://github.com/pytorch/ci-infra/pull/723#issuecomment-4674359103

are there anti-flap mechanisms prevent the fast scale down of pods/nodes (keeping them warm for a while before tearing them down), explain them.

can you look into this PR and tell me if https://github.com/pytorch/pytorch/pull/186955 can safely handle failure to schedule up and wait for 2-4 average build jobs to finish? (you can look for build jobs on pytorch/pytorch for x86-64 and aarch)

huydhn:
Add KEDA fallback is an interesting point, let me see what we can do there, probably keep the current fixed number of 32 x86 and 8 aarch64 Buildkit pods, which has proven to be working ok since last Friday.

For KEDA metrics, let's do that in a separate PR. It should have some we want to collect.

jeanschmidt:
no-flap protections are too weak for a real use case IMO.

we should try to keep pods alive for **at least** 20 minutes of idling, so instead of one every 2 minutes, it makes much more sense to have max(10, 20%) every 20 minutes only for pods idle for 20+ minutes. Or something similar.

By the other hand, karpenter can be quite agressive on scaling down nodes (like they are now). We should gate the anti-flap mechanism on pods, yes they are less expensive to start, but more costly for jobs (loss of local decompressed layer caches). 

also monitoring seems a requirement :)

huydhn:
Yeah, a minimum of 20 minutes of idling can be set easily for pods by tweaking the `drain.sh` script.  Also, we only evict pods one at a time, so with a 20-minute of idling wait, it would take 10x20 = 200 minutes to teardown 10 pods for example, more than enough for the infrequent nature of building Docker images.

huydhn:
@jeanschmidt Done — scale-down now holds a pod ~20 min after it goes idle (`stabilizationWindowSeconds: 1200`), then sheds at most `max(10 pods, 20%)` per 20 min, as you suggested. Gated on pods; node churn stays with Karpenter. Monitoring is split out: KEDA operator metrics in #726, alerts in #727.

jeanschmidt (approved):

PR #724, opened 2026-06-10
buildkit: enable autoscaling on prod (arc-cbr-production)

huydhn:
* #727
* #728
* #726
* #725
* __->__ #724

Same min per arch as staging (amd64 2 / arm64 4). Max sized from 14-day
docker-build concurrency: amd64 128 (peak 105 + headroom), arm64 16
(peak 8, likely capped by the old fixed pool).


osdc/clusters.yaml
```diff
@@ -294,11 +294,15 @@ clusters:
...
       amd64_instance_type: m6id.24xlarge
-      amd64_replicas: 32
       amd64_pods_per_node: 2
       arm64_instance_type: m7gd.16xlarge
-      arm64_replicas: 8
       arm64_pods_per_node: 4
+      autoscaling:
+        enabled: true
+        amd64_min: 2   # 1x m6id.24xlarge (2 pods/node)
+        amd64_max: 128  # 14d peak 105; headroom to spare above it
```
jeanschmidt: why not look into 90 days peak?

Still we can, in theory, be safe if CI can wait for 2 or 3 jobs average lenght. This in theory would be more optimal as it allows to reuse decompressed local cache on NVME

jeanschmidt (approved):

osdc/clusters.yaml
huydhn: Sure, I can ask Claude to look into 90 days period and update these numbers if needed

huydhn: Pulled an approximate 90-day window (most recent 1000 `docker-builds` runs, ~Apr 1 → now): **amd64 peak ≈ 180, arm64 peak ≈ 13**. Rounding up and using **2× as the max capacity** → `amd64_max: 360`, `arm64_max: 30`.

`max` is just a ceiling — baseline stays at `min` (1 node/arch) and autoscaling only climbs toward it during a real burst, so the 2× headroom costs nothing at idle.

PR #727, opened 2026-06-10
monitoring: add buildkit autoscaling alerts

huydhn:
* __->__ #727
* #728

KEDA scaler/scaledobject errors (fallback risk) + HAProxy queue backlog
(pool not scaling fast enough). Uses metrics from the KEDA ServiceMonitor
and the existing buildkit-haproxy scrape.


osdc/modules/monitoring/kubernetes/alerts/buildkit-autoscaling-alerts.yaml
```diff
@@ -0,0 +1,49 @@
...
+            priority: P3
+          annotations:
+            summary: "KEDA ScaledObject {{ $labels.scaledObject }} reconcile errors"
+            description: "KEDA failed to reconcile ScaledObject {{ $labels.scaledObject }} in the last 15m; autoscaling for that arch may be stale."
+
+        # Builds stuck waiting for a pod — the pool isn't scaling up fast enough.
+        - alert: BuildkitQueueBacklog
+          expr: |
+            haproxy_backend_current_queue{proxy=~"bk_amd64|bk_arm64"} > 0
+          for: 15m
```
jeanschmidt, line 44: I suspect that during peak this one will mis fire, say batches of 2 arrive at minute 0, 3, 10, 14. 

This will fire, even they are all scaling in 7-8 minutes.

jeanschmidt: maybe we should `haproxy_backend_current_queue{proxy=~"bk_amd64|bk_arm64"} > 20` or other value

jeanschmidt:
@claude

please review the alerts, make sure they are using the correct metrics, and they don't misfire on regular non-failure scenarios.

osdc/modules/monitoring/kubernetes/alerts/buildkit-autoscaling-alerts.yaml
huydhn: Done — bumped it to `> 20`. With the 15m window that won't fire on normal burst churn (your 0/3/10/14 example stays well under 20 and drains in minutes); it only trips on a real sustained backlog the pool isn't clearing.

huydhn:
Thanks — all three fixed (verified against KEDA `v2.16.1` source):

- **Queue alert can never fire** — added `haproxy_backend_current_queue` to the buildkit-haproxy `keep` allowlist so it reaches Mimir.
- **Scaler errors** — `keda_scaler_errors_total` → `keda_scaler_detail_errors_total` (per-scaler counter, carries `scaledObject`; the `_total` one is the deprecated unlabeled grand total).
- **ScaledObject errors** — `keda_scaledobject_errors_total` → `keda_scaled_object_errors_total` (subsystem is `scaled_object`, so the old name never existed).

PR #706, opened 2026-06-05
Populate startup taints registry with base and module-gated entries

jeanschmidt:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* __->__ #706
* #707
* #705
* #704
* #703

**Impact:** All Karpenter NodePools across OSDC clusters
**Risk:** medium

## What
Populates the previously empty `STARTUP_TAINTS` registry with three concrete taint entries and restructures the registry from a module-keyed dict to a flat list supporting unconditional (base) taints alongside module-gated ones.

## Why
The startup taints framework (67273fe) and taint-remover DaemonSets (6118763) were wired up in prior commits but the registry was left empty. This commit activates the actual taints so new nodes are held from scheduling until the corresponding DaemonSets complete initialization — preventing jobs from landing on nodes before registry mirrors, performance tuning, and cache enforcement are ready.

Signed-off-by: Jean Schmidt <contato@jschmidt.me>

PR #731, opened 2026-06-11
Fix smoke test for catch-all toleration matching

jeanschmidt:
Fix regression: https://github.com/pytorch/ci-infra/actions/runs/27315703174/job/80698504415#step:7:177

introduced by stack: https://github.com/pytorch/ci-infra/pull/706

- Replace simple key-set lookup with proper K8s toleration matching logic
- Handle catch-all tolerations (operator=Exists with no key) that match every taint
- Add cpu-type and CriticalAddonsOnly to required taint keys

The previous test extracted toleration keys into a flat set and checked membership. This broke when the DaemonSet uses a catch-all toleration (no key + operator: Exists), which K8s treats as matching all taints but never appeared in the key set. The new `tolerates()` helper mirrors actual K8s scheduling semantics.


huydhn (approved):
LGTM!

PR #734, opened 2026-06-11
Show queued jobs throughout their wait period

jeanschmidt:
- Rewrite queue time distribution queries to emit a data point every 5-minute bucket from created_at until started_at (or now for still-queued jobs), replacing the old single-point-at-start approach
- Include currently queued/waiting jobs with empty runner_group_name so they appear before being assigned
- Remove scale_set variable from panel titles and filters, use runner_name_prefix via labels array instead
- Update panel descriptions to explain the bucketing behavior

Previously each job only appeared as a single dot when it started running. Long-queued jobs were invisible during their entire wait, making it impossible to see queue pressure in real time. The new approach shows each job's growing queue duration across every 5-minute bucket of its wait, giving immediate visibility into queue depth and stalled jobs.

**Note on `runner_name_prefix` filter:** the previous filter matched against `runner_name`, which is only populated after a job is assigned to a runner. That's incompatible with surfacing still-queued jobs (the whole point of this change), so the filter now matches against `labels` instead.

view: https://pytorchci.grafana.net/d/ci-infra-f4vfmq-osdc/osdc


jeanschmidt:
@claude Please review the query changes, if there are gotchas, cases not handled, if it is written in a way that makes its performance reasonable, etc

malfet (approved):
runner prefix is a valid criticism, please mention in PR description before merging (or perhaps address)

PR #746, opened 2026-06-12
arc: move hooks-warmer DaemonSet from arc-runners to arc module

jeanschmidt:
**Impact:** clusters with `arc` but without `arc-runners` (e.g. UW1) **Risk:** low

## What
Relocates the `hooks-warmer` DaemonSet and its smoke tests from `modules/arc-runners/` to `modules/arc/`, so clusters that enable `arc` without `arc-runners` still get the hooks-warmer deployed.

## Why
The UW1 cluster runs `arc` but not `arc-runners`. The hooks-warmer DaemonSet — which pre-downloads patched runner-container-hooks onto c7i-runner nodes — was incorrectly scoped to `arc-runners`, making it unavailable on UW1. It belongs to the `arc` module because it is a prerequisite for any runner pod, not specific to the runner scale-set definitions managed by `arc-runners`.


huydhn (approved):
LGTM! This moves the hooks-warmers from one place to another.  Let's verify it on staging before landing

PR #749, opened 2026-06-13
ue1: remove pypi-cache and cache-enforcer; runners use public PyPI

huydhn:
Drop `pypi-cache` and `cache-enforcer` from `meta-prod-aws-ue1`; its runners install from public PyPI directly. `arc-runners` gates the pypi-cache pip/uv env on module membership (default on), so every other cluster is unchanged.

**Why:** the cache-enforcer SNI egress rule is non-deterministically defeated by kube-proxy resync, causing flaky `spmd-types` failures installing nightly torch (ci-infra#745). ue1 moves off the cache+enforcer stack.

**Rollout** (`just deploy` doesn't prune; delete the live resources by hand while the region is drained, so nothing breaks and new nodes come back clean). Merge first, then:
1. `just drain-runners meta-prod-aws-ue1` — `maxRunners=0`, taint, wait for in-flight. Jobs route to siblings. Drain bridges the window where the cache is gone but the runner env isn't flipped yet.
2. Delete the live k8s resources while drained (`remove-module` doesn't cover these — no `osdc.io/module` label):
   - `kubectl delete -k osdc/modules/cache-enforcer/kubernetes` (DaemonSet so it can't reschedule onto fresh nodes)
   - `kubectl delete namespace pypi-cache` + its karpenter `nodepool`/`ec2nodeclass`
3. Deploy ue1 via CI — runners re-render gated off; the two modules aren't redeployed (already gone → new nodes come up clean).
4. `just resume-runners meta-prod-aws-ue1` + `just untaint-nodes meta-prod-aws-ue1`. Verify a canary job installs from public PyPI.

**Follow-up (not cutover-critical):** `tofu destroy` the `modules/pypi-cache/terraform` root (IRSA + EFS wheelhouse) once the region is healthy — orphaned infra, only costs money, and keeping it out of the cutover keeps rollback cheap.

**Rollback:** re-add both modules to ue1 in `clusters.yaml` and redeploy (pypi-cache infra rebuilds from its tofu root).

**Test:** `just lint` + `just test` pass; ue1 renders no pypi-cache env, arc-cbr-production unchanged.

### DIFF
--- a/osdc/clusters.yaml
+++ b/osdc/clusters.yaml
@@ -241,15 +241,16 @@ clusters:
       github_secret_name: meta-prod-aws-ue1
       runner_name_prefix: "mt-"
       runner_group: "meta-prod-aws-ue1"
-    pypi_cache:
-      replicas: 10
+    # ue1 runs without pypi-cache and cache-enforcer: runners install from
+    # public PyPI directly (no wheel cache, no SNI egress firewall). The
+    # arc-runners generator gates the pypi-cache pip env on pypi-cache module
+    # membership, so omitting the module also drops the PIP_INDEX_URL/UV_*
+    # injection -- otherwise pip would target a removed service.
     modules:
       - karpenter
       - arc
       - nodepools
       - arc-runners
-      - pypi-cache
-      - cache-enforcer
       - zombie-cleanup
       - harbor-cache-recovery
       - monitoring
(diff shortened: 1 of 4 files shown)


huydhn:
Replaced by https://github.com/pytorch/ci-infra/pull/754

PR #752, opened 2026-06-15
Rewrite just remove-module into a 5-phase teardown

jeanschmidt:
* #754
* #753
* __->__ #752

Old recipe (24 lines) only deleted ConfigMaps/NodePools/EC2NodeClasses
labelled osdc.io/module=<module> and helm-uninstalled arc-runner-hook
helm releases. That made it a no-op for most modules whose resource
set is broader. Combined with no terraform teardown, no namespace
cascade, and no operator guidance, removing a module reliably required
manual kubectl + tofu work.

New recipe phases:
  1. helm uninstall arc-runner-hook-* releases (unchanged, runs first
     so the ConfigMap → release mapping survives the sweep)
  2. kubectl delete -A on namespaced kinds (deployment, daemonset,
     statefulset, service, configmap, secret, serviceaccount, role,
     rolebinding, networkpolicy, pdb, pvc)
  3. kubectl delete on cluster-scoped kinds (clusterrole,
     clusterrolebinding, storageclass, nodepool.karpenter.sh,
     ec2nodeclass.karpenter.k8s.aws)
  4. kubectl delete namespace -l osdc.io/module=<module> (cascade
     defense-in-depth; only ever deletes labelled namespaces)
  5. tofu destroy on modules/<module>/terraform/main.tf if present.
     Sub-roots like pypi-cache/terraform/wheel-cache-bucket/ are
     account-wide and explicitly skipped.

Operator UX:
- Prints a plan + a live snapshot of matching resources before any
  destructive action, gated by OSDC_CONFIRM=yes|no|ask (default ask).
- Tracks PARTIAL failures and exits non-zero if any phase failed.
- Module-specific warnings:
  * cache-enforcer: redeploy nodepools BEFORE removal (else the
    cache-enforcer startup taint stays in NodePool specs and strands
    new nodes); recycle nodes AFTER (iptables CACHE_ENFORCER chain
    persists on existing nodes until replacement).
  * pypi-cache: drain runners + recycle nodes AFTER removal (stale
    PIP_INDEX_URL env vars on in-flight pods → DNS NXDOMAIN); EFS
    wheelhouse wiped but recoverable from S3 via wheel-syncer; the
    monitoring module owns a pypi-cache ServiceMonitor that survives
    independently and goes orphan.

Depends on the previous commit's labels — without them this recipe
would still match nothing for pypi-cache / cache-enforcer.


osdc/justfile
```diff
@@ -531,34 +531,254 @@ deploy-module cluster module force="":
...
+        echo "(after the nodepools redeploy above):"
+        echo "  just recycle-nodes $CLUSTER"
+        echo ""
+        echo "Verify a node is still affected (any non-empty output = chain present):"
+        echo "  kubectl debug node/<node> -it --image=alpine -- chroot /host sh -c \\"
+        echo "    'iptables -S CACHE_ENFORCER 2>/dev/null; ip6tables -S CACHE_ENFORCER 2>/dev/null'"
+        echo ""
+    fi
+    if [[ "$MODULE" == "pypi-cache" ]]; then
+        echo "━━━ OPERATOR WARNING: stale env vars on existing runner pods ━━━"
```
huydhn, line 750: Yeah, this part is exactly the reason I want to drain traffics to the region before deleting `pypi-cache`

huydhn (approved):
LGTM!

Nit: It would be nicer if we could move the sections for `cache-enforcer` and `pypi-cache` to another PR to decouple them from the 5-phase teardown change. It's easier to clean up later that way once we remove `cache-enforcer` and `pypi-cache` everywhere

PR #766, opened 2026-06-16
arc-runners: smoke test + operator docs for runner version lock

jeanschmidt:
* __->__ #766
* #765
* #764

**Impact:** adds `just smoke` validation for the runner-version lock ConfigMap and an operator doc. No runtime behavior change.
**Risk:** low

## What
Adds the live-cluster smoke test that validates the `arc-runner-version-lock` ConfigMap (written by `resolve_runner_version.py`) has an entry matching the current OSDC SHA and that the recorded `tag@digest` matches the runner image actually deployed in every `AutoscalingRunnerSet`. Adds a short operator doc covering the per-commit pinning model, the 20-SHA window, rollback flow, and failure modes.

## Why
PR 1 added the resolver script (keyed by OSDC commit SHA). PR 2 wired it into `deploy.sh` and switched `generate_runners.py` to read `$RUNNER_IMAGE`. This PR closes the loop:
- Smoke validates that the entry the resolver wrote for THIS commit's SHA is the same `tag@digest` actually running in pod manifests (drift detection between the lock and deployed state for this commit).
- The operator doc gives one page to look at when something fails.

## Notes
- The smoke test lives under `modules/arc-runners/tests/smoke/` so `just test` skips it (per the existing `find ... -path '*/modules/*/tests/smoke' -prune` rule). It runs only under `just smoke <cluster>`.
- AutoscalingRunnerSet is the CRD the chart produces; the runner container is named `runner` in `spec.template.spec.containers`. The test reuses `run_kubectl` rather than rolling its own k8s client to keep the surface identical to every other smoke test.

### DIFF
--- a/osdc/clusters.yaml
+++ b/osdc/clusters.yaml
@@ -54,7 +54,6 @@ defaults:
     pdb_min_available: 1
   arc:
     chart_version: "0.14.1-jeanschmidt.12"  # MUST match for controller + runner charts (minor version mismatch = ARS deletion). Fork chart published from jeanschmidt/actions-runner-controller.
-    runner_image_tag: "2.334.0"  # ghcr.io/actions/actions-runner tag — pin to avoid :latest
     replica_count: 4
     log_level: info
     controller_cpu_request: "1"
--- a/osdc/docs/runner-image-autoresolve.md
+++ b/osdc/docs/runner-image-autoresolve.md
@@ -1,0 +1,147 @@
+# Runner Image Auto-Resolve
+
+## What
+
+Each `arc-runners` deploy pins the runner image
+(`ghcr.io/actions/actions-runner:<tag>@sha256:<...>`) **per OSDC commit**.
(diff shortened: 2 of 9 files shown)


zxiiro (approved):

zxiiro (approved):

huydhn:
Sorry for the late review, but this stack has one change that I'm still trying to justify. Per my understanding, `runner_image_tag` isn't explicitly spelled out anymore, but will be dynamically resolved at deployment time and mapped to the OSDC commit. Two thoughts I have on this:
* Not being able to version track the `runner_image_tag` anymore seems wrong. `history.json` should be committed to git
* Rolling back looks complex with that sha to tag mapping. Does this mean that we might potentially need to revert a change because it unknowingly pulls in a new runner image that causes a regression? What's about redeploying the same latest commit, but with a newer / older runner image? I guess this depends on how good the integration test suite does in catching issue with runner image

PR #773, opened 2026-06-16
Target integration test jobs to cluster-specific runner groups

jeanschmidt:
* #774
* __->__ #773

**Impact:** Integration tests only — no infrastructure or runner changes
**Risk:** low

## What
Updates all integration test workflow jobs to use `runs-on: { group: "<runner_group>", labels: ["<runner_label>"] }` instead of bare `runs-on: <runner_label>`, ensuring jobs are dispatched to the correct cluster's runner group. Release runner jobs use a separate group variable so clusters with a dedicated runner group don't accidentally target the org-wide `release-runners` group.

## Why
With multiple OSDC clusters registered under the same GitHub org, a bare `runs-on: <label>` can match runners from any cluster — integration tests meant for one cluster could land on another. Using the `group` constraint scopes job placement to the intended cluster. For clusters that define `arc-runners.runner_group` in `clusters.yaml` (e.g. `meta-staging-aws-uw1`, `arc-cbr-prod-uw1`), the release runner group must also be overridden to match — otherwise release test jobs would target a `release-runners` group that doesn't exist in that cluster's org-scoped runner group.

## Notes
- Clusters without an explicit `runner_group` default to `"default"` for regular jobs and `"release-runners"` for release jobs, preserving existing behavior for the us-east-2 production cluster.
- The `build-image.yaml` reusable workflow now accepts a `runner_group` input so callers can propagate the group constraint to BuildKit jobs.

### DIFF
--- a/osdc/integration-tests/workflows/build-image.yaml
+++ b/osdc/integration-tests/workflows/build-image.yaml
@@ -16,12 +16,19 @@ on:
         required: false
         type: string
         default: "l-x86iavx512-2-4"
+      runner_group:
+        description: "Runner group to target (cluster-scoped)"
+        required: false
+        type: string
+        default: "default"
 
 jobs:
   build:
     # NOTE: Always runs on x86 — BuildKit is a *remote* builder.
     # The arch input selects which BuildKit endpoint to connect to.
-    runs-on: ${{ inputs.runner_label }}
+    runs-on:
+      group: ${{ inputs.runner_group }}
+      labels: [ "${{ inputs.runner_label }}" ]
     container:
       image: ghcr.io/actions/actions-runner:latest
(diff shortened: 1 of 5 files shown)


zxiiro (approved):

osdc/integration-tests/workflows/build-image.yaml
```diff
@@ -16,12 +16,19 @@ on:
         required: false
         type: string
         default: "l-x86iavx512-2-4"
+      runner_group:
+        description: "Runner group to target (cluster-scoped)"
+        required: false
+        type: string
+        default: "default"
```
huydhn, line 23: Should this be `release-runners` as the default group here?

PR #774, opened 2026-06-16
Filter integration test jobs by cluster module availability

jeanschmidt:
* __->__ #774
* #773

**Impact:** Integration test orchestrator and CI workflows for all OSDC clusters
**Risk:** low

## What
Integration test workflow generation now uses the cluster's `modules` list from `clusters.yaml` to decide which test job sections to include, instead of separate boolean flags for each capability. Clusters that lack the required modules for all test suites get a no-op workflow so the GitHub Actions run stays valid.

## Why
Previously, only B200, cache-enforcer, and release blocks were conditionally toggled via individual booleans. Test suites for ARC runners, PyPI cache, GPU T4, and BuildKit always ran regardless of whether the cluster actually had those modules deployed. This caused integration test failures on clusters with partial module sets (e.g. new clusters still onboarding). Deriving the keep/strip decision from `clusters.yaml` modules makes it automatic and consistent -- adding a module to a cluster is all that's needed for its tests to start running.

## Notes
- New conditional tags added to the workflow template: `ARC_RUNNERS`, `PYPI_CACHE`, `GPU_T4`, `BUILDKIT` (join existing `B200`, `CACHE_ENFORCER`, `RELEASE`).
- Tag-to-module mapping is defined once in `TAG_REQUIREMENTS` in `phases.py`.
- If all tags are stripped (no modules match), the workflow emits a single `no-op` job instead of an empty `jobs:` block, which GitHub Actions would reject.
- The `.coverage` binary diff is a test artifact and can be ignored.


huydhn:
You probably want to do a rebase now as the bottom PR has been merge.  Also is `osdc/integration-tests/scripts/python/.coverage` accidentally added?

jeanschmidt:
> You probably want to do a rebase now as the bottom PR has been merge. Also is `osdc/integration-tests/scripts/python/.coverage` accidentally added?

removed the accidental .coverage file
rebased on main

osdc/integration-tests/scripts/python/phases.py
```diff
@@ -22,6 +22,16 @@
 
 SCRATCH_DIR_NAME = ".scratch"
 
+TAG_REQUIREMENTS: dict[str, list[str]] = {
+    "ARC_RUNNERS": ["arc-runners"],
+    "PYPI_CACHE": ["arc-runners", "pypi-cache"],
+    "GPU_T4": ["arc-runners", "nodepools"],
+    "BUILDKIT": ["arc-runners", "buildkit"],
+    "B200": ["arc-runners-b200", "nodepools-b200"],
```
huydhn, line 30: Maybe adding one for `h100` too or maybe remove this line? I'm not sure when we are going to have H100/B200 on staging :)

huydhn (approved):

osdc/integration-tests/scripts/python/phases.py
jeanschmidt: this is used for prod as well

PR #787, opened 2026-06-17
Add drain/undrain CI workflows for OSDC clusters

jeanschmidt:
**Impact:** OSDC operators performing cluster maintenance
**Risk:** low

## What
Adds two `workflow_dispatch` workflows (`osdc-drain-prod`, `osdc-undrain-prod`) that let operators drain and restore OSDC clusters from the GitHub Actions UI, and makes node tainting optional in the `drain-runners` just recipe via `OSDC_TAINT_NODES`.

## Why
Draining a cluster before maintenance currently requires CLI access and running just recipes manually. These workflows expose the same operations (scale runners to zero, taint/untaint nodes, resume runners) behind the Actions UI with environment protection gates and proper AWS credential assumption, so any operator can safely drain or undrain a cluster without local tooling.

The `OSDC_TAINT_NODES` env var was added to `drain-runners` so the drain workflow can run the drain step and the taint step independently — the workflow calls `drain-runners` with `OSDC_TAINT_NODES=no` and handles tainting as a separate step with a 180s settle sleep in between.

## Notes
- Both workflows share the `osdc-tofu-<cluster>` concurrency group with deploy/plan to prevent concurrent mutations on the same cluster.
- Drain and taint are independently toggleable via workflow input checkboxes (default: both enabled).
- Drain timeout is 90 minutes; undrain timeout is 60 minutes.

### DIFF
--- a/.github/workflows/osdc-drain-prod.yml
+++ b/.github/workflows/osdc-drain-prod.yml
@@ -1,0 +1,101 @@
+name: "OSDC: Drain cluster"
+
+# Manual cordon for an OSDC cluster before maintenance. Two independent
+# checkboxes: drain-runners scales every AutoscalingRunnerSet to maxRunners=0
+# and waits for in-flight runner pods to finish; taint-nodes applies
+# deploy.osdc.io/refresh-pending:NoSchedule to runner nodes so the scheduler
+# stops placing new pods on them. Default = both enabled (typical pre-maintenance
+# flow). Per-cluster concurrency group serializes against the deploy and plan
+# workflows on the same cluster. Reverse with `osdc-undrain-prod.yml`.
+on:
+  workflow_dispatch:
+    inputs:
+      cluster:
+        description: "Cluster to drain"
+        required: true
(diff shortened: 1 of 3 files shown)


georgehong:
What are typical cases where we would want to use only the drain or taint operation by itself?

georgehong (approved):

.github/workflows/osdc-drain-prod.yml
```diff
@@ -0,0 +1,101 @@
...
+  workflow_dispatch:
+    inputs:
+      cluster:
+        description: "Cluster to drain"
+        required: true
+        type: choice
+        default: "-- select cluster --"
+        options:
+          - "-- select cluster --"
+          - meta-staging-aws-uw1
```
huydhn, line 20: Nit: It looks like this works with every cluster, not only production one, so maybe the workflow name is redundant

jeanschmidt:
FFixes on https://github.com/pytorch/ci-infra/pull/788

PR #790, opened 2026-06-17
Add ECR image pull test to integration tests

jeanschmidt:
**Impact:** CI integration tests only — no cluster or production changes

**Risk:** low

## What
Adds a new `test-ecr-pull` job to the integration test workflow that verifies runner pods can pull PyTorch CI images directly from the upstream ECR registry (account `308535385114`). The image tag is resolved at orchestration time using the git tree-SHA of `pytorch/pytorch:.ci/docker`, matching how pytorch's own `docker-builds.yml` resolves tags.

## Why
OSDC runners need cross-account ECR pull access to run PyTorch CI jobs. Today there's no integration test coverage for this — a misconfigured IAM role or network policy would only surface as job failures in production. This test catches ECR pull regressions early, before they affect real workloads.

## Notes
- The test is gated on the `arc-runners` module — clusters without it skip both the GitHub API call and the workflow job.
- Tag resolution happens before pool drain so a GitHub API failure exits fast without disrupting the cluster.
- A `--ecr-pull-image-name` CLI flag allows overriding the default image name (`pytorch-linux-jammy-py3.10-clang18`) for debugging.
- A leftover-placeholder guard was added to `generate_workflow` — any `{{...}}` surviving substitution now raises, catching template/code drift.

example run: https://github.com/pytorch/pytorch-canary/actions/runs/27722726877/job/82011860076


huydhn:
lol, you need to update `.gitignore` to include that `.coverage` once and for all

huydhn:
@claude Review the new osdc/integration-tests/scripts/python/resolve_pytorch_image.py script
* Is `resolve_ci_docker_hash` equivalent to how PyTorch CI computes the Docker image hash https://github.com/pytorch/pytorch/blob/main/.github/workflows/docker-builds.yml#L133 using git rev-parse HEAD:.ci/docker
* Is there a way to get the latest available 308535385114.dkr.ecr.us-east-1.amazonaws.com/pytorch/ci-image:{ecr_image_name} instead of computing the hash? For example, latest available 308535385114.dkr.ecr.us-east-1.amazonaws.com/pytorch/ci-image:pytorch-linux-jammy-py3.10-clang18-ANY_HASH_IS_OK.  If yes, how complex is that approach when comparing with the current implementation using `resolve_ci_docker_hash`

jeanschmidt:
> lol, you need to update `.gitignore` to include that `.coverage` once and for all

hahahaha I always miss this one, I'll do a follow up PR on this

huydhn (approved):
LGTM!

huydhn:
@claude Is gh cli a dependency covered by mise already?

jeanschmidt:
> @claude Is gh cli a dependency covered by mise already?

should we? I guess we should, but curious on your take

huydhn:
Nit: It might be useful to add gh to mise.toml in another PR per Claude's comment above

jeanschmidt:
@huydhn FYI: https://github.com/pytorch/ci-infra/pull/792

.coverage is already on .gitignore

PR #794, opened 2026-06-18
Remove pypi-cache from LF Clusters

zxiiro:


zxiiro:
After talking to @jeanschmidt on Slack. We thought it'd be good to do the removal of pypi-cache testing on the LF clusters which will help with cost reduction on the LF side. If things fail with pypi-cache being disabled we can use the runner determinator to switch off the LF experiment.

huydhn (approved):

jeanschmidt:
There is a fix I need to push before merging this one, wait a bit

jeanschmidt:
fix here: https://github.com/pytorch/ci-infra/pull/795

PR #803, opened 2026-06-22
bin-pack-scheduler: MostAllocated secondary scheduler module

georgehong:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #807
* #804
* __->__ #803

EKS's managed kube-scheduler scores with LeastAllocated (spread) and can't be
reconfigured. Add a standalone module that runs a stock kube-scheduler as a
secondary scheduler with NodeResourcesFit -> MostAllocated, so pods that opt in
via spec.schedulerName get bin-packed onto the fullest node. 

Adds the following:
- config.yaml for scheduler settings
- deployment to make the scheduler available <--> cross-referenced with configs/examples used in https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/ 
- associated rbacs and kustomizations.

```
just test
```

PR #804, opened 2026-06-22
arc-runners: let runner defs select a workflow scheduler

georgehong:
* #807
* __->__ #804
* #803

Add a per-def scheduler_name knob. generate_runners.py stamps it onto both the
real workflow pod (schedulerName) and the listener's
CAPACITY_AWARE_WORKFLOW_SCHEDULER_NAME env, which the ARC fork applies to the
ph-w-* workflow placeholder. Keeping both in sync ensures the placeholder
reserves a slot the real pod can actually claim. Empty = default scheduler.

Opt two CPU defs (l-x86iavx512-8-64, l-x86iavx2-8-32) into bin-pack-scheduler
as a pilot. Unit tests cover both the set and unset paths.

### DIFF
--- a/osdc/clusters.yaml
+++ b/osdc/clusters.yaml
@@ -246,7 +246,7 @@ clusters:
       - karpenter
       - arc
       - nodepools
-      - bin-pack-scheduler    # MostAllocated secondary scheduler; inert until a def opts in
+      - bin-pack-scheduler
       - arc-runners
       - keda
       - buildkit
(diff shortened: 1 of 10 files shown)


jeanschmidt (approved):

jeanschmidt:
@claude can you explain this changes? How is the custom scheduler toggled on/off for particular clusters?

jeanschmidt:
this change only allows you to deploy all at once, or none.

Consider making it conditional on the existence of the custom scheduler module on cluster definitions

PR #811, opened 2026-06-23
osdc/hf-cache: terraform — per-cluster S3 bucket + read-only IRSA role

huydhn:
* #833
* #825
* #818
* #817
* #816
* #813
* #812
* __->__ #811

First commit of the hf-cache stack. AWS infrastructure only, in the module's own
per-cluster terraform root so `just deploy-module <cluster> hf-cache` provisions
it (no separate bucket step):

  * a private S3 bucket per cluster, pytorch-hf-model-cache-<cluster_id>, in the
    cluster's region. SSE; no public access; force_destroy=false so module
    removal won't wipe a populated cache.
  * a read-only IRSA role for the hf-cache-mount service account (the rclone
    mount reads the cache). Writes are NOT done here — they go through a
    GitHub-OIDC role in pytorch-gha-infra, so untrusted job pods can't write.

No Kubernetes resources and no runner changes yet — see the commits above.


osdc/modules/hf-cache/terraform/main.tf
```diff
@@ -0,0 +1,146 @@
...
+    }
+  }
+}
+
+# --- IRSA role for the rclone mount (read-only) ---
+# Runners only read the cache. Writes go through a GitHub-OIDC role in
+# pytorch-gha-infra (gha_workflow_hf-cache-write), assumed by ci-refresh-hf-cache
+# runs — so untrusted job pods can't write.
+
+resource "aws_iam_role" "hf_cache" {
```
jeanschmidt, line 99: do we want to tie yet another thing to AWS exclusively? Maybe we could think in a way to make this cloud agnostic? Is there a specific reason for a S3 bucket over a PCV on K8s?

huydhn: Good point, I was drilling about this exact point that week. The design depends on using https://github.com/rclone/rclone as a cloud-agnostic backend, it works in the same way in many other cloud providers.  Basically, only we can add more cloud storage in the future if needed and the stack would work in the same way.

Per my understanding, using PVC still requires a storage class / CSI driver behind it.  And a share cache needs something like EBS, EFS, or FSx anyway.  Another data point pushing me to use rclone is that while the remote storage is S3, the cache itself is node local and good for read-heavy workload with a warm cache on the node itself.  It also can't run out of disk space because the cache is capped at [75%](https://github.com/pytorch/ci-infra/pull/813/changes#diff-82b61cda543427ce26ac5658181b0f3a5173fef119b5e44974333d81e4486983R119).  I'm pretty happy with how rclone provides all these features out of the box.

The final piece is the ability to update the cache automatically.  Using OIDC to upload directly to S3 is really easy in CI, so I don't need to setup anything on infra to handle that (no schedule update jobs, no. need to maintain list of models to)

huydhn: Another data point I get from Claude after asking it to go over your comment:

> Worth adding: a PVC/CSI path would actually be less portable — the non-AWS targets we'd care about (e.g. DGX Cloud / Run:ai) expose only FSx + Mountpoint-S3 and no custom CSI, so a generic K8s PVC wouldn't be available there, whereas object storage is universal. And the only AWS-specific surface here is this terraform (bucket + IRSA); 
The mount, runner wiring, and offline HF_HOME contract are backend-agnostic, so a future cloud is a swap of this file + the rclone remote, not a redesign.

jeanschmidt (approved):

PR #817, opened 2026-06-24
osdc/hf-cache: integration tests — mount, strict offline read, large-model read

huydhn:
* #833
* #825
* #818
* __->__ #817
* #816
* #813
* #812
* #811

BEGIN_HF_CACHE job block (gated on arc-runners + hf-cache):

  * test-hf-cache-mount — /mnt/hf_cache mounted; HF_HOME/HF_HUB_CACHE/
    HF_CACHE_S3_BUCKET/HF_CACHE_S3_REGION correct; a write is rejected (read-only).
  * test-hf-cache-offline-read — mirrors normal CI jobs: offline-loads a small
    model (prajjwal1/bert-tiny) from /mnt/hf_cache via huggingface_hub. FAILS if
    the model isn't cached or won't load.
  * test-hf-cache-large-read — offline-reads a large model (Qwen/Qwen2.5-7B-
    Instruct) from the cache and times the weight read (cache benefit: no
    multi-GB Hub download). FAILS if not cached; seed once with
    `just hf-cache-seed <cluster>`.

Inert until a cluster enables the hf-cache module.


osdc/integration-tests/workflows/integration-test.yaml.tpl
```diff
@@ -239,6 +239,147 @@ jobs:
   # END_NO_CACHE_ENFORCER
   # END_ARC_RUNNERS
 
+  # BEGIN_HF_CACHE
```
jeanschmidt, line 242: check how this plays out with our h100/b200 cluster, if those tests will run there.

BC we don't have regular runners there, but if we're going to enable it there we can't use those h100 ones as this would be a problem with queuing for testing/releases

osdc/integration-tests/workflows/integration-test.yaml.tpl
```diff
@@ -239,6 +239,147 @@ jobs:
...
+  # ── HF Cache: read-only mount, env, and offline model load ──────────
```
jeanschmidt, line 243: can you also add a test to do a parallel/concurrent upload and make sure there are no conflicts?

osdc/integration-tests/workflows/integration-test.yaml.tpl
huydhn: Handled by two independent gates, so the test never queues for a runner that can't come online:

- **H100-only region (no regular `arc-runners`), e.g. us-west-1 prod:** the `HF_CACHE` block requires the `arc-runners` module (its jobs run on CPU/L4 runners). A cluster with only `arc-runners-h100` fails that gate, so the whole hf-cache suite strips to a no-op — nothing is scheduled, nothing waits. The mount itself still works on the H100 nodes if hf-cache is enabled there; only the *test* is skipped.
- **Region without L4, e.g. us-west-1 (`g6` has `exclude_regions: [us-west-1]`):** the large-read job runs on an L4 (g6). It's now wrapped in a `HF_CACHE_GPU` block that's region-gated against the fleet's `exclude_regions`, so only that one job is stripped there while the CPU jobs (mount/offline/refresh) still run. Where L4 exists (us-east-1/2) it runs unchanged.

osdc/integration-tests/workflows/integration-test.yaml.tpl
huydhn: Added as `test-hf-cache-concurrent-sync` in #818 (the refresh/write PR, where this OIDC `aws s3 sync` path lives). It runs **N=3 parallel `aws s3 sync`** of the same model to a per-run prefix and asserts:

- all writers exit 0 (no error under contention),
- the set is **complete** — a reconciling `--dryrun` sync finds nothing left to upload (nothing lost to a race), and
- the upload is **uncorrupted & usable** — it pulls the copy back from S3, loads it with `transformers` + a forward pass, and asserts the output matches the freshly-downloaded source **bit-for-bit**.

huydhn:
Those tests are updated, plz take another look @jeanschmidt

jeanschmidt (approved):

PR #832, opened 2026-06-25
[RELAND] Tune node compaction speed and fix transient taint false-negatives

jeanschmidt:
**Impact:** OSDC runner clusters — node compaction and Karpenter provisioning latency
**Risk:** medium

RELAND of https://github.com/pytorch/ci-infra/pull/827
REVERT and explanation of the issue found on https://github.com/pytorch/ci-infra/pull/831

## What
Makes the node compactor react faster to utilization changes and fixes a bug where transient/startup taints caused the compactor to incorrectly reject pod-to-node matches.

## Why
The compactor was too slow to reclaim underutilized nodes and too slow to release them back when pending pods appeared. Three timing knobs were conservative from initial rollout and are now tightened based on production observations:

- **Compactor loop interval**: 20s → 7s — the compactor polls node state on this cadence; 20s left pods waiting unnecessarily when demand spiked.
- **Fleet cooldown**: 900s → 120s — the delay between tainting rounds within the same fleet. 15 minutes was far too conservative once the compactor proved stable; 2 minutes still prevents cascading disruption while allowing faster scale-down.
- **Karpenter batch window**: added `BATCH_MAX_DURATION` and `BATCH_IDLE_DURATION` at 45s (up from Karpenter's 10s default) — gives Karpenter more time to accumulate pending pods into a single provisioning decision, reducing unnecessary node churn from back-to-back launches.

Separately, the pod-matching logic was producing false negatives: when a node carried OSDC startup taints (`node-init.osdc.io/*`), Karpenter's `unregistered` taint, or Kubernetes transient taints (`node.kubernetes.io/unreachable`, etc.), the compactor rejected pods that would actually schedule there once those taints cleared. The compactor now skips these known-transient taints during matching, and also skips `PreferNoSchedule` taints (which never block scheduling). `karpenter.sh/disrupted` is intentionally **not** ignored — pods shouldn't land on nodes committed to termination.

# Notes
- The toleration matching logic was also restructured for clarity (early-continue on key mismatch, `None` value treated as wildcard on the toleration side to match upstream kube-scheduler behavior). No behavioral change beyond the transient-taint filtering.
- New unit tests cover every ignored taint key, the `karpenter.sh/disrupted` reject case, `PreferNoSchedule` passthrough, and an integration-style test through `check_pending_pods` with stacked ignored taints.


huydhn (approved):

jeanschmidt:
@claude can you verify if the issues that occurred on the original version and identified in the revert and on the smoke https://github.com/pytorch/ci-infra/actions/runs/28139399748/job/83336464199 have been correctly addressed here? 

Can you validate that now the behavior is as intended? And can you validate that the flag will be correctly applied?

jeanschmidt:
just deployed to uw1 and verified that this change now generates the same intended config and should not reach the 3-way merge bug introduced.

PR #834, opened 2026-06-25
osdc/hf-cache: fix mount DaemonSet CrashLoop (stale FUSE mount on restart)

huydhn:
* __->__ #834
* #833
* #825
* #818
* #817
* #816
* #813
* #812
* #811

The rclone mount pods were stuck in CrashLoopBackOff with "failed to mount FUSE
fs: directory already mounted" — on a crash/kill the preStop unmount doesn't run,
so the dead process's FUSE mount stays on the host and rclone refuses to mount
over it, never recovering. So /mnt/hf_cache was never mounted and every read job
failed ("model not in the cache").

Fix: clear any stale mount at container startup (fusermount -uz / umount -l)
before mounting, and add --allow-non-empty as a fallback, so restarts recover.

### DIFF
--- a/osdc/modules/hf-cache/kubernetes/mount-daemonset.yaml.tpl
+++ b/osdc/modules/hf-cache/kubernetes/mount-daemonset.yaml.tpl
@@ -67,6 +67,10 @@ spec:
               set -eu
               MOUNT=/mnt/hf_cache
               CACHE=/mnt/hf-cache-vfs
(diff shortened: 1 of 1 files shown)


huydhn:
Folded into #812 (the mount DaemonSet PR) — the CrashLoop fix belongs with the DaemonSet definition. No longer a standalone change.

PR #839, opened 2026-06-25
Node compactor: pending-pod awareness and peak-window taint dampening

jeanschmidt:
**Impact:** Node compactor behavior on all OSDC clusters — tainting becomes less aggressive
**Risk:** medium

## What
The node compactor now includes pending (unschedulable) pods in its bin-pack calculation and tracks a 45-minute sliding window of peak demand per fleet, preventing taints from firing on transient load dips. Burst untainting now uses bin-packing instead of summed totals, so workloads with large pods get the right number of nodes released.

## Why
The compactor was tainting too aggressively in two ways: it ignored pending pods waiting for capacity (tainting nodes those pods could have used), and it reacted instantly to momentary drops in load, oscillating between taint/untaint cycles. The peak-window floor smooths out short dips so nodes only get reclaimed after demand has stayed low for 45 minutes. Including pending pods in the bin-pack input means the compactor keeps enough nodes untainted to absorb queued workloads instead of starving them.

Burst untainting separately had a sum-vs-sum capacity check that under-released nodes when individual pods were large relative to per-node free capacity (three 20-vCPU pods vs three 30-vCPU nodes: sums say two suffice, packing shows all three are needed). It now uses the same First-Fit-Decreasing bin-pack as the rest of the compactor.

# Notes
- \`_pod_matches_node\` was split into \`_pod_constraints_match\` (tolerations/selector/affinity only) and \`_pod_fits_resources\` (resource capacity). The pending-pod filter uses \`_pod_constraints_match\` with the compactor taint stripped to check fleet compatibility without requiring current-usage fit.
- Pending pods are bounded to [0s, 4h] age — every pending pod counts the moment it appears; the upper bound stops pods stuck on non-capacity issues from inflating the floor forever. The remaining \`age < 0\` check defends against clock skew.
- The pending-pod resource-fit pre-filter now subtracts DaemonSet overhead from allocatable, matching the rest of the bin-pack accounting.
- \`check_pending_pods\` now returns \`(nodes_to_untaint, compatible_count)\` so the \`pending_pods_compatible\` metric reflects actual compatible count rather than total pending count.
- \`_pods_fit_on_nodes\` extracted to \`fit.py\` and \`pod_to_podinfo\` added to \`models.py\` to eliminate duplicated conversions and break a potential import cycle.
- Peak history lives in process memory and resets on restart, which is safe since the next reconcile rebuilds from observed load.


jeanschmidt:
will submit a stacked PR with ghstack to make reviewing easier

PR #854, opened 2026-06-29
Add multi-cluster HUD queue sharding for capacity-aware listeners

jeanschmidt:
**Impact:** All ARC runner listeners across prod and staging clusters
**Risk:** low

see: https://github.com/jeanschmidt/actions-runner-controller/pull/10

## What
Automatically derive per-cluster sharding parameters (`CLUSTER_INDEX`, `CLUSTER_COUNT`, `AGE_THRESHOLD_SECONDS`) from `clusters.yaml` so each listener claims only its fair share of the HUD queued-jobs feed, with age-based failover when a peer cluster is unhealthy.

## Why
Multiple clusters (e.g. the three `mt-` prod regions) advertise the same runner labels and all poll the same HUD queue. Without coordination each cluster reacts to the full queue independently, causing redundant scale-ups. This change partitions fresh jobs across peer clusters deterministically — peers are clusters that deploy the same `arc-runners`-family module and share a `runner_name_prefix` — while jobs older than the age threshold are claimed by all peers as a failover signal. No cross-cluster communication is needed; everything is derived at generate-time from the existing cluster definitions.

## Notes
- Sharding values are injected as env vars on the listener pod; the actual slicing logic lives in the ARC fork's capacity-aware controller (not in this repo).
- Default age threshold is 900s (15 min). Clusters can override via `arc-runners.capacity_aware_age_threshold_seconds` in `clusters.yaml`.
- Single-cluster or mismatched-config scenarios degrade safely to `(0, 1)` (no sharding).


osdc/modules/arc-runners-b200/defs/l-bx86iamx-176-1800-b200-8.yaml
```diff
@@ -12,5 +12,6 @@ runner:
   vcpu: 176
   memory: 1800Gi
   gpu: 8
+  fresh_multiplier: 1.0
```
huydhn: Do we need to add this explicitly here given that the default is 1 anyway? I think it's clearer for these pet instances as they are not scalable

osdc/modules/arc-runners/deploy.sh
```diff
@@ -30,7 +30,7 @@ source "$UPSTREAM_ROOT/scripts/kubectl-apply.sh"
 CFG="$UPSTREAM_ROOT/scripts/cluster-config.py"
 
 # Chart version — MUST match the controller (arc module). Read from clusters.yaml.
-ARC_CHART_VERSION=$(uv run "$CFG" "$CLUSTER" arc.chart_version 0.14.1)
+ARC_CHART_VERSION=$(uv run "$CFG" "$CLUSTER" arc.chart_version 0.14.1-jeanschmidt.16)
```
huydhn, line 33: Is this a bug? Same goes for the change on `osdc/modules/arc/deploy.sh`?

osdc/modules/arc-runners/scripts/python/generate_runners.py
```diff
@@ -219,6 +243,23 @@ def generate_runner(
     if proactive_cap is not None:
         proactive_capacity = min(proactive_capacity, proactive_cap)
 
+    fresh_multiplier = runner.get("fresh_multiplier")
+    if fresh_multiplier is None:
+        fresh_multiplier = cluster_config.get("capacity_aware_fresh_multiplier", 1.0)
```
huydhn: My read is that `capacity_aware_fresh_multiplier` is set for `arc-runners`. Is this a correct understanding that this would not affect `arc-runners-h100` and `arc-runners-b200`?

osdc/modules/arc-runners/scripts/python/generate_runners.py
```diff
@@ -219,6 +243,23 @@ def generate_runner(
...
+    try:
+        fresh_multiplier = float(fresh_multiplier)
+    except (ValueError, TypeError):
+        log_error(f"Invalid fresh_multiplier for runner {runner_name}: {fresh_multiplier!r} (must be a number)")
+        sys.exit(1)
+    aged_multiplier = runner.get("aged_multiplier")
```
huydhn: I'm not sure if we need this `aged_multiplier` to be honest, maybe we do, but the change would look simpler without it.  This future knob is not set anywhere yet.

osdc/modules/arc-runners-b200/defs/l-bx86iamx-176-1800-b200-8.yaml
jeanschmidt: yes, bc we set the default cluster-wide as 0.8. The order is: no-value-default (1.0) < cluster.yaml cluster default < per-runner override.

osdc/modules/arc-runners/deploy.sh
jeanschmidt: why is this a bug? do you want to keep the default to 0.14.1? Like I have nothing against that, but not sure if I care in this case. Basically we override everywhere anyways...

osdc/modules/arc-runners/scripts/python/generate_runners.py
jeanschmidt: ohh I see what you mean, set up at the runners definition level for the module that installs it, instead of overriding only at the runner level for those h100/b200 instances..

good point, would make the usage simpler

osdc/modules/arc-runners/scripts/python/generate_runners.py
jeanschmidt: yeah, I fought claude to add this... do you think we should remove it for now?

jeanschmidt:
@huydhn after a restful night of sleep, I am changing the following agreeing with you:

1) strip the aged_multiplier from OSDC
2) change the multiplier lookup in generate_runners.py to read from cluster_cfg[module_name] so we can setup multiplier at module name level and avoid the confusion on where values should be set (very odd behavior and naming).
3) use 2 to remove the configuration from the h100/b200 scalesets configs

still I'll keep the per-runner scape hatch in case we find the need a particular runner that is hard to get and is queuing (I am thinking a10g and potentially others)

huydhn (approved):
LGTM!

osdc/modules/arc-runners/deploy.sh
huydhn: I mean the old code looks like a bug `0.14.1` instead of your fork `0.14.1-jeanschmidt.16`, or is this `0.14.1-jeanschmidt.16` a test branch and you will replace before landing this?

PR #856, opened 2026-06-29
Node compactor: pending-pod awareness and peak-window taint dampening

jeanschmidt:
* #860
* #859
* #858
* #857
* __->__ #856

**Impact:** Node compactor behavior on all OSDC clusters — tainting becomes less aggressive
**Risk:** medium

## What
The node compactor now includes pending (unschedulable) pods in its bin-pack calculation and tracks a 45-minute sliding window of peak demand per fleet, preventing taints from firing on transient load dips.

## Why
The compactor was tainting too aggressively in two ways: it ignored pending pods waiting for capacity (tainting nodes those pods could have used), and it reacted instantly to momentary drops in load, oscillating between taint/untaint cycles. The peak-window floor smooths out short dips so nodes only get reclaimed after demand has stayed low for 45 minutes. Including pending pods in the bin-pack input means the compactor keeps enough nodes untainted to absorb queued workloads instead of starving them.

# Notes
- \`_pod_matches_node\` was split into \`_pod_constraints_match\` (tolerations/selector/affinity only) and \`_pod_fits_resources\` (resource capacity). The pending-pod filter uses \`_pod_constraints_match\` with the compactor taint stripped to check fleet compatibility without requiring current-usage fit.
- Pending pods are bounded to [30s, 4h] age — too-young pods are still scheduling normally, too-old pods are likely stuck on non-capacity issues.
- \`check_pending_pods\` now returns \`(nodes_to_untaint, compatible_count)\` so the \`pending_pods_compatible\` metric reflects actual compatible count rather than total pending count.
- Peak history lives in process memory and resets on restart, which is safe since the next reconcile rebuilds from observed load.


huydhn (approved):
LGTM! Asking my agent to find alternatives for peak_window and pending and verify how the rest of the change integrate with the stack, and the results look good. There are some nits but they are not worth mentioning IMO

zxiiro (approved):

PR #857, opened 2026-06-29
Node compactor: use bin-packing for pending-pod untainting

jeanschmidt:
* #860
* #859
* #858
* __->__ #857
* #856

**Impact:** Node compactor behavior on all OSDC clusters — pending-pod untainting now respects per-node capacity
**Risk:** medium

## What
Replace the sum-based resource-matching used by `check_pending_pods` with the same First-Fit-Decreasing bin-pack the compactor uses elsewhere, and extract the packing helper into its own module to break a circular import.

## Why
The previous pending-pod path compared summed demand against summed availability across all compatible tainted nodes. For workloads where individual pods are large relative to per-node capacity, that arithmetic under-counts the required nodes (three 20-vCPU pods vs two 30-vCPU nodes: sums say two suffice but bin-packing needs three). Running the real FFD packing yields the correct minimum node count and avoids starving large pods.

# Notes
- `_pods_fit_on_nodes` moves out of `packing.py` into a new `fit.py` so both `packing` and `taints` can import it without cycling.
- DaemonSet overhead is now subtracted from per-node allocatable in the pending-pod fit filter, so pods requesting exactly node allocatable are correctly rejected.
- `pod_to_podinfo()` is consolidated in `models.py` and reused from `pending.py` and `taints.py`, dropping duplicated conversion code.
- `PENDING_POD_MIN_AGE_SECONDS` drops to 0 so freshly-created pods are considered immediately; only future-timestamped pods are excluded.


huydhn:
_(Review note from Claude, an AI assistant, on behalf of @huydhn.)_

Confirming an interaction with #856 worth a sanity check before this lands: `PENDING_POD_MIN_AGE_SECONDS = 0` also affects the bin-pack/peak-window path, not just burst detection. `pending_pods_for_group` gates on this same constant, so with it at 0 a pending pod feeds `bin_pack_min_nodes` → `update_peak_history` the moment it appears — and the 45-min peak-window floor from #856 then pins `min_needed` elevated for the full window even after the pod schedules or is deleted.

The net effect: normal sub-30s scheduling latency (previously filtered out by the `30` threshold) now writes into the 45-min floor, so transient pending blips hold capacity for 45 min. That's the opposite direction from #856's dampening intent. Is that intended? If the min-age guard was only meant to relax the *burst-untaint* path, it might be worth keeping a non-zero floor on the bin-pack/peak path specifically (or documenting why immediate counting is desired there).

huydhn (approved):

jeanschmidt:
> The net effect: normal sub-30s scheduling latency (previously filtered out by the `30` threshold) now writes into the 45-min floor, so transient pending blips hold capacity for 45 min. That's the opposite direction from #856's dampening intent. Is that intended? If the min-age guard was only meant to relax the _burst-untaint_ path, it might be worth keeping a non-zero floor on the bin-pack/peak path specifically (or documenting why immediate counting is desired there).

yes the 30s threshold was a bug IMO, and setting to 0 makes sense. The problem is that compactor would not count them and not untaint nodes to acomodate capacity or use those pods for capacity calculation, being one of the sources for fluctuations

zxiiro (approved):

PR #858, opened 2026-06-29
Node compactor: make peak/pending timing knobs env-configurable

jeanschmidt:
* #860
* #859
* __->__ #858
* #857
* #856

**Impact:** OSDC operators tuning compactor behavior per cluster
**Risk:** low

## What
Move `PEAK_WINDOW_SECONDS`, `PENDING_POD_MAX_AGE_SECONDS`, and `PENDING_POD_MIN_AGE_SECONDS` from module-level constants into `Config`, sourced from `COMPACTOR_*` environment variables with validation, and thread them through the call sites that previously imported the constants.

## Why
Different clusters have different scaling characteristics — a fleet with slow Karpenter provisioning needs a different peak-window than one with warm capacity, and pending-pod age bounds depend on workload churn. Hard-coding these values meant any tuning required a code change and redeploy. Surfacing them as env vars lets operators iterate per-cluster from `clusters.yaml` overrides.

# Notes
- New e2e test exercises a 30-second peak window end-to-end to confirm damping engages and releases on the configured schedule.


osdc/base/node-compactor/scripts/python/fit.py
```diff
@@ -0,0 +1,50 @@
+"""Bin-pack fit check shared between packing and taint decisions.
```
huydhn, line 1: Nit: It looks like this file should be on https://github.com/pytorch/ci-infra/pull/857 instead, but it's ok as the whole stack will be landed together anyway

huydhn (approved):

osdc/base/node-compactor/scripts/python/models.py
```diff
@@ -92,8 +83,24 @@ def env(key: str) -> str:
             spare_capacity_ratio=float(env("COMPACTOR_SPARE_CAPACITY_RATIO")),
             spare_capacity_threshold=float(env("COMPACTOR_SPARE_CAPACITY_THRESHOLD")),
             capacity_reservation_nodes=int(env("COMPACTOR_CAPACITY_RESERVATION_NODES")),
+            peak_window_seconds=int(env("COMPACTOR_PEAK_WINDOW_SECONDS")),
+            pending_pod_max_age_seconds=int(env("COMPACTOR_PENDING_POD_MAX_AGE_SECONDS")),
```
huydhn, line 87: The agent suggests a validation check here to confirm that `pending_pod_min_age_seconds < pending_pod_max_age_seconds` here, which makes sense.

> If an operator sets PENDING_POD_MIN_AGE_SECONDS=20000 with MAX=14400 (min > max), the filter age < min or age > max excludes every pod — pending-pod awareness silently turns off with no error

jeanschmidt: added the change

zxiiro (approved):

PR #859, opened 2026-06-29
Node compactor e2e: pause ARC scalesets during tests

jeanschmidt:
* #860
* __->__ #859
* #858
* #857
* #856

**Impact:** Node compactor e2e test stability — no production runtime impact
**Risk:** low

## What
Add pause/restore helpers that set `maxRunners=0` on every `AutoscalingRunnerSet` CRD during compactor e2e setup and restore the original values on teardown.

## Why
ARC runner scaling creates cluster-wide node churn that races with the compactor — nodes appear and disappear unpredictably during a test run, causing flaky taint/untaint assertions. Pausing scalesets gives the test suite a stable fleet to reason against.

# Notes
- Original `maxRunners` values (including `None`) are preserved so restore is faithful, using JSON Patch add/replace/remove.
- Missing CRD or namespace returns an empty dict so the helpers are safe to run on clusters without ARC installed.


zxiiro (approved):

osdc/base/node-compactor/tests/e2e/conftest.py
huydhn, line 492: The agent suggests that to move `atexit.register(restore)` closer to line `arc_originals = pause_arc_scalesets(client)` above to avoid the possibility of leaving the cluster in a broken state with max_runner set to 0, which I agree with.  There are too many things running in between `pause_arc_scalesets` and `atexit.register(restore)`

huydhn (approved):
There is a nit point on `atexit.register(restore)`, otherwise LGTM!

osdc/base/node-compactor/tests/e2e/conftest.py
jeanschmidt: asked agents to find and fix other occurences of cleanup operations being performed after long/risky tasks, submitting a fix.

PR #860, opened 2026-06-29
Node compactor: add per-decision logging, switch e2e to log-based assertions

jeanschmidt:
* __->__ #860
* #859
* #858
* #857
* #856

**Impact:** Node compactor observability on all OSDC clusters; e2e test reliability
**Risk:** low

## What
Emit a structured `log.info` line for every taint/untaint decision the compactor evaluates, and rewrite the peak-window e2e assertions to grep these log lines instead of polling node state.

## Why
Production debugging previously required reconstructing the compactor's reasoning from before/after node state, which is lossy — there was no record of "considered, skipped because reserved" or "would have tainted, suppressed by peak-window". The new log lines emit fleet, candidate node, uptime, utilization, and outcome for every decision, so a single `kubectl logs` query gives the full reasoning chain. The e2e suite was racing with Karpenter on node state, so the same log lines now drive the peak-window assertions via the existing `_wait_for_compactor_reconcile` helper, replacing fragile `--tail` reads with `--since`-based polling.

# Notes
- `COMPACTOR_PEAK_WINDOW_SECONDS=0` now disables peak damping entirely.
- Peak-window e2e config is isolated to `TestPeakWindow` via a dedicated `GROUP_PEAK_WINDOW_CONFIG`; other groups run with `peak_window=0` to avoid cross-test interference.
- The B3 fleet-cooldown test was removed because it could not be made deterministic — its scenario is already covered by other tests.


osdc/base/node-compactor/tests/e2e/test_e2e.py
```diff
@@ -899,58 +846,44 @@ def test_peak_window_blocks_then_releases(self) -> None:
...
-                elif n not in pool_names and n in observed_tainted:
-                    deleted_after_tainted.add(n)
+                pattern = rf"Tainted node {re.escape(n)}\b"
+                if search_compactor_logs(self.logs, pattern, log_pos):
+                    tainted.add(n)
+            return tainted
 
-        # Phase 1 — within the 30s peak window: drained nodes MUST NOT be tainted.
-        # Poll for ~15s and assert taints never appear on the drained nodes.
         while time.monotonic() < drain_start + 15:
```
zxiiro, line 868: Should this be 30 seconds rather than 15 since the comment above says the peak window is 30s?

```suggestion
        while time.monotonic() < drain_start + 30:
```

zxiiro (approved):

huydhn (approved):

osdc/base/node-compactor/tests/e2e/test_e2e.py
jeanschmidt: the timeout seems correct actually

PR #864, opened 2026-07-02
osdc: enable hf-cache on the rest of the meta prod fleet (ue1, ue2)

huydhn:
Enable the shared HuggingFace model cache (`hf-cache`) on the remaining **meta** production clusters — `meta-prod-aws-ue1` and `meta-prod-aws-ue2` — completing the fleet after #863 (uw1) and the staging clusters. Inserted before `nodepools` so the rclone mount is up before nodepools emit the node-init gate taint.

Cache starts empty; seed via `scripts/hf-cache-seed.py` after deploy. `lf-prod-*` are a separate (LF) fleet and are intentionally left out.

### DIFF
--- a/osdc/clusters.yaml
+++ b/osdc/clusters.yaml
@@ -428,6 +428,7 @@ clusters:
     modules:
       - karpenter
       - arc
+      - hf-cache              # before nodepools: mount must be up before the node-init gate taint
       - nodepools
       - bin-pack-scheduler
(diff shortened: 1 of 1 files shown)


huydhn:
Split into one PR per cluster: #865 (meta-prod-aws-ue1) and #866 (meta-prod-aws-ue2).

PR #868, opened 2026-07-02
[Grafana][PyTorch DevX] Add pull/trunk/docs-build and pull TTS over-t…

georgehong:
…ime panels

Port HUD's single-stat p50 TTS metric (hud.pytorch.org/metrics) to an over-time view on the Core Metrics dashboard, adding two ClickHouse-backed time-series panels:

- p50 pull, trunk, docs-build TTS: daily max of the per-workflow p50 durations over successful run_attempt=1 runs, matching HUD's workflow_duration_percentile definition.
- p50 pull TTS: the pull workflow on its own.

Both scope to $repository and overlay a 7-day rolling average. Thresholds are calibrated to this half's targets (combined: amber 3h / red 4.5h; pull: amber 90m / red 2.75h).

Note: repo-scoped, so pytorch/pytorch reads slightly higher than HUD's headline, which pools pytorch/pytorch + executorch + tokenizers under the same workflow names.

Claude driven validation against HUD (reproduces ~3.2h) and by diffing the deployed rawSQL against an independent reimplementation day-by-day (identical).

**Additional sanity checks:**
- Roughly around the 4 hours for total TTS and pull shows the recent rebalancing fix from Monday. 
- View changes at https://pytorchci.grafana.net/goto/sbkhzj?orgId=stacks-1172652

### DIFF
--- a/grafana/pytorch_devx.json
+++ b/grafana/pytorch_devx.json
@@ -677,6 +677,284 @@
           "version": "13.1.0-25668120414"
         }
       }
+    },
+    "panel-8": {
+      "kind": "Panel",
+      "spec": {
+        "data": {
+          "kind": "QueryGroup",
+          "spec": {
+            "queries": [
+              {
+                "kind": "PanelQuery",
+                "spec": {
+                  "hidden": false,
+                  "query": {
+                    "datasource": {
+                      "name": "ceczcsck1b20wb"
+                    },
+                    "group": "grafana-clickhouse-datasource",
+                    "kind": "DataQuery",
+                    "spec": {
+                      "editorType": "sql",
+                      "format": 0,
+                      "meta": {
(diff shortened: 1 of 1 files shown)


jathu (approved):
I wonder if it would be easier if they were both just one graph? One line for pull, and another for pull+trunk+docs. Then the warning line could be 90 min and red line at 3 hours

georgehong:
> I wonder if it would be easier if they were both just one graph? One line for pull, and another for pull+trunk+docs. Then the warning line could be 90 min and red line at 3 hours

I can look into this as a follow-up. Currently, each of the target goals has its own dashboard (full TTS) versus Pull, and since we track both fine-grain and rolling averages, I'm wondering if it would be easier to read separately.

PR #874, opened 2026-07-06
osdc/hf-cache: raise rclone memory limit on large-GPU nodes

huydhn:
rclone was OOM-killed (1Gi) on H100 nodes pulling multi-GB models under `--vfs-cache-mode full`, dropping the mount node-wide and wedging every job pod on the node. Split the mount into two DaemonSets so only large-GPU nodes (`instance-gpu-name` affinity, default h100,b200 — configurable) get a raised 4Gi limit; standard nodes keep 1Gi.


georgehong (approved):
LGTM, would you have an example job affected by this? What does it mean for every job pod on the node to be wedged?

huydhn:
> LGTM, would you have an example job affected by this? What does it mean for every job pod on the node to be wedged?

Yes, like other critical start-up modules, it's important to not crash them because jobs can't start otherwise. So, the usual error is job can't start on H100 https://github.com/pytorch/pytorch/actions/runs/28798869394/job/85461534765.

Claude suggests there might be a way to self-heal this by remounting the volume, but I think I will do that in a separate PR

huydhn:
Deploy and test on ue1 staging https://github.com/pytorch/pytorch-canary/actions/runs/28836115484

PR #876, opened 2026-07-07
osdc/hf-cache: self-heal rclone mount across a dead FUSE

huydhn:
When rclone dies (e.g. OOM), the FUSE at `/mnt/hf_cache` goes stale ("transport endpoint is not connected") and containerd can't stat the volumeMount source, so the rclone container can't restart and every job pod on the node wedges in `CreateContainerError` until a manual unmount.

rclone now Bidirectional-mounts the stable parent `/mnt` instead of `/mnt/hf_cache`, so containerd always stats a live dir and the mount self-heals on restart. The init container also clears any dead FUSE on pod recreation.

**Updated 2026-07-17:**
- Rebased onto current main (was ~18 commits behind; now sits on top of the GPU tiering / GOMEMLIMIT / per-tier buffer changes and the smoke-test refactor).
- Switched all FUSE-clear calls from `fusermount` to **`fusermount3`**: the rclone image (rclone/rclone:1.69.1) and the AL2023 host both ship only the FUSE3 binary, so the previous `fusermount -uz` cleanups silently no-op'd (command not found) and never cleared a stale mount — which is exactly why mounts wedged. Verified in-cluster: `command -v fusermount` is empty, `/usr/bin/fusermount3` is present.

⚠️ Still changes mount propagation (now the parent `/mnt`) — needs a staging soak before prod.

### DIFF
--- a/osdc/modules/hf-cache/kubernetes/mount-daemonset.yaml.tpl
+++ b/osdc/modules/hf-cache/kubernetes/mount-daemonset.yaml.tpl
@@ -75,6 +75,21 @@ spec:
               set -eu
               exec /proc/1/root/usr/bin/nsenter -t 1 -m -- /bin/sh -c '
                 mkdir -p /mnt/hf_cache /mnt/hf-cache-vfs
+                # A crashed rclone leaves a dead FUSE at /mnt/hf_cache that fails to stat,
+                # wedging every pod on the node in CreateContainerError. rclone self-heals
+                # on in-place restart (it Bidirectional-mounts the parent /mnt, always a
+                # live dir), but a pod recreation only runs this init — so clear the dead
+                # endpoint here. fusermount3 is the FUSE3 binary on AL2023 hosts (fusermount
+                # is not installed); umount -l is the last-resort peel. Both leave the bind
+                # under /mnt intact.
+                while ! stat /mnt/hf_cache >/dev/null 2>&1; do
+                  fusermount3 -uz /mnt/hf_cache 2>/dev/null || umount -l /mnt/hf_cache 2>/dev/null || break
+                done
+                # rclone Bidirectional-mounts the parent /mnt, so /mnt must be a shared
+                # mountpoint: bind it first, then keep /mnt/hf_cache a shared submount so
+                # the FUSE still propagates to job pods.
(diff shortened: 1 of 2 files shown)


osdc/modules/hf-cache/kubernetes/mount-daemonset.yaml.tpl
```diff
@@ -75,6 +75,21 @@ spec:
...
                 mkdir -p /mnt/hf_cache /mnt/hf-cache-vfs
+                # A crashed rclone leaves a dead FUSE at /mnt/hf_cache that fails to stat,
+                # wedging every pod on the node in CreateContainerError. rclone self-heals
+                # on in-place restart (it Bidirectional-mounts the parent /mnt, always a
+                # live dir), but a pod recreation only runs this init — so clear the dead
+                # endpoint here. fusermount3 is the FUSE3 binary on AL2023 hosts (fusermount
+                # is not installed); umount -l is the last-resort peel. Both leave the bind
+                # under /mnt intact.
+                while ! stat /mnt/hf_cache >/dev/null 2>&1; do
+                  fusermount3 -uz /mnt/hf_cache 2>/dev/null || umount -l /mnt/hf_cache 2>/dev/null || break
```
jeanschmidt, line 86: yeahp, this is the bigger leaver here :)

osdc/modules/hf-cache/kubernetes/mount-daemonset.yaml.tpl
```diff
@@ -160,7 +176,7 @@ spec:
                 command:
                   - /bin/sh
                   - -c
-                  - "fusermount -uz /mnt/hf_cache || true"
+                  - "fusermount3 -uz /mnt/hf_cache || true"
```
jeanschmidt, line 184: why fusemount3?

jeanschmidt (approved):

jeanschmidt (approved):

osdc/modules/hf-cache/kubernetes/mount-daemonset.yaml.tpl
huydhn: That's the one AmazonLinux 2023 has. I only learn about this after jumping into a live node.  I think I can harden this part by checking if `fusermount` or `fusermount3` is available

PR #878, opened 2026-07-07
[CRCR] Increase reserved concurrent executions for callback Lambda function from 50 to 100

can-gaa-hou:
`50` concurrent callback Lambdas are not enough for handling job-level callbacks. Consider we have 10 L2+ repos, and each repo has 10 workflows and each workflow has only 1 callback job; that would need 100 Lambdas working at the same time. In practice, each workflow could have more than 1 callback job. `100` is the minimum in the current stage, and we can increase it further in the future.


subinz1 (approved):
lgtm

atalman (approved):

PR #883, opened 2026-07-08
osdc/hf-cache: tier rclone memory by GPU count

huydhn:
rclone OOMs on high-concurrency GPU nodes regardless of GPU model — observed on an a10g `g5.48xlarge`, not just H100/B200. #874 keyed the 4Gi split on `instance-gpu-name In [h100,b200]`, leaving a10g at 1Gi.

Example failure (a10g): https://github.com/pytorch/pytorch/actions/runs/28908017258/job/85761532112 — node `ip-10-4-176-7` (`g5.48xlarge`, 8× a10g) on meta-prod-aws-ue2, rclone OOMKilled → `/mnt/hf_cache` dead FUSE → job pod `CreateContainerError`.

Render one mount DaemonSet per GPU-count tier (`karpenter.k8s.aws/instance-gpu-count`), reserved (`request == limit`) memory scaled to concurrency:

| tier | affinity | DaemonSet | memory |
|---|---|---|---|
| 8-GPU | `In [8]` | hf-cache-mount-gpu8 | 4Gi |
| 4-GPU | `In [4]` | hf-cache-mount-gpu4 | 2Gi |
| 2-GPU | `In [2]` | hf-cache-mount-gpu2 | 1Gi |
| 1-GPU | `In [1]` | hf-cache-mount-gpu1 | 512Mi |
| CPU / rest | `NotIn [1,2,4,8]` | hf-cache-mount | 256Mi |

Affinity keeps tiers mutually exclusive (the `NotIn` catch-all also matches non-GPU nodes, so CPU is covered and every node still gets a mount to clear the startup taint). Retires the pre-tier `-largegpu` DaemonSet.

Also counts hf-cache in the packing analysis (it's a `.tpl` the YAML scan skipped): `daemonset_overhead` adds a 256Mi base on all nodes, and `analyze_node_utilization` adds the per-GPU-count top-up (`hf_cache_gpu_topup_mib`) — a flat constant can't express the tiers.

### Packing impact (`just analyze-utilization`)

Large-RAM GPU nodes (p5/p6-b200/p4d, 2TB / 1TB) have 50–86Gi of slack — the reservation is negligible there. The small-RAM GPU nodes run a single runner sized to ~full node memory, so reserving there needs a matching trim. Job memory dropped by the tier reservation (labels keep their `-NNN-`, names stable):

| runner | node | trim |
|---|---|---|
| a10g-8 / t4-8 | g5.48xlarge / g4dn.metal | −4Gi (704→700, 344→340) |
[...]

PR #890, opened 2026-07-09
osdc/hf-cache: raise 1-GPU rclone reserve to 1Gi

huydhn:
## What
Raise the `hf-cache-mount-gpu1` rclone memory tier **512Mi → 1Gi**, and trim the three single-GPU runner defs (`arc-runners` + `arc-runners-opt`) to fit under the larger `request == limit` reserve.

## Why
#883 tiered rclone memory by GPU count and gave the 1-GPU tier 512Mi. But under `--vfs-cache-mode full`, rclone RSS tracks the **model files a job reads, not GPU count** — and 1-GPU a10g nodes run the `inductor_huggingface` suite, which loads the largest models. rclone crosses 512Mi → **OOMKilled** → the node-wide `/mnt/hf_cache` FUSE mount dies → the job's mmap'd safetensors read faults with **SIGBUS (`Run failed with return code: -7`)**, which surfaces as a *misleading* `LocalEntryNotFoundError` ("missing weights"). 1Gi restores the floor these nodes ran at before #883.

### Evidence
Live rclone OOMs, both prod clusters, **gpu1 tier only** (cpu/gpu4/gpu8 = 0):
- meta-prod-aws-ue2: **17/59** mounts `OOMKilled`
- meta-prod-aws-ue1: **9/80** `OOMKilled`

Flaky because the OOM only trips when rclone crosses 512Mi mid-load — same job, same ue1 bucket:
- fail: pytorch/pytorch runs/29433922962/job/87421947313 (`Loading weights: 25%|…| 36/146 … return code: -7`)
- pass: pytorch/pytorch runs/29436575323/job/87456375877

The file isn't missing (the load reached 25% of 146 shards); the `LocalEntryNotFoundError` is a retry cascade after the SIGBUS.

## Runner trims (label keeps `-NNN-`)
- a10g `g5.8xlarge` / l4 `g6.8xlarge`: 113 → 112Gi
- t4 `g4dn.8xlarge`: 115 → 114Gi

## Deploy
Redeploy `hf-cache` + `arc-runners` (+`arc-runners-opt`) to the clusters running these 1-GPU classes.

### DIFF
--- a/osdc/modules/arc-runners-opt/defs/l-x86aavx2-29-113-a10g.yaml
+++ b/osdc/modules/arc-runners-opt/defs/l-x86aavx2-29-113-a10g.yaml
@@ -1,12 +1,12 @@
 # ARC runner definition: l-x86aavx2-29-113-a10g
 # Instance type: g5.8xlarge — 1-GPU A10G, 32c/128Gi node
-# Actual K8s capacity ~118.4Gi. Max job memory after all overhead: 113Gi.
+# Actual K8s capacity ~118.4Gi. Job 112Gi after overhead incl. hf-cache 1Gi reserve; label keeps -113-.
 runner:
   name: l-x86aavx2-29-113-a10g
   instance_type: g5.8xlarge
   disk_size: 150
   vcpu: 29
-  memory: 113Gi
+  memory: 112Gi
   gpu: 1
   proactive_capacity: 0
   hud_failure_base_capacity: 30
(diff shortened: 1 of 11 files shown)


jeanschmidt:
lets pause on this, as I believe this can be a problem to CPU on -opt fleet

jeanschmidt (approved):

jeanschmidt (changes requested):
we should reduce the memory usage, not raise memory for the pod

jeanschmidt:
we can SUBSTANTIALLY reduce memory usage, without significant performance drop, potentially to 128Mib.

rclone is a go binary, and we can easily control go GC behavior and memory footprint without requiring infinite memory. The problem here is GC, NOT how much memory the process requires.

few knobs:

* --buffer-size - per file buffer size
* --max-buffer-memory - total memory we can use for buffering
* GOMEMLIMIT / GOGC - Set this to a few MiB lower than the total memory available, this will tell GO VM when it should pause and run a GC, over infinite piling memory

jeanschmidt:
with this, we can probably get rid of the stagger multi-size memory we set up for GPUs

huydhn:
Closing this as I haven't seen OOM for a couple of days, so I would assume that the GOMEMLIMIT fix does the trick

huydhn:
Not need this anymore

PR #891, opened 2026-07-09
integration-tests: derive resource assertions from runner defs instead of hardcoding

jeanschmidt:
**Impact:** integration-test workflow generation (CI only)
**Risk:** low

## What
The generated integration-test workflow now pulls its CPU/memory/GPU assertion values straight from the enabled arc-runners `defs/` at generation time, via `{{VCPU__...}}`/`{{MEMGI__...}}`/`{{TORCHMEM__...}}`/`{{GPU__...}}` placeholders, replacing the hardcoded `EXPECTED=...` literals in `integration-test.yaml.tpl`.

## Why
Every time a runner's vcpu/memory/gpu spec changed, the integration-test template's hardcoded expectations drifted and had to be updated by hand (see the recent `-opt` runner right-sizing and the t4-multi memory fixup). Deriving the values from the same runner defs the cluster actually deploys keeps the tests in sync automatically.

# Notes
- `resource_placeholders()` processes the base `arc-runners` module first, then specialized variants (`-opt`/`-b200`/`-h100`) sorted, so a variant that redefines a shared runner name wins — this is what lets the ue1 `-opt` sizes flow through.
- Substitution runs before conditional-block stripping; a dedicated guard after stripping fails generation if a resource placeholder survives inside a live job (unresolved runner def = real bug), while placeholders in removed jobs are cleaned up harmlessly.
- Adds an import path to `modules/arc-runners/scripts/python` to reuse `parse_memory_bytes` rather than reimplement memory parsing.

### DIFF
--- a/osdc/integration-tests/scripts/python/phases.py
+++ b/osdc/integration-tests/scripts/python/phases.py
@@ -16,8 +16,10 @@ from pathlib import Path
 import yaml
 
 sys.path.insert(0, str(Path(__file__).resolve().parents[3] / "scripts" / "python"))
+sys.path.insert(0, str(Path(__file__).resolve().parents[3] / "modules" / "arc-runners" / "scripts" / "python"))
 
 from conditional_blocks import strip_conditional_block
+from generate_runners import parse_memory_bytes
 from nodepool_defs import is_excluded_for_region
 from run import (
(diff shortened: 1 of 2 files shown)


huydhn (approved):
Nice! This is much easier

PR #907, opened 2026-07-16
osdc/hf-cache: cut rclone RSS with --buffer-size 4M --use-mmap

huydhn:
## Problem
On the 1-GPU mount tier (512Mi), concurrent large-model (safetensors) reads OOM-kill rclone → the node-wide `/mnt/hf_cache` FUSE mount dies → the job's mmap'd read faults with **SIGBUS**, surfacing as a misleading `LocalEntryNotFoundError` ("missing weights").

Example — failing `inductor_huggingface` a10g job:
```
Loading weights: 25%|██▍| 36/146 … Run failed with return code: -7   # SIGBUS
huggingface_hub.errors.LocalEntryNotFoundError: Cannot find the requested files in the disk cache ...
```
Stats — live rclone `OOMKilled` mounts, **gpu1 tier only** (cpu/gpu4/gpu8 = 0): ue2 **17/59**, ue1 **9/80**.

## Fix
rclone RSS under `--vfs-cache-mode full` is dominated by per-open-file read-ahead buffers, not GPU count:
- `--buffer-size 4M` — 4x cut from rclone's 16Mi-per-open-file default read-ahead (the RSS driver when a model opens many shards); kept non-zero so cold reads retain some prefetch, with the on-disk cache serving the rest.
- `--use-mmap` — return freed buffers to the OS instead of Go retaining them as RSS.

Keeps rclone inside the reserve without raising the limit. Complements #890 (512Mi→1Gi headroom).

### DIFF
--- a/osdc/modules/hf-cache/README.md
+++ b/osdc/modules/hf-cache/README.md
@@ -18,6 +18,12 @@ rclone's memory is **reserved** (`request == limit`) and **tiered by GPU count**
 `deploy.sh` renders one DaemonSet per tier — 8-GPU → 4Gi, 4-GPU → 2Gi, 2-GPU → 1Gi,
 1-GPU → 512Mi, and the CPU catch-all → 256Mi. See `MOUNT_TIERS` in `deploy.sh`.
 
+To keep RSS inside those reserves the mount runs with `--buffer-size 0` (no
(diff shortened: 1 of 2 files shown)


jeanschmidt (approved):

huydhn:
Deploy on ue1 staging, smoke and integration tests are passing

PR #914, opened 2026-07-17
Add `just force-unlock` recipe to clear stuck OpenTofu state locks

jeanschmidt:
**Impact:** dev/ops tooling only — no cluster or runtime behavior changes
**Risk:** low

## What
Adds a `force-unlock <cluster> <slot> <lock-id>` justfile recipe that inits the correct backend for a cluster's base or module state and runs `tofu force-unlock` against the given lock ID.

### DIFF
--- a/osdc/justfile
+++ b/osdc/justfile
@@ -836,6 +836,69 @@ plan cluster:
             -var="cluster_id=${CLUSTER}"
     done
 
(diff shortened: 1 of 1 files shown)


huydhn:
@claude Verify if all the checks and setup before tofu init and tofu force-unlock are needed

huydhn (approved):
Look like there might be some redundant steps (probably copy/paste) but overall LGTM!

PR #915, opened 2026-07-17
Revert #907 + #913: restore pre-#907 hf-cache rclone config

huydhn:
Reverts #913 and #907, returning the hf-cache rclone mount to its pre-#907 config (no `--use-mmap`, no `--buffer-size 4M` — rclone defaults + the per-tier `GOMEMLIMIT`). This is the same config `meta-prod-aws-ue1` runs today with zero small-tier OOMs.

#907 added `--buffer-size 4M --use-mmap` and #913 later dropped only `--use-mmap`; post-deploy verification on ue2 showed small-tier (256Mi CPU / 512Mi gpu1) OOMs continued regardless, so neither flag was the driver. The real cause is that rclone mount memory is tiered by GPU count while per-node RSS scales with pod density — ue2 packs runners onto 48xl/metal nodes, overwhelming the single 256Mi per-node mount. Reverting to the known-good baseline while the tiering fix is developed separately.

### DIFF
--- a/osdc/modules/hf-cache/README.md
+++ b/osdc/modules/hf-cache/README.md
@@ -18,14 +18,6 @@ rclone's memory is **reserved** (`request == limit`) and **tiered by GPU count**
 `deploy.sh` renders one DaemonSet per tier — 8-GPU → 4Gi, 4-GPU → 2Gi, 2-GPU → 1Gi,
 1-GPU → 512Mi, and the CPU catch-all → 256Mi. See `MOUNT_TIERS` in `deploy.sh`.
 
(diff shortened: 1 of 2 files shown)


jeanschmidt:
I believe we should keep the `-buffer-size 4M` anyways

huydhn:
Superseding this in favor of #917, which keeps a per-tier `--buffer-size` (CPU 0 / GPU 4M) rather than reverting to rclone defaults. Post-deploy evidence showed the OOM driver is node pod-density on the small tiers, not the mmap/buffer flags per se, so a full revert doesn't address it.

PR #923, opened 2026-07-17
hf-cache: gate deploy on rollout being applied, not on every pod Ready

jeanschmidt:
**Impact:** hf-cache module deploys (CI/deploy tooling only)
**Risk:** low

## What
Replaces `kubectl rollout status` for the mount DaemonSets with a small polling helper (`wait_ds_rollout`) that waits for the current pod template to be *scheduled* to every targeted node (`updatedNumberScheduled == desiredNumberScheduled`) rather than for every pod to become Available.

## Why
The mount DaemonSets run on the github-runner fleet, where nodes churn constantly and each fresh node's rclone mount takes ~30-60s to warm. `kubectl rollout status` waits on `numberAvailable == desiredNumberScheduled`, which never settles on a churning fleet, so every deploy burned the full 300s timeout waiting on pods that are just normal steady-state churn. Gating on the controller having applied the current generation to all targeted nodes is the correct "rollout done" signal here.

### DIFF
--- a/osdc/modules/hf-cache/deploy.sh
+++ b/osdc/modules/hf-cache/deploy.sh
@@ -127,12 +127,46 @@ echo "[hf-cache] Applying mount DaemonSets (per GPU-count tier)..."
   done
 } | kubectl_apply_if_changed -f -
 
+# Gate on the rollout being *applied* to every targeted node, not on every pod being
+# Available. `kubectl rollout status` waits for numberAvailable == desiredNumberScheduled;
+# these DaemonSets run on the github-runner fleet, where nodes churn continuously and each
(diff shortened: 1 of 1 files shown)


jeanschmidt:
@claude

Please review the commit and certify that the `wait_ds_rollout` logic will produce the desired effects. Few things to consider:

a) Could this cause the release to move forward in case of not fully deployed changes (except from timeout)?
b) is this change crash-proof from the common scenarios?
c) Are there pieces of bash script that are bad practices/risky and should be improved? Could those cause breakages in the specific cases they are being used?
f) Is this change correct, and will it cause the desired intent (still wait for the update to happen and old pods be replaced, while not blocking on node churning)?

huydhn (approved):

PR #930, opened 2026-07-21
Bump bin-pack-scheduler resources and make replica count per-cluster

jeanschmidt:
**Impact:** bin-pack-scheduler pods on all OSDC clusters
**Risk:** medium

## What
Raises the bin-pack-scheduler container resources to guaranteed 3 CPU / 6Gi (requests == limits, up from 200m/256Mi requests) and makes the replica count configurable per cluster via a new `bin_pack_scheduler.replicas` key in `clusters.yaml` (default 3, set to 1 on the three staging clusters). The kustomize manifests move under `kubernetes/base/` and a new `deploy.sh` renders them while substituting the per-cluster replica count.

## Why
The scheduler was hitting its old CPU/memory ceiling under real scheduling load, degrading placement decisions (and OOMing). Prod clusters now get more headroom and an extra HA replica, while staging stays lean at a single replica to avoid wasting capacity.

# Notes
- HA is still via leader election — only one replica schedules at a time; the others are hot standbys.
- `deploy.sh` validates the replica value is a positive integer and asserts exactly one `replicas:` line is substituted, failing the deploy otherwise.


huydhn (approved):
LGTM! I think this is responsible for the queue today where all prod fleets seem to hit a limit and stop scaling up

PR #932, opened 2026-07-21
Add retrying OSDC setup composite action and use it across CI workflows

jeanschmidt:
* #933
* __->__ #932

**Impact:** CI only — OSDC deploy, drain, undrain, and pre-merge workflows
**Risk:** low

## What
Extracts the repeated "install just + mise + `uv sync`" setup into a new local composite action (`.github/actions/osdc-setup`) and swaps the four OSDC workflows over to it. Each setup step (plus the QEMU setup in `_osdc-deploy.yml`) is now retried on transient failure via retry-by-duplication.

## Why
The toolchain install steps were copy-pasted across `_osdc-deploy.yml`, `osdc-drain.yml`, and `osdc-undrain.yml`, and occasionally failed on transient network hiccups — flaking deploy and merge-queue runs. Consolidating into one action removes the duplication and adds retries in a single place.

The retry is done by duplicating each step (`continue-on-error` on the first attempt + a copy gated on its failure) rather than pulling in a third-party retry action, so no unpinnable code runs on the deploy / merge-queue path. `actions/checkout` and `configure-aws-credentials` are left alone since they retry internally.

# Notes
The SKILL.md doc update also corrects a stale reference: the `_osdc-deploy.yml` skip input is `skip_lint_test`, not `skip-checks`.

### DIFF
--- a/.github/actions/osdc-setup/action.yml
+++ b/.github/actions/osdc-setup/action.yml
@@ -1,0 +1,42 @@
+name: OSDC setup
+description: Install just, the mise toolchain, and Python dependencies for OSDC workflows, retrying transient failures.
+runs:
+  using: composite
+  steps:
+    # GitHub Actions has no native per-step retry, so each step below is retried
+    # by duplication: continue-on-error on the first attempt plus a copy gated on
+    # its failure. Preferred over a retry action so that no unpinnable
+    # third-party code runs on the deploy / merge-queue paths.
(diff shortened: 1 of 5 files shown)


huydhn:
@claude Do a sanity check to confirm if all the steps we need to setup OSDC are now added to the new composite GHA

huydhn (approved):

PR #933, opened 2026-07-21
Retry OSDC smoke tests once after a settle wait

jeanschmidt:
* __->__ #933
* #932

**Impact:** CI only — the `smoke` job in the OSDC deploy workflow
**Risk:** low

## What
Wraps the smoke test step in `_osdc-deploy.yml` with a retry-by-duplication: the first run is marked `continue-on-error`, and on failure the job waits 180s for the cluster to settle before running smoke a second time. The job timeout is bumped from 30 to 60 minutes to accommodate the extra attempt, and the deployment skill doc is updated to note the new smoke retry.

## Why
Smoke failures right after a deploy are usually transient — the cluster needs time to converge before everything is healthy. A single automatic retry avoids spurious red on the deploy path without pulling in an unpinnable third-party retry action, staying consistent with the existing retry-by-duplication pattern used for the `osdc-setup` composite action.


.github/workflows/_osdc-deploy.yml
```diff
@@ -129,7 +129,17 @@ jobs:
           aws-region: us-west-1
           role-duration-seconds: 7200
 
+      # Smoke failures are usually transient cluster churn (the cluster needs
+      # time to converge after deploy), so retry once after a settle wait.
       - name: Run smoke tests
+        id: smoke
```
huydhn, line 135: Why not use https://github.com/nick-fields/retry#retry_wait_seconds here? It looks nicer with that GHA

jeanschmidt: Believe or not, this was the first implementation agents came up with. Then the full set of security review flagged that it uses internally a js script that is not pinned. So even if you pin the action to a specific hash, the code it actually runs is downloaded from a tag and is not safe from supply-chain attacks.

This leads me to drop the idea of using it for retry on context/workflows with critical infra tokens. I asked the agent to comment that, but it limited itself to mention in the previous PR "preferred over third party retry mechanisms that are unpinneable" - fail to mention the nick-fields and why it should not be used.

Do you believe I should document this?

izaitsevfb (approved):

PR #934, opened 2026-07-21
dcgm-exporter: lower GOMEMLIMIT to 410MiB and only fail smoke test on fleet-wide OOM

jeanschmidt:
**Impact:** GPU-node monitoring (dcgm-exporter) and the monitoring smoke test — CI signal only, no runner impact
**Risk:** low

## What
Drops the dcgm-exporter `GOMEMLIMIT` from 450MiB to 410MiB (~88% → ~80% of the 512Mi cgroup limit) and relaxes the smoke test so it only fails when dcgm-exporter is OOM/crashlooping across several nodes rather than on a single pod.

## Why
dcgm-exporter memory scales with GPU count, so the occasional pod on a large multi-GPU node still gets OOMKilled near the ceiling. Two problems followed: the Go GC ceiling sat too close to the cgroup limit, and the smoke test failed CI on a single isolated OOM — expected noise, not a regression. Lowering `GOMEMLIMIT` gives GC more headroom before the kernel steps in, and the test now only flags a fleet-wide pattern (≥3 nodes with restarting/crashlooping pods), which is the signal that actually indicates a real problem.

# Notes
The old threshold was "any single pod with >3 restarts"; the new logic tracks distinct bad nodes and fails at ≥3 nodes with ≥2 restarts (or CrashLoopBackOff). The SKILL.md observability notes are updated to match the new values.


huydhn (approved):
LGTM!

PR #944, opened 2026-07-23
Right-size lf-prod base nodes to m7i.4xlarge

zxiiro:
lf-prod-aws-ue1/ue2 base nodes on m7i.8xlarge sit well under capacity: 14-day peak CPU was 37% and real memory usage measured via kubelet stats/summary topped out at 10.1% of 128GiB. Real memory drivers (git-cache-central, karpenter) and CPU peaks both fit comfortably within m7i.4xlarge, freeing up cost without affecting workload scheduling.

### DIFF
--- a/osdc/clusters.yaml
+++ b/osdc/clusters.yaml
@@ -525,7 +525,7 @@ clusters:
         - lf_osdc_admin
     base:
       base_node_count: 4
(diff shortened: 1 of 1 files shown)


atalman (approved):
lgtm

PR #946, opened 2026-07-24
Hard-spread base-node controllers to prevent CPU pile-up

zxiiro:
The ue2 base-node resize to m7i.4xlarge (COPS-418) exposed a latent scheduling gap: bin-pack-scheduler, the ARC controller, and Harbor's own multi-replica components had no hard per-node spread, so a rolling base node replacement could pile every replica of each onto whichever node came up first. On the smaller instance type this saturated one node's CPU requests, blocking harbor-database (which is AZ-pinned by its EBS volume) from scheduling and cascading into a full Harbor outage.

- bin-pack-scheduler: preferred pod anti-affinity was too weak to stop the pile-up; switch to required.
- ARC controller and Harbor's core/portal/nginx: add topologySpreadConstraints (maxSkew 1, DoNotSchedule).
- registry/jobservice deliberately excluded: their RWO PVCs already force same-node placement across replicas.

### DIFF
--- a/osdc/base/helm/harbor/values.yaml
+++ b/osdc/base/helm/harbor/values.yaml
@@ -27,6 +27,17 @@ x-base-affinity: &base-affinity
               values:
                 - base-infrastructure
 
+# Hard-spread each component's replicas one-per-node. Without this, a
+# rolling base node replacement (nodes become Ready one at a time) can pile
+# every replica of core/registry/etc. onto whichever node comes up first,
+# starving other pods on that node of CPU — including harbor-database,
+# whose EBS volume pins it to one specific AZ/node. Component name is
+# templated in per-component below since each has a distinct label.
+x-spread-template: &spread-template
+  maxSkew: 1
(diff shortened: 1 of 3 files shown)


jeanschmidt:
humm, I am not sure if this shouldn't be managed via podAntiAffinity. Ideally `preferredDuringSchedulingIgnoredDuringExecution` to increase system resilience.

zxiiro:
`preferredDuringSchedulingIgnoredDuringExecution` is what allowed the pods to all pile on a single node. I don't think want this. Requiring ensures that doesn't happen.

@claude wouldn't `preferredDuringSchedulingIgnoredDuringExecution` allow the situation we are trying to prevent deploy multiple pods to the same node?

jeanschmidt (approved):

jeanschmidt:
@zxiiro sorry forgot about this PR, OK got your explanation and it seems correct, approved.

Thanks for the contribution!

PR #947, opened 2026-07-24
Remove Legacy ARC

zxiiro:
These files are from the legacy ARC deployment that is now superseeded by OSDC.


huydhn (approved):
LGTM! Let's clean this up

PR #948, opened 2026-07-24
Enable CUDA fabric-handle IPC on H100 nodes and add a dedicated fabric runner

jeanschmidt:
**Impact:** H100 nodes
**Risk:** medium

## What
Creates the IMEX channel device on H100 nodes so PyTorch expandable-segments IPC can use CUDA fabric handles, adds an optional `imex_channels` field to runner defs (renders `NVIDIA_IMEX_CHANNELS` onto the job container), a dedicated low-cap fabric runner def, and a manual validation probe to confirm fabric handles are actually used.

## Why
CUDA VMM fabric handles (`CU_MEM_HANDLE_TYPE_FABRIC`), used by torch expandable-segments IPC, need `/dev/nvidia-caps-imex-channels/channel0` to exist even on a single host. The driver doesn't auto-create it, so without this the allocator silently falls back to POSIX-fd handles — a "false green" where the workload passes but never exercised the fabric path. This wires up the channel end-to-end (node device → container env → validation).

## Notes
- `h100-node-setup.sh` creates the channel for the current boot and enables a `oneshot` systemd unit for reboots. Creation runs under `if` so a failure logs but never trips `set -e` — a node without fabric handles is recoverable, a failed cloud-init is not. This runs on every H100 node provision.
- The runner template change is byte-identical to prior output when `imex_channels` is unset; the field gates on `is not None` so `0` is honored (covered by the new tests).
- The new fabric runner is capped at `max_runners: 1` and depends on p5.48xlarge Capacity Blocks being available.


jeanschmidt:
@claude 

This change node startup script osdc/modules/nodepools-h100/scripts/h100-node-setup.sh

My main concern is that this can introduce failure modes that will make nodes fail to startup. Please review very carefully it. Is it written defensively? Does it have retries in place? If it fails, can this blows up the instance startup preventing it from joining? In case of failures, are there fallbacks and protections?

huydhn (approved):
Overall LGTM! Once this is confirmed to work in prod, I think we should consider rename `-fab` runners to keep only one set of H100 runner definitions because the underlying nodes are the same.

huydhn:
@claude Are there any metrics we can use to check the health of H100 runners when they start to use the new imex fabric channel? We want to be proactive if it starts causing issues

jeanschmidt:
> Overall LGTM! Once this is confirmed to work in prod, I think we should consider rename -fab runners to keep only one set of H100 runner definitions because the underlying nodes are the same.

the reason I decided to create a separate pool of runners is that those tests are gated on runner availability, and if we enable this feature we might see a big spike on test time and tests runs, I want us to selectively enable what we want, or not.

PR #950, opened 2026-07-27
multi-tenant: add B200 GHA runner provisioning

huydhn:
## What

Adds the `multi-tenant/` toolkit that provisions on-prem **DGX B200** hosts as multi-tenant PyTorch self-hosted GitHub Actions runners.

Ansible creates per-GPU users with dedicated **cgroup slices** (CPU / memory / GPU partitioned by GPU count), installs **rootless Docker** + the nvidia container toolkit, and runs the `ghad-manager` systemd service that keeps one ephemeral runner container alive per user.

## Layout

| Path | Purpose |
|------|---------|
| `playbooks/` | host provisioning (`setup-host.yml`) + service-restart playbooks |
| `services/ghad-manager/` | per-user runner-container manager daemon + systemd unit |
| `scripts/multi-tenant-gpu-main.sh` | in-container runner bootstrap |
| `images/multi-tenant-gpu/Dockerfile` | the runner image |
| `inventory/manual_inventory` | manually-managed B200 fleet inventory |
| `justfile`, `mise.toml`, `pyproject.toml`, `uv.lock` | task runner + pinned tooling / Python deps |
| `README.md`, `doc/dev-howto.md` | runbook + dev how-to |

## Running it

`just setup-host <group|host>` (e.g. `just setup-host b200_1`) provisions the fleet. A few behaviors worth calling out:

- **Auth:** the `pytorch` login uses SSH password auth (no keys), so recipes prompt for the SSH (`-k`) and sudo (`-K`) password — same value on this fleet.
- **VPN resilience:** `ANSIBLE_TIMEOUT=60` plus SSH reconnect retries and keepalives (in `mise.toml`) ride over slow sudo / transient drops on the live GPU hosts behind the VPN.
- **Image refresh:** every `setup-host` re-pins the newest ECR image tag into `temp/` (prerequisite recipe), and fails loudly rather than shipping a bogus tag if the query is empty.

## Cleanup included

On top of the straight migration, this applies a zero-risk cleanup pass on the code being moved in:

- **`ghad-manager.py`** — drop dead code (unused IMDSv2 / drain-label constants, the never-called `get`/`remove_self_hosted_runner_org` helpers, the unreachable `trap_stuck_inactive_daemon` and its call, and the now-unused PyGithub imports); fix `login_to_dockerhub`'s mis-declared type hints (they were default values, `=str`/`=int`) and the `get_github_app_id`/`installation_id` return annotations (`-> str` → `-> int`).
- **playbooks** — drop the unused `runner_version` var and the dead `/opt/runner` task; de-duplicate the `/opt/docker` task name and give directories mode `0755` (was `0644`).

## Notes

- **Network access:** the B200 hosts sit behind NVIDIA's Atmos (Axis Security) ZTNA VPN, so the Ansible recipes must be run from a host that is already connected to that VPN. Automating this from ephemeral CI is not currently feasible — see the comment below for details.
- Secrets (GH app key, ECR creds, instance labels) are loaded at runtime from a git-ignored `secrets.env` — nothing sensitive is committed.
- Deeper follow-ups (token-on-CLI → `--password-stdin`, the per-job `/mnt/hf_cache` `0777` re-chmod, ECR-registry constant dedup, `awscli` off pip) are intentionally left for a separate PR to keep this one reviewable.

### DIFF
--- a/.checkov.yml
+++ b/.checkov.yml
@@ -5,6 +5,7 @@ skip-path:
   - arc
   - crcr
   - osdc
+  - multi-tenant
   - '.github/workflows/arc*'
 
 # Skip INFO and other unresolvable checks
--- a/multi-tenant/.gitignore
+++ b/multi-tenant/.gitignore
@@ -1,0 +1,3 @@
+.venv/
+temp/multi-tenant-gpu-tag
+__pycache__/
--- a/multi-tenant/README.md
+++ b/multi-tenant/README.md
@@ -1,0 +1,46 @@
+# multi-tenant
+
+## A brief warning
+Please be aware that running ansible locally can be harmful and dangerous. Those operations are destructive and running multiple in parallel can hinder hosts in a bad state, potentially hard to recover. So, if you are planning to do so, please communicate and reach out to teamates to make sure nothing is running in parallel.
+
+The recommended approach is to open a PR and test your changes by pushing to the PR, it is safer: keys are already there. Also, other users can see your actions and changes. And more importantly: it should be concurrency protected.
+
+## Setup
+Prerequisites: [`mise`](https://mise.jdx.dev/) and [`just`](https://just.systems/).
+mise pins the toolchain (`mise.toml`: python, uv, awscli, jq) and
+[uv](https://docs.astral.sh/uv/) manages the Python deps (`pyproject.toml`).
+
+Sync the Python environment once:
+```bash
+just setup
+```
+
+Secrets are loaded automatically from `secrets.env` (git-ignored). It must export
+`GH_APP_PK`, `ECR_READONLY_AWS_CREDENTIALS`, and `INSTANCE_LABELS`.
+
+## Running ansible against the B200 hosts
+```bash
+just setup-host                 # all B200 hosts
+just setup-host dgxb200-03      # a single host / group
+```
+Both prompt for the `pytorch` sudo password (`-K`). See `just --list` for the
+full set of recipes.
+
+## Inventory
+The ansible inventory is manually managed and can be found on `inventory/manual_inventory`, please make sure to keep it up to date.
+
+Some commands for debug:
+
+```bash
+# for checking which cgroups a process is aligned with
+systemd-cgls
+# logging in as a specific user with systemd enabled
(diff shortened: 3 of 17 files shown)


huydhn:
cc @nWEIdia (Let me see how to add you as a reviewer)

jeanschmidt:
I trust that you migrated the code carefully :)

jeanschmidt (approved):

huydhn:
## VPN access & why the Ansible run can't (yet) be a plain CI job

The B200 hosts are only reachable over NVIDIA's **Atmos (Axis Security / HPE Aruba) ZTNA VPN**, so today the playbooks are run from a workstation already on that VPN. I looked into whether an **ephemeral CI runner could bring the VPN up per job**. With the stock agent, it can't — three blockers:

1. **No headless / CLI login.** `agentd` is a background daemon driven by an Electron GUI; the only login path is an **interactive browser IdP sign-in**. There is no scriptable `login`/`connect` command.
2. **Enrollment state is encrypted and device-bound.** The stored `authToken` is opaque ciphertext (not a reusable JWT), and workspace/tenant/tunnel state are encrypted at rest against the local device — so a token can't be captured on one machine and dropped onto a runner.
3. **Device-posture enforcement** (osquery / CrowdStrike). A throwaway CI runner won't pass managed-device posture even if 1 and 2 were solved.

The Linux **agent install** itself is fine (`.deb` for Ubuntu/Debian, `.rpm` for Fedora/RHEL/CentOS).

### Ideas to explore
- A persistent self-hosted runner sitting inside the VPN.
- Agentless access via NVIDIA IT (Atmos clientless SSH, an IPsec tunnel, or an Atmos Connector).

This PR is unaffected — it only adds the tooling; the runner/VPN wiring will be a follow-up once we pick a direction.

huydhn:
> I trust that you migrated the code carefully :)

Better yet, I test deploy this on DGX fleet :P

huydhn:
As an example, I use `INSTANCE_LABELS=linux.dgx.b200 just setup-host b200_1` to test the setup of B200 runners using the change from this PR

PR #953, opened 2026-07-28
zombie-cleanup: protect busy ARC runners, reap only hung ones

jeanschmidt:
**Impact:** zombie-cleanup
**Risk:** high

## What
Teaches the zombie-cleanup CronJob to tell an idle runner apart from one actively running a job. Runners executing a workflow are now protected from age-based reaping until they cross a much larger busy hard cap (48h, presumed hung), and every delete candidate is re-checked against fresh per-pod state immediately before deletion so a runner that picks up a job mid-run is never killed. Also bumps `running_max_age_hours` 12h→24h and adds the `busy_max_age_hours` knob.

## Why
The previous logic reaped any runner pod once it had been `Running` past the age threshold, with no notion of whether it was doing work. A long-running-but-legitimate job (up to GHA's 6h max plus prepare time) sitting past the threshold would get its pod deleted out from under it, failing the job. This keeps the GC's value — cleaning up genuinely stuck/orphaned pods — while no longer disrupting live jobs. Busy state is derived from the owning `EphemeralRunner` (`status.jobId`) and from live workflow/step pods that pin a runner via the `runner-pod` label.

### DIFF
--- a/osdc/modules/zombie-cleanup/scripts/python/runner_busy.py
+++ b/osdc/modules/zombie-cleanup/scripts/python/runner_busy.py
@@ -1,0 +1,254 @@
+"""GitHub Actions runner busy-state detection and protection classifier.
+
+Runner pods are owned by an EphemeralRunner custom resource and stay eligible for
+age-based cleanup, unlike controller-managed pods. This module distinguishes an
+idle runner (reapable) from one actively executing a job (protected): a runner is
+busy when its EphemeralRunner carries a status.jobId, or when a live workflow/step
+job pod pins it via the runner-pod label.
+
+classify_pod is the single decision function shared by the find pass and the
+pre-delete recheck, so the two paths cannot drift. The find pass feeds it liveness
+resolved from one batch listing; the recheck feeds it liveness resolved from a
+targeted per-pod GET.
+"""
+
+import logging
+from dataclasses import dataclass
+from enum import Enum
+
+from lightkube import Client
+from lightkube.core.exceptions import ApiError
+from lightkube.generic_resource import create_namespaced_resource
+from lightkube.resources.core_v1 import Pod
+
+log = logging.getLogger("zombie-cleanup")
+
+# GitHub Actions runner pods carry this ownerReference kind. Deliberately not a
+# controller-managed kind: idle runners stay reapable, busy ones are gated instead.
+EPHEMERAL_RUNNER_KIND = "EphemeralRunner"
+
+# Bare workflow/step job pods carry this label; its value is the owning runner
+# pod's name and is the join key back to the runner. Kubernetes label VALUES are
+# capped at 63 chars (pod names may reach 253), so this join holds only because
+# OSDC runner names stay well under 63 (the ~42-char runner-naming convention);
+# a longer runner name would be truncated in the label value and break the match.
+RUNNER_POD_LABEL = "runner-pod"
+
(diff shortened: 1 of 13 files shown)


osdc/modules/zombie-cleanup/scripts/python/runner_busy.py
```diff
@@ -0,0 +1,254 @@
...
+def age_over_threshold(phase: str | None, age_hours: float, pending_max: int, running_max: int) -> int | None:
+    """Return the exceeded age threshold (hours) for a pod, or None if within limits."""
+    if phase == "Pending" and age_hours > pending_max:
+        return pending_max
+    if phase in ("Running", "Unknown") and age_hours > running_max:
+        return running_max
+    return None
+
+
+def classify_pod(
```
huydhn, line 128: This function doesn't classify placeholder pods, asking Claude to walk me through the logic there:

```
1. is_managed_pod → owner is a ConfigMap, not in {ReplicaSet,DaemonSet,StatefulSet,Job} → not skipped.
2. is_ephemeralrunner_pod → no EphemeralRunner owner → False.
3. get_runner_pod_label → no runner-pod label → None.
4. → falls straight through to _age_decision.
```

My understand is that keeps the current status quo for placeholder pods. So far so good, but is there any chance for shenanigan here? Like zombie-clean up reaps placeholder runner but not workflow, or vice versa

osdc/modules/zombie-cleanup/scripts/python/zombie_cleanup.py
```diff
@@ -67,89 +106,131 @@ def get_pod_age_hours(pod: Pod, now: datetime) -> float:
...
-    zombies = []
+
+    pods = list(client.list(Pod, namespace=namespace))
+    state = gather_busy_state(client, namespace, pods)
+
+    zombies: list[Pod] = []
     total_count = 0
     managed_count = 0
+    busy_skipped = 0
+    job_protected = 0
```
huydhn, line 123: Do we want to keep track of `failsafe_protected` here too? it looks like an important metric IMO as a high value of `failsafe_protected` means that zombie-cleanup is having issue doing its job

osdc/modules/zombie-cleanup/scripts/python/runner_busy.py
jeanschmidt: - It can happen — the two placeholders are reaped independently by age, and the cleanup cap (max(10%,10)) can split them across runs. But that's true before and after this PR (unchanged), so it's not a regression here.
- It's benign. A placeholder isn't running a job — it's a slot reservation. Reaping one doesn't kill a workload; ARC's capacity controller reconciles the pair (recreates/cleans the orphan), the same way the EphemeralRunner controller recreates a reaped runner pod. Worst case is reservation churn.
- It will only happen if placeholders misbehave - given they are designed to self destruct after a timeout, the arc listener removes them after being there for a while and never getting capacity, and destroyed with the listener when it restart via K8s dependencies thing I forgot the name.

This is the intended correct behavior

osdc/modules/zombie-cleanup/scripts/python/zombie_cleanup.py
jeanschmidt: Ok, I can add the metric here

huydhn (approved):

Issue #952, opened 2026-07-28
High GPU queue + flaky jobs


huydhn:
## Symptom
GPU (a10g/g5) jobs sit for the full 2h `PREPARE_JOB` timeout with the workflow pod `PodScheduled=False`, then fail with the generic `Executing the custom container implementation failed`. Scheduler events on those pods: `Insufficient nvidia.com/gpu … Preemption is not helpful`.

## Root cause (two layers)

**1. Primary — AWS g5 capacity exhaustion.** Karpenter can't launch g5 nodes: ~5,748 `InsufficientInstanceCapacity` / `UnfulfillableCapacity` (CreateFleet) in 2h on `meta-prod-aws-ue1`, across **all** g5 sizes (12/16/24/48xl); g6 also affected but ~50× less. This is genuine AWS-side A10G scarcity in us-east-1, not a config/quota issue (`g5-8xlarge` pool has `limits: null`, 0 nodes). Even a perfect system queues here.

**2. Amplifier — capacity-advertising leak in the ARC fork.** The listener's advertised MaxRunners = `runningRunners + runningPairs`, where `runningRunners` counted runner pods in `Running` phase. A runner pod runs on cheap c7i and is `Running` even while its GPU job pod is stuck `Pending`. So GPU-starved runners advertised phantom capacity → over-acquisition → more stuck runners → worse queue + more 2h-timeout failures.

Evidence (a10g scale set, incident window): `advertised_max_runners` climbed **123 → 202** while `running_pairs` stayed **0**; at 21:00 the warm pool was 59 pairs with **all 59 runner halves Running (c7i) but all 59 job halves Pending (no GPU)**.

## Fixes
- **ARC fork** — jeanschmidt/actions-runner-controller#12: reporter now counts runners whose **job pod is Running** (compute secured), not runner pods in Running phase. Restores back-pressure so jobs queue instead of being acquired-then-failed. Includes a scoped-List + fallback optimization.
- **OSDC** — pytorch/ci-infra#955: label job pods `osdc.io/scale-set-name` so the fork can scope its List (enables #12's optimization). Deploy #955 before bumping the fork chart.

## Remaining levers (fixes stop amplification, not the scarcity)
- **Instance-type diversification** in the g5/a10g nodepools (allow g6/g6e fallback — far less ICE) — highest leverage.
- Cap a10g/g5 scale-set `MaxRunners` to the real GPU ceiling.
[...]

PR #955, opened 2026-07-28
arc-runners: label job pods with osdc.io/scale-set-name

huydhn:
Stamp `osdc.io/scale-set-name` (= `runnerScaleSetName`) on job (workflow) pods via the hook template.

**Why:** job pods carry no scale-set label, so the ARC-fork capacity monitor can't scope them to a scale set and has to do a namespace-wide List. This label lets it use a single labeled List. Enables the reporter fix in jeanschmidt/actions-runner-controller#12 (count runners whose job pod is Running, scoped to the scale set) to drop the namespace scan.

**Notes:**
- Custom key (not `actions.github.com/*`) to avoid colliding with ARC's own selectors; value matches the scale-set label ARC already puts on runner + placeholder pods.
- `generated/` is git-ignored and regenerated at deploy, so only the template + smoke test are tracked. Verified the label resolves correctly across all 56 generated defs (base + h100 + b200).
- `just lint` + `just test` pass.

Context: pytorch/ci-infra#952

### DIFF
--- a/osdc/modules/arc-runners/templates/runner.yaml.tpl
+++ b/osdc/modules/arc-runners/templates/runner.yaml.tpl
@@ -387,6 +387,13 @@ metadata:
 data:
   job-pod.yaml: |
     metadata:
+      labels:
+        # Lets the capacity monitor scope job pods to this scale set with a
+        # single labeled List (job pods otherwise carry no scale-set label).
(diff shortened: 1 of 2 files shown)


huydhn:
We don't need this now after https://github.com/jeanschmidt/actions-runner-controller/pull/13

PR #963, opened 2026-07-29
Bump ARC chart to 0.14.1-jeanschmidt.18 on staging + prod-ue1

jeanschmidt:
*Impact:** ARC runner autoscaling on the three staging clusters and prod-ue1
**Risk:** high

## What
Pins the ARC fork chart to `0.14.1-jeanschmidt.18` on `meta-staging-aws-{uw1,ue1,ue2}` and `meta-prod-aws-ue1`, overriding the `defaults.arc.chart_version` of `.17`. The new chart ships the capacity-accounting rework from jeanschmidt/actions-runner-controller#13.

## Why
The current chart over-advertises capacity, so GitHub dispatches more jobs than the cluster can actually run, over-provisioning runners relative to real capacity. This was causing failure to start workflow pod that stayed on pending state. The fork PR fixes two failure modes behind this:

- **Count bug:** capacity was computed as `runningRunners + runningPairs` and double-counted workflow pods. A job needs both a runner pod *and* a separately-created workflow pod, so runners with no free workflow slot were still counted. The new logic reports `min(runnerSide, workflowSide)` and credits real workflow pods once per runner (matched by the `-workflow` suffix) instead of letting container/Docker-action step pods inflate the count.
- **Placeholders on tainted nodes:** placeholder pods sitting on cordoned or disruption-tainted nodes were counted as usable capacity even though no replacement real pod could schedule there. A node-schedulability check now drops those placeholders, and the `karpenter.sh/do-not-disrupt` annotation is removed so Karpenter can reclaim their nodes.

Rolling to staging + one prod region first to validate before promoting the default.

# Notes
- Upstream PR: https://github.com/jeanschmidt/actions-runner-controller/pull/13 (author-flagged **high** risk).
- The new node-schedulability check (`CAPACITY_AWARE_ENABLE_NODE_SCHEDULABILITY_CHECK`, default **true**) needs node get/list/watch RBAC; without it the reporter self-degrades to core counting rather than wedging. Confirm the runner chart grants that RBAC on these clusters.
- Controller and runner charts must stay on the same minor version — a mismatch triggers ARS deletion.

### DIFF
--- a/osdc/clusters.yaml
+++ b/osdc/clusters.yaml
@@ -150,6 +150,7 @@ clusters:
       pdb_enabled: false
       pdb_min_available: 0
     arc:
+      chart_version: "0.14.1-jeanschmidt.18"
       replica_count: 1
       log_level: debug
       controller_cpu_request: "500m"
@@ -227,6 +228,7 @@ clusters:
       pdb_enabled: false
       pdb_min_available: 0
     arc:
+      chart_version: "0.14.1-jeanschmidt.18"
       replica_count: 1
       log_level: debug
       controller_cpu_request: "500m"
@@ -306,6 +308,7 @@ clusters:
       pdb_enabled: false
       pdb_min_available: 0
     arc:
+      chart_version: "0.14.1-jeanschmidt.18"
       replica_count: 1
       log_level: debug
       controller_cpu_request: "500m"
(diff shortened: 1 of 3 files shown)


huydhn (approved):

huydhn:
I see a note on the PR about a RBAC permission to do scheduability check that is legit. Are u doing it in a separate PR? My agent was independently review the ARC change, so I didn't notice your exact same note until now

jeanschmidt:
Ok, I'll add them, do a full code review again, a full round of tests and then decide to merge or not.

jeanschmidt:
I ran the full tests, it seems to be working. I also enabled smoke tests to catch regressions in RBAC permissions

PR #967, opened 2026-07-31
grafana: add macOS runners dashboard

huydhn:
Adds `grafana/macos.json`, a macOS equivalent of the DGX B200 dashboard: total runners, queue time, max simultaneous jobs, and retry-flakiness across the five `macos-*` labels.

Total runners comes from the new `misc.runner_fleet_count` table (pytorch/test-infra#8403), with expected counts drawn as dashed lines from pytorch-gha-infra Terraform — 111 m1/sonoma, 36 m2/sequoia, 8 m2/tahoe. `macos-m1-stable` / `macos-m2-stable` are aliases on the same hosts as `macos-m1-14` / `macos-m2-15`, so that panel merges them per fleet instead of summing.

### Preview

https://pytorchci.grafana.net/d/hu896vp/macos-runners?orgId=1&from=now-3M&to=now&timezone=browser&var-view=all&var-percentile=0.9


georgehong:
Nice, great to track runner counts here. Would the underlying data allow for different time granularities? I'm wondering if a line chart could be nice for this when we want to drill down to a specific mac runner type through the filtering toggles, and it might have support for an expected count/fleet-size there with thresholds? 

With the bar chart view, I wonder if similar expected runner populations could eventually become difficult to read.

huydhn:
> Nice, great to track runner counts here. Would the underlying data allow for different time granularities? I'm wondering if a line chart could be nice for this when we want to drill down to a specific mac runner type through the filtering toggles, and it might have support for an expected count/fleet-size there with thresholds?
> 
> With the bar chart view, I wonder if similar expected runner populations could eventually become difficult to read.

I copy the format of the exiting B200 dashboard, but yeah, we can experiment with line chart and different representation for the dashboard here. Agree that being able to set different time granularities is needed if we want faster response. The underlying data is refresh every 30 minutes

huydhn:
Good call on both — switched it in c6274af.

**Granularity:** yes, the underlying data is finer than the chart was showing. `misc.runner_fleet_count` is sampled every 30 min, and I was collapsing it with `toStartOfDay`. Now it buckets with `$__timeInterval(ts)`, so it follows the panel interval — daily-ish at 3 months, down to the raw 30-min samples when you zoom in or drill into one runner type.

**Line chart:** agreed, done. It's a gauge, not a daily count, so bars were the wrong shape anyway. Drilled down via `view` you now get two lines — actual and expected — for the selected type.

**Thresholds:** I tried that first and backed off. Grafana draws threshold lines from the panel's default field config, so a per-series `thresholds` override doesn't give you one line per fleet — and the expected values differ (111 m1/sonoma, 36 m2/sequoia, 8 m2/tahoe). A static panel threshold would work once you've filtered to a single type, but thresholds aren't variable-interpolated, so it can't follow the `view` toggle. Emitting expected as a dashed series keeps it correct in every view and shows up in CSV export too. Happy to revisit if you know a way to drive thresholds from a variable.

huydhn:
I can get this to a usable state and keep it in my playground in the meantime, let's wait for Jean in https://github.com/pytorch/test-infra/issues/8411

PR #970, opened 2026-07-31
grafana/osdc: dashboard quality-of-life tweaks

jeanschmidt:
- Hide meta-prod-aws-uw1 from both Week-over-Week % Change panels when all clusters are selected; it reappears for explicit subsets
- Set default time range to now-24h (was now-6h)
- Enable 1m auto-refresh (was disabled)

uw1's low traffic makes its week-over-week percentage too spiky to be useful in the aggregate view, so the queries add an `unless on(cluster)` guard that drops it only when the current selection equals the full cluster set. Descriptions updated to document the behavior.

dashboard: https://pytorchci.grafana.net/d/ci-infra-f4vfmq-osdc/osdc

### DIFF
--- a/grafana/osdc.json
+++ b/grafana/osdc.json
@@ -1436,6 +1436,131 @@
           }
         }
       }
+    },
+    "panel-12": {
+      "kind": "Panel",
+      "spec": {
+        "data": {
+          "kind": "QueryGroup",
+          "spec": {
+            "queries": [
+              {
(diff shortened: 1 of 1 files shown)


huydhn:
@claude Check if the change is sane

huydhn (approved):
LGTM! Greenlight bot candidate :)

PR #978, opened 2026-08-05
Don't restart multi-tenant runners that are processing a job

huydhn:
## Problem

The `Restart runners` step in `setup-host.yml` ran `docker stop $(docker ps -q)` under `become_user`, with two issues:

1. **Wrong docker daemon.** `become_user` runs a non-login shell that never sources `~/.bashrc`, so `DOCKER_HOST` is unset and the CLI hits the (empty) root socket instead of the tenant's rootless dockerd — so it actually found and stopped nothing.
2. **Kills in-flight jobs.** It stopped runners unconditionally, so re-running the playbook could interrupt a job mid-execution.

## Fix

- Export `DOCKER_HOST`/`XDG_RUNTIME_DIR`/`PATH` for the per-tenant rootless daemon.
- Scope to the `ghad-main-shared-instance-container` and check `docker top` for a `Runner.Worker` child — that process exists only while a job is running. If present, leave it alone; `ghad-manager` respawns the runner (with current config) after the ephemeral runner finishes and exits.

Host-side signal only — no GitHub token or `docker exec` needed.

### DIFF
--- a/multi-tenant/playbooks/setup-host.yml
+++ b/multi-tenant/playbooks/setup-host.yml
@@ -377,11 +377,25 @@
       become: true
       become_user: "{{ item.name }}"
       ansible.builtin.shell: |
-        containers=$(docker ps -q)
-        if [ -n "$containers" ]; then
-          docker stop $containers
+        set -o pipefail
+        uid="$(id -u)"
+        # become_user has no ~/.bashrc, so set the rootless docker env ourselves.
+        export DOCKER_HOST="unix:///run/user/${uid}/docker.sock"
+        export XDG_RUNTIME_DIR="/run/user/${uid}"
+        export PATH="$HOME/bin:$PATH"
+        [ -S "/run/user/${uid}/docker.sock" ] || { echo "no docker socket"; exit 0; }
+        c=$(docker ps -q -f "name=ghad-main-shared-instance-container")
+        [ -n "$c" ] || { echo "no runner container"; exit 0; }
+        # Leave a running job alone (Runner.Worker present); ghad-manager respawns it after.
+        if docker top "$c" 2>/dev/null | grep -q "Runner.Worker"; then
+          echo "skip: job in progress"
+        else
+          echo "stopping idle runner"
+          docker stop "$c"
         fi
       loop: "{{ created_users.results }}"
+      loop_control:
+        label: "{{ item.name }}"
 
     - name: Create a HF cache directory
       ansible.builtin.file:
--- a/multi-tenant/playbooks/setup-host.yml
+++ b/multi-tenant/playbooks/setup-host.yml
@@ -377,21 +377,23 @@
       become: true
       become_user: "{{ item.name }}"
       ansible.builtin.shell: |
-        set -o pipefail
+        set -euo pipefail
         uid="$(id -u)"
+        NAME="ghad-main-shared-instance-container"
         # become_user has no ~/.bashrc, so set the rootless docker env ourselves.
         export DOCKER_HOST="unix:///run/user/${uid}/docker.sock"
         export XDG_RUNTIME_DIR="/run/user/${uid}"
         export PATH="$HOME/bin:$PATH"
         [ -S "/run/user/${uid}/docker.sock" ] || { echo "no docker socket"; exit 0; }
-        c=$(docker ps -q -f "name=ghad-main-shared-instance-container")
-        [ -n "$c" ] || { echo "no runner container"; exit 0; }
+        # Exact-name lookup via docker inspect (docker ps --filter name= is a substring match).
+        docker inspect -f '{% raw %}{{.State.Running}}{% endraw %}' "$NAME" 2>/dev/null | grep -qx true \
+          || { echo "no running runner container"; exit 0; }
         # Leave a running job alone (Runner.Worker present); ghad-manager respawns it after.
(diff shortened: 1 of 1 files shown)


huydhn:
cc @nWEIdia

multi-tenant/playbooks/setup-host.yml
```diff
@@ -377,11 +377,25 @@
...
       become_user: "{{ item.name }}"
       ansible.builtin.shell: |
-        containers=$(docker ps -q)
-        if [ -n "$containers" ]; then
-          docker stop $containers
+        set -o pipefail
+        uid="$(id -u)"
+        # become_user has no ~/.bashrc, so set the rootless docker env ourselves.
+        export DOCKER_HOST="unix:///run/user/${uid}/docker.sock"
+        export XDG_RUNTIME_DIR="/run/user/${uid}"
```
nWEIdia, line 385: Just for my knowledge, what does XDG stand for and what is XDG's role?

nWEIdia (commented):
Pasting claude review fyi: 

Overview

This PR fixes two real bugs in the "Restart runners" step:
1. become_user doesn't source ~/.bashrc, so DOCKER_HOST was unset and all docker commands were silently talking to the root socket (which is empty), making the old stop-all-containers logic a no-op.
2. The old unconditional docker stop would kill in-flight jobs if a runner happened to be processing one.

The fix explicitly sets the rootless Docker env vars, targets only the known runner container by name, and gates the docker stop on whether a Runner.Worker child process is present.

---
Positives

- The guard-clause pattern (exit 0 early returns) is clean and keeps the happy path readable.
- loop_control.label is a nice usability improvement for Ansible output.
- Scoping the stop to the named container is strictly safer than the old "stop everything" approach.
- Suppressing docker top stderr with 2>/dev/null prevents noisy failures from polluting output.

---
Issues & Suggestions

1. Shell is not explicitly set to bash — pipefail may silently do nothing

set -o pipefail is a bashism not guaranteed in /bin/sh. ansible.builtin.shell defaults to sh. Add args: executable: /bin/bash to be safe:

ansible.builtin.shell: |
  set -o pipefail
  ...
args:
  executable: /bin/bash

2. Container name is hardcoded with a partial-match filter

docker ps -f "name=ghad-main-shared-instance-container" performs a substring/prefix match, not an exact match. A container named ghad-main-shared-instance-container-2 would also match. Use an anchored regex if Docker supports it on the target version, or use docker inspect for an exact match:

c=$(docker ps -q -f "name=^ghad-main-shared-instance-container$")

More broadly, hardcoding the container name here is a maintenance hazard. Consider making it a variable (e.g., {{ runner_container_name | default('ghad-main-shared-instance-container') }}).

3. 3. TOCTOU race between docker top check and docker stop

The Runner.Worker check and the subsequent docker stop are not atomic. A job could start in the window between the two commands. The PR description acknowledges that ghad-manager will respawn idle runners, but a job starting mid-restart would still be killed. This is a low-probability race, but worth a comment acknowledging it.

4. Scope change: only the runner container is stopped, not all containers

The old code stopped all containers (docker ps -q). The new code only stops ghad-main-shared-instance-container. If other containers exist alongside the runner that also need restarting on host setup, they are now silently skipped. This is probably intentional (since the old code was broken and stopped nothing due to the wrong socket), but it deserves a comment or explicit confirmation in the PR that no other containers need to be restarted.

5. No set -e (errexit)

The script uses pipefail but not errexit. If intermediate commands like id -u or the docker top pipeline fail unexpectedly, the script will continue executing. The guard clauses handle the expected paths, but an unexpected mid-script failure could lead to silent partial execution. Consider set -euo pipefail for stricter failure handling.

6. Minor: $HOME/bin prepended to PATH without a guard

export PATH="$HOME/bin:$PATH" will silently prepend even if $HOME/bin doesn't exist. This is harmless but slightly untidy. Not a blocker.

Summary

The fix is correct and addresses genuine bugs. The two most actionable changes before merging are: (1) adding args: executable: /bin/bash to ensure pipefail actually applies, and (2) using an exact container name match to avoid false positives from the docker ps filter. The TOCTOU race and hardcoded name are worth a comment at minimum.

multi-tenant/playbooks/setup-host.yml
huydhn: That's just a fancy name for `X Desktop Group`. Any folder can work here I think I just need to point it to one with `docker.sock` , that's the rootless Docker socket.

atalman (approved):

nWEIdia (approved):

huydhn:
Thanks for the review! Pushed the two actionable ones:

- **Exact name match** — now `docker inspect "$NAME"` (exact) instead of `docker ps --filter name=` (substring); name is a variable used throughout.
- **errexit** — `set -o pipefail` → `set -euo pipefail`, so a real `docker stop` failure surfaces (the `|| { …; exit 0; }` guards keep the no-socket / not-running paths clean).

On the others, intentionally leaving as-is:

- **bash / pipefail** — already guaranteed by the play's `module_defaults: shell: executable: /bin/bash`.
- **TOCTOU** — unavoidable for any check-then-stop (the GitHub-API approach races too); sub-second window in a manual provisioning run.
- **Only stops the runner container** — deliberate: not touching a live job's sibling containers is the point, and `ghad-manager`'s `stop_all_containers` + prune cleans leftovers on the next respawn.
- **`$HOME/bin` on PATH** — harmless even if the dir is absent.

PR #986, opened 2026-08-11
Add nodepools-agent-sandbox module: a gVisor node fleet

huydhn:
Thin shim over `modules/nodepools` (mirrors `nodepools-h100`/`-b200`) generating a dedicated `ai-sandbox` Karpenter fleet whose nodes install gVisor (runsc) as a containerd runtime handler via userData. Being its own module with its own `defs/`, the fleet only exists on clusters that list it — it can't land on prod through the shared nodepools defs.

Nothing schedules here until `agent-sandbox` adds the `gvisor` RuntimeClass (PR 2 of this stack), and the gVisor install is fail-closed: if it fails, only `runtimeClassName: gvisor` pods fail to start. No cluster enables this module yet.

**Stack** (review bottom-up, 1/5 — this PR):
1. **nodepools-agent-sandbox: gVisor fleet** ← this PR
2. agent-sandbox module
3. canary integration test
4. enable on ue1 staging
5. build the agent image in deploy.sh

`just lint` 12/13 (the one failure, hadolint on `runner-base`, is pre-existing on main), `just test` green.

PR #987, opened 2026-08-11
Add agent-sandbox module: an AI agent sandbox callable like BuildKit

huydhn:
Phase-1 PoC of the [AI Agent Sandbox design](https://docs.google.com/document/d/1Iyp05vtBULwQ2RmPD60n6-lgzT9l_nmE4XVWcEI0lxk/edit). `sandbox-agent` is a long-running HTTP worker on the gVisor fleet that clones a public repo and asks Bedrock about it. A runner calls it over the network exactly like buildkitd — `curl sandbox-agent.ai-sandbox.svc:8080/run` — gated by a NetworkPolicy allowing the `arc-runners` namespace, so **the caller needs no RBAC**.

The agent holds **exactly one credential**: a read-only Bedrock IRSA role. It clones public GitHub anonymously, has no K8s RBAC and no token automount, and IMDS hop-limit 1 keeps the node role out of reach.

Bedrock must be a `us.` inference profile — Anthropic models in us-east-1 are `INFERENCE_PROFILE`-only and the one still advertising `ON_DEMAND` (Claude 3 Haiku) is refused as provider-legacy — so the IAM policy grants `InvokeModel` on `foundation-model/*` across regions, which is what cross-region routing authorizes against.

Prototype limits are listed in the module README: egress is **open**, the worker is warm rather than per-task-ephemeral, and output is trusted as-is. No cluster enables this module yet.

**Stack** (review bottom-up, 2/5 — this PR):
1. nodepools-agent-sandbox: gVisor fleet (#986)
2. **agent-sandbox module** ← this PR
3. canary integration test
4. enable on ue1 staging
5. build the agent image in deploy.sh

`server.py` has 100% unit coverage (real git clone via `insteadOf`, real HTTP against a socket); smoke tests cover the isolation and credential assertions. `just lint` 12/13 (hadolint on `runner-base` is pre-existing on main).

### DIFF
--- a/osdc/modules/agent-sandbox/README.md
+++ b/osdc/modules/agent-sandbox/README.md
@@ -1,0 +1,141 @@
+# agent-sandbox (PROTOTYPE)
+
+A sandbox for running an untrusted AI agent in OSDC CI. The agent clones a
+**public** repo, calls **Bedrock**, and returns a result — isolated under gVisor,
+holding exactly one credential (a **read-only Bedrock IRSA role**) and nothing else.
+
+It is **callable over the network like BuildKit**: a runner does
+`curl sandbox-agent.ai-sandbox.svc:8080/run …`, gated by a NetworkPolicy that
+allows the `arc-runners` namespace — **no RBAC on the caller**.
+
+This is a **Phase-1 PoC**. There are **no credential proxies** and **no operator
+secrets**: the agent clones public GitHub anonymously and calls Bedrock directly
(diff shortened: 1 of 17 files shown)


huydhn:
Superseded by the ghstack version of the same stack: #991 → #992 → #993 → #994 → #995. Identical content; resubmitted with ghstack to match how stacks are normally landed in this repo (and so restacking after review feedback is one command).

PR #989, opened 2026-08-11
Enable the agent sandbox on meta-staging-aws-ue1

huydhn:
Turns on `nodepools-agent-sandbox` (gVisor fleet) and `agent-sandbox` on ue1 staging only, which also activates the `AGENT_SANDBOX` canary job, and sets the Bedrock model — a `us.` inference profile, since Anthropic models in us-east-1 are `INFERENCE_PROFILE`-only.

Staging only; no prod cluster lists these modules. The fleet scales from zero, so the idle cost is one warm sandbox pod.

Verified on `meta-staging-aws-ue1`: `/run` returns `cloned=true`, 21952 files, 74 top-level entries, `errors={}`, and a grounded Bedrock report (asked which top-level file pins the Python build requirements → `requirements-build.txt`).

**Stack** (review bottom-up, 4/5 — this PR):
1. nodepools-agent-sandbox: gVisor fleet (#986)
2. agent-sandbox module (#987)
3. canary integration test (#988)
4. **enable on ue1 staging** ← this PR
5. build the agent image in deploy.sh

PR #988, opened 2026-08-11
Add the agent-sandbox canary integration test

huydhn:
`test-agent-sandbox`, gated by the `AGENT_SANDBOX` tag (requires `arc-runners` + `agent-sandbox`), calls the sandbox from a normal runner exactly like a runner calls buildkitd, and asserts all three properties: the Service is reachable from `arc-runners` with no RBAC, the repo clones with no token anywhere, and Bedrock really answers.

Bedrock is a **hard** assertion — `errors.bedrock`, an empty report, or an ungrounded prompt (empty `top_level`) all fail the job. It used to be a `WARN`, which let a green run hide the fact that Bedrock was never actually invoked.

Inert until a cluster enables the modules (PR 4 of this stack).

**Stack** (review bottom-up, 3/5 — this PR):
1. nodepools-agent-sandbox: gVisor fleet (#986)
2. agent-sandbox module (#987)
3. **canary integration test** ← this PR
4. enable on ue1 staging
5. build the agent image in deploy.sh

PR #990, opened 2026-08-11
Build the agent image in deploy.sh instead of by hand

huydhn:
The module was undeployable from a clean checkout: the operator had to run `buildx` + `crane` by hand before `just deploy-module` would work, against a mutable `:prototype` tag that needed `imagePullPolicy: Always` to roll at all.

`deploy.sh` now builds and pushes the image to the in-cluster Harbor `osdc` project under a content-hash tag, mirroring `modules/zombie-cleanup` — an unchanged tree skips the build, a code change deploys a new immutable tag, and the default `IfNotPresent` pull policy becomes correct. Drops `agent_sandbox.agent_image` from `clusters.yaml`. Needs a local docker daemon, like the other image-building modules.

Verified on `meta-staging-aws-ue1`: first deploy built and pushed `ci-agent-sandbox:e78c54d7f13e`, second skipped the build, and `/run` still returns a grounded Bedrock report.

**Stack** (review bottom-up, 5/5 — this PR):
1. nodepools-agent-sandbox: gVisor fleet (#986)
2. agent-sandbox module (#987)
3. canary integration test (#988)
4. enable on ue1 staging (#989)
5. **build the agent image in deploy.sh** ← this PR

### DIFF
--- a/osdc/clusters.yaml
+++ b/osdc/clusters.yaml
@@ -263,9 +263,6 @@ clusters:
     pypi_cache:
       replicas: 1
     agent_sandbox:
-      # PROTOTYPE. Built + pushed to the in-cluster Harbor `osdc` project; nodes
-      # pull it via the harbor:30002 mirror (see modules/agent-sandbox/README.md).
-      agent_image: "harbor:30002/osdc/ci-agent-sandbox:prototype"
       # Bedrock model the agent invokes. Must be a cross-region inference profile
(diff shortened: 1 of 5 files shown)


huydhn:
Superseded by the ghstack version of the same stack: #991 → #992 → #993 → #994 → #995. Identical content; resubmitted with ghstack to match how stacks are normally landed in this repo (and so restacking after review feedback is one command).

PR #991, opened 2026-08-11
Add nodepools-agent-sandbox module: a gVisor node fleet

huydhn:
* #994
* #993
* #992
* #997
* #996
* __->__ #991
* #1005

Thin shim over modules/nodepools (mirrors nodepools-h100/-b200) generating a
dedicated `ai-sandbox` Karpenter fleet whose nodes have gVisor (runsc) available
as a containerd runtime handler. Being its own module with its own defs/, the
fleet only exists on clusters that list it — it can never land on prod through
the shared nodepools defs.

Nodes boot from the custom AMI built by the previous commit, selected by tag
rather than name glob so the AMI naming scheme can change without touching the
def. That needs a new per-def `ami_selector_tags` knob in the nodepools
generator; the generated EC2NodeClass then carries no userData script at all,
just Karpenter's own nodeadm NodeConfig.

Trade-off worth knowing: `alias: al2023@latest` picks up AL2023 CVE fixes on node
rotation, and a pinned AMI only moves when rebuilt. Each rebuild re-resolves the
base from the EKS SSM parameter, so rebuilding is the whole refresh procedure —
the README documents the cadence.

c7i.2xlarge is sized against the sandbox pod that lands on it (2 vCPU / 4 GiB,
requests == limits) so the node holds exactly 3, with both dimensions running out
together. The arithmetic is in the def, using allocatable and daemonset requests
measured on a live node. Adds c7i.2xlarge to instance_specs.py, memory_mi
collected from a real node.

Verified on meta-staging-aws-ue1: a node recycled onto the gVisor AMI registers
runsc release-20260803.0 in the generated containerd config with no restart and
no userData script, and the sandbox pod reports "Starting gVisor..." on a
4.19.0-gvisor kernel.


osdc/modules/nodepools-agent-sandbox/scripts/install-gvisor.sh
```diff
@@ -0,0 +1,44 @@
...
+CONFIG=/etc/containerd/config.toml
+if ! grep -q 'containerd.runtimes.runsc' "${CONFIG}"; then
+  cat >>"${CONFIG}" <<'EOF'
+
+[plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runsc]
+  runtime_type = "io.containerd.runsc.v1"
+EOF
+fi
+
+systemctl restart containerd
```
jeanschmidt: No, lets NOT restart containerd on user-data.sh if we can avoid.

osdc/modules/nodepools-agent-sandbox/defs/ai-sandbox.yaml
```diff
@@ -0,0 +1,33 @@
...
+fleet:
+  name: ai-sandbox
+  arch: amd64
+  gpu: false
+  instances:
+    - type: c7i.2xlarge
+      weight: 100
+      node_disk_size: 100
+      # gVisor (runsc) install runs as a userData MIME part on these nodes only.
+      user_data_script: scripts/install-gvisor.sh
```
jeanschmidt: lets reconsider this approach, I rather not have a user_data_script on a cluster managed by karpenter. In this case the correct thing would be to have a custom AMI.

jeanschmidt (changes requested):

osdc/modules/nodepools-agent-sandbox/defs/ai-sandbox.yaml
huydhn: True, let's go with a custom AMI, the default AL2023 images doesn't have this installed

huydhn: Agreed, done — gVisor now comes from a custom AMI built by `packer/`, and the fleet selects it by tag. The generated EC2NodeClass carries no script at all, just Karpenter's own nodeadm NodeConfig.

Worth recording why the userData version was worse than it looked: it also fetched the gVisor release over the network on *every* scale-up and tracked `latest`, so two nodes launched a day apart could run different runsc builds. The AMI pins it.

Trade-off this introduces: `alias: al2023@latest` picks up AL2023 CVE fixes on node rotation, and a pinned AMI only moves when rebuilt. Each rebuild re-resolves the base from the EKS SSM parameter, so `just build-agent-sandbox-ami <cluster>` is the whole refresh procedure — documented in the module README, and worth a scheduled rebuild if this fleet outlives the prototype.

osdc/modules/nodepools-agent-sandbox/scripts/install-gvisor.sh
huydhn: Gone — containerd is never restarted now.

The reason it was there: nodeadm regenerates `/etc/containerd/config.toml` at every boot, so appending the runsc stanza afterwards only took effect with a restart — which, as you point out, happens *after* kubelet has started and races anything already scheduled on the node.

The AMI instead ships a nodeadm NodeConfig at `/etc/eks/nodeadm.d/10-gvisor.yaml`. nodeadm merges that directory with the user-data config and writes the containerd config **before** starting containerd, so the runsc handler is registered at first start:

```yaml
spec:
  containerd:
    config: |
      [plugins.'io.containerd.cri.v1.runtime'.containerd.runtimes.runsc]
        runtime_type = "io.containerd.runsc.v1"
```

Verified on a recycled `meta-staging-aws-ue1` node: `runsc release-20260803.0` present, the stanza in the generated `/etc/containerd/config.toml`, no restart, and the sandbox pod reports `Starting gVisor...` on a `4.19.0-gvisor` kernel. There's also a smoke test asserting the userData contains neither `runsc` nor `systemctl restart containerd`, so this can't quietly come back.

jeanschmidt (approved):

osdc/modules/nodepools-agent-sandbox/defs/ai-sandbox.yaml
```diff
@@ -0,0 +1,46 @@
...
+  arch: amd64
+  gpu: false
+  instances:
+    - type: c7a.2xlarge
+      weight: 100
+      node_disk_size: 100
+      # Select the gVisor AMI built by packer/, by tag rather than name glob so
+      # the naming scheme can change without touching this def.
+      ami_selector_tags:
+        osdc.io/ami: ai-sandbox-gvisor
```
izaitsevfb, line 54: is this intended? do we want to select by K8sVersion?

if not for PoC, let's explicitly document this as a TODO

---

<details>
<summary>🟡 The AMI is selected by one tag and nothing else, so the fleet ignores the K8sVersion the build recorded — after an EKS bump it keeps launching pre-upgrade nodes until someone rebuilds the AMI, silently. (ai-generated section)</summary>

`amiSelectorTerms: [{tags: {osdc.io/ami: ai-sandbox-gvisor}}]` matches every self-owned AMI in the region carrying that tag. The packer build does record `K8sVersion` (and `Cluster`), but the selector reads neither. The only other CPU branch in the generator uses `alias: al2023@latest`, which resolves against the cluster's own version and therefore cannot have this failure mode. Nothing else catches it either: the new smoke test only asserts that SOME AMI with the tag exists. Latent rather than live today — clusters.yaml carries a single `eks_version: "1.35"` default for every cluster. Cheapest fix that fits in this PR: have test_ami_exists_in_region assert the matching AMI's K8sVersion tag equals resolve_config('eks_version'). Robust fix, and it fails closed: put K8sVersion into the selector, which means plumbing the cluster's eks_version into the generator.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 23af44ed7.</sub>

</details>

osdc/modules/nodepools/scripts/python/generate_nodepools.py
```diff
@@ -302,6 +302,23 @@ def generate_nodepool_yaml(nodepool_def, module_name, defs_dir=None):
         gpu_taints = ""
         gpu_tags = ""
 
+    # ----- Custom AMI override -----
```
izaitsevfb, line 305: https://github.com/pytorch/ci-infra/blob/23af44ed71933469f28255b3d76ead12f1f6fb7b/osdc/modules/nodepools/scripts/python/generate_nodepools.py#L260-L264

https://github.com/pytorch/ci-infra/blob/23af44ed71933469f28255b3d76ead12f1f6fb7b/osdc/clusters.yaml#L27-L35

https://github.com/pytorch/ci-infra/blob/23af44ed71933469f28255b3d76ead12f1f6fb7b/osdc/modules/nodepools/scripts/python/generate_nodepools.py#L93-L105

this appears important.

at least worth clarifying the TODO

---

<details>
<summary>⚪ The in-tree plan for retiring two live kernel-CVE mitigations rests on every node picking the patched AL2023 image up by itself on rotation. This fleet's AMI is pinned by tag, so it will not, and this PR adds the first selector in the generator that tracks neither the alias nor a name glob. (ai-generated section)</summary>

The TODO above the is_gpu branch in generate_nodepools.py says "the AL2023 aliases / name globs below already track @latest, so node rotation picks up the fix automatically", and that base/kubernetes/dirtyfrag-mitigation.yaml should be removed "once rolled out across all nodes". clusters.yaml carries the same pair of TODOs against base_node_ami_version, which governs only the terraform base node group. That function has exactly three amiSelectorTerms branches: the nvidia name glob, alias: al2023@latest, and the ami_selector_tags override added here about fifty lines under the TODO — so this is the first that tracks neither, and the module README says as much under "Rebuild cadence". Nothing removes a mitigation automatically; the cost is that the retirement criterion can read as met while this fleet is behind, because for these nodes a patched kernel arrives only after the AMI is rebuilt and existing nodes are recycled, which is two deliberate steps rather than rotation. Removing the DaemonSet stops new nodes getting the esp4/esp6/rxrpc modprobe blacklist, and its STARTUP_TAINTS entry has to go first or at the same time, never after — dropping the DaemonSet while the registry still emits the taint strands every new node, which is the ordering the registry's own comment warns about. Cheapest fix: a line on each of the two TODOs noting that a fleet with ami_selector_tags moves only when its AMI is rebuilt and its nodes recycled. Stronger, with a precedent in this repo: a guard test shaped like the bin-pack-scheduler one that pins the scheduler image minor to eks_version, asserting the tagged AMI's recorded SourceAMI is the one the EKS AL2023 SSM parameter resolves to today, or is in an approved set — an AMI id has no ordering, and K8sVersion would not show the kernel is patched either. Worth confirming rather than assuming: whether a sandboxed process can reach the host xfrm path at all under runsc — if it cannot, what is left exposed here is the cluster daemonsets the def's own sizing arithmetic already measures on these nodes.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 23af44ed7.</sub>

</details>

osdc/modules/nodepools/scripts/python/generate_nodepools.py
```diff
@@ -302,6 +302,23 @@ def generate_nodepool_yaml(nodepool_def, module_name, defs_dir=None):
...
+    # Trade-off worth knowing: `alias: al2023@latest` picks up AL2023 CVE fixes
+    # automatically on node rotation; a custom AMI only moves when it is rebuilt.
+    ami_selector_tags = nodepool_def.get("ami_selector_tags") or {}
+    if ami_selector_tags:
+        tag_lines = "\n".join(f'        {k}: "{v}"' for k, v in sorted(ami_selector_tags.items()))
+        ami_family_block = "  amiFamily: AL2023"
+        ami_selector_block = f"""  amiSelectorTerms:
+    - tags:
+{tag_lines}"""
+
```
izaitsevfb, line 323: worth adding a test

---

<details>
<summary>⚪ The new generator knob has no unit test, in a file whose 1814-line test suite already has the two sibling tests to copy — and the only coverage it does have needs a live cluster. (ai-generated section)</summary>

modules/nodepools/scripts/python/test_generate_nodepools.py contains test_gpu_ec2nodeclass_ami_family and test_cpu_ec2nodeclass_no_ami_family, which is exactly this shape, and CI runs that suite on every PR (the `test` check is green here). The module's own tests are pytest.mark.live, so nothing in PR CI asserts the YAML this branch renders: an accidental change that still parses would ship. Worth asserting: the tags block renders and sorts deterministically, `amiFamily: AL2023` is emitted, no `alias:` term survives, and the key passes through _build_fleet_nodepool_def.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 23af44ed7.</sub>

</details>

izaitsevfb (approved):
lgtm, with a couple of nits.

please take a look at the K8sVersion selector one

osdc/modules/nodepools-agent-sandbox/defs/ai-sandbox.yaml
huydhn: I can add a TODO here, my plan is to have a workflow to automate the build of this custom AMI eventually

osdc/modules/nodepools/scripts/python/generate_nodepools.py
huydhn: I check, we are good here:
* These patches have been rolled out to the base AL2023 AMI, I think we can clean them up now. Let me follow up on this in another PR
* Having a workflow to rebuild the AMI periodically would help keep it up to date

osdc/modules/nodepools/scripts/python/generate_nodepools.py
huydhn: Done

PR #992, opened 2026-08-11
Add agent-sandbox module: a credential-free sandbox, one gVisor pod per request

huydhn:
* #994
* #993
* __->__ #992
* #997
* #996
* #991
* #1005

PoC of the AI Agent Sandbox design:
https://docs.google.com/document/d/1Iyp05vtBULwQ2RmPD60n6-lgzT9l_nmE4XVWcEI0lxk/edit

A runner calls the sandbox over the network exactly like buildkitd — `curl
sandbox-agent.ai-sandbox.svc:8080/run` — gated by a NetworkPolicy allowing the
arc-runners namespace, so the caller needs no RBAC.

Each request becomes one Kubernetes Job rather than a slot on a shared worker. The
dispatcher receives /run, creates a Job running the task image under gVisor, waits for
it and reads the result out of the pod's log. So requests run in parallel, capacity
comes from Karpenter — a pending task pod adds an ai-sandbox node — and every task
gets a fresh pod, which is an isolation reset a long-lived worker cannot give.
Concurrency is bounded by MAX_CONCURRENT_TASKS per dispatcher replica and by a
namespace ResourceQuota (12 slots); past that a caller gets 429. Both bounds matter
because /run is unauthenticated, so an unbounded caller would otherwise turn a loop
into nodes until an AWS quota noticed.

The sandbox holds no credentials of any kind. Task pods run under runtimeClassName
gvisor on the sandbox fleet with no K8s RBAC and no token automount, and Bedrock is
reached through sigv4-proxy: the task sends an unsigned request and the proxy signs it
with the IRSA role bound to its own SA, pinned to bedrock-runtime with --host/--name so
a caller cannot choose which service gets signed for. The dispatcher is the trusted
half — it can create Jobs in this namespace and nothing else, runs on base nodes away
from the fleet, and never clones or prompts a model. deploy.sh still strips any IRSA
annotation left on the sandbox-agent SA, since apply cannot remove an annotation it no
longer sets and an upgraded cluster would otherwise hand every task pod a credential.

The proxy is treated as load-bearing infrastructure, since a task has no retry: a
readiness probe (it is otherwise Ready ~4s before it can serve), maxUnavailable 0, a
preStop drain (maxUnavailable 0 does nothing for connections already on the terminating
pod), 2 replicas one-per-node with generation-scoped anti-affinity, and PDBs. It runs
without --verbose, which would log the live X-Amz-Security-Token into pod logs that
alloy-logging ships to Loki.

A task slot is 2 vCPU / 4 GiB / 20 GiB with requests == limits, so per-node capacity is
a division rather than a guess (3 per fleet node — arithmetic in the README), one
untrusted task cannot burst into another's CPU, and a caller-named repo cannot fill the
node's disk.

Egress caveat, stated in the README: under IPv6-only VPC-CNI the NetworkPolicy does not
cover IPv4, so it is defense-in-depth, not containment. What holds regardless is that
the sandbox has no credential to exfiltrate.

No cluster enables this module yet.


osdc/modules/agent-sandbox/kubernetes/base/sandbox-agent.yaml
```diff
@@ -0,0 +1,95 @@
...
+            # AWS_WEB_IDENTITY_TOKEN_FILE automatically (from the SA annotation).
+            - name: AWS_REGION
+              value: "__AWS_REGION__"
+            # Default model for /run — a caller can override it per request with
+            # "model". A cross-region inference profile ID; see clusters.yaml.
+            - name: BEDROCK_DEFAULT_MODEL_ID
+              value: "__BEDROCK_DEFAULT_MODEL_ID__"
+          readinessProbe:
+            httpGet:
+              path: /healthz
```
jeanschmidt: :)

jeanschmidt (approved):

osdc/modules/agent-sandbox/kubernetes/base/sandbox-agent.yaml
huydhn: I wonder where the agent gets this convention from

osdc/modules/agent-sandbox/kubernetes/base/networkpolicy.yaml
```diff
@@ -0,0 +1,143 @@
...
+          port: 8080
+    - to:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: kube-system
+      ports:
+        - protocol: UDP
+          port: 53
+        - protocol: TCP
+          port: 53
```
izaitsevfb, line 91: have you tested the new version? is github reachable?

---

<details>
<summary>🔴 The worker's new egress rule contains no path to github.com, and cloning from github.com is the first thing every task does. (ai-generated section)</summary>

default-deny covers Egress for the whole namespace, and sandbox-agent-egress then permits the worker exactly two destinations: the sigv4-proxy pods on 8080, and kube-system on 53. The worker clones with git straight to https://github.com from inside the pod; nothing routes that through the proxy, and the proxy signs AWS requests only. So the clone is not expressed anywhere in the allow-list, and whether it works depends on the enforcement gap this PR documents itself — under IPv6-only VPC-CNI a NetworkPolicy does not cover the IPv4 egress path. Worth confirming which is intended, because both readings need a change: if the policy is enforced on the path the clone takes, every task fails at step one; if it is not, the module works only because the policy it ships is ineffective, and the dedicated no-NAT subnet the README names as the Phase-2 boundary will break cloning when it lands. Either way the comment heading that policy — "Agent egress: ONLY the two proxies + DNS. No direct internet path." — does not describe what the pod can do. Smallest fix: add an explicit egress rule for the clone path and say what it covers, or put git behind a proxy the way Bedrock now is.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f9cc83d7a.</sub>

</details>

osdc/modules/agent-sandbox/kubernetes/base/sigv4-proxy.yaml
```diff
@@ -0,0 +1,135 @@
...
+            # No --name: the target service/host comes from each request's Host
+            # header, so one proxy serves bedrock-runtime, cloudwatch, logs, etc.
+            #
+            # NOT --verbose: it dumps every signed request, including the live
+            # X-Amz-Security-Token, into pod logs — which alloy-logging then ships
+            # to Loki. That would put the credential we just took away from the
+            # agent into the logging pipeline instead. Failed requests only.
+            - --log-failed-requests
+            - --port
+            - ":8080"
```
izaitsevfb, line 106: wonder if we can tighten that (at least in the final version)

---

<details>
<summary>🔴 The proxy signs for whichever AWS service its caller names, and the role behind it grants account-wide CloudWatch and CloudWatch Logs read. (ai-generated section)</summary>

aws-sigv4-proxy runs without --name, so it takes the target service from each request's Host header; the comment above the args says that is deliberate, so one proxy can serve bedrock-runtime, cloudwatch and logs. Nothing on the way in restricts which of those a caller may ask for, which makes the IAM policy the entire boundary. That policy has a second statement beyond Bedrock: cloudwatch:GetMetricData, GetMetricStatistics and ListMetrics plus logs:GetLogEvents, FilterLogEvents, DescribeLogGroups and DescribeLogStreams, all on Resource "*" — so every log group in the account, not only ones related to the sandbox, and log output is a common place for tokens to end up. Reachability, stated precisely: today's worker only ever addresses bedrock-runtime, and /run exposes no way to pick a host, so nothing a caller sends reaches Logs. The exposure is what the sandbox can do once it runs anything more than a fixed prompt, which is the module's declared purpose and the threat the README's own "compromised agent" wording assumes. The proxy does refuse a Host it cannot resolve to an AWS service, so this does not extend to arbitrary internet hosts. Smallest fix: pin the upstream with --host and the signing service with --name plus --region so only Bedrock is reachable through the proxy, and drop the observability statement until something needs it.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f9cc83d7a.</sub>

</details>

osdc/modules/agent-sandbox/kubernetes/base/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
...
+kind: Kustomization
+
+resources:
+  - namespace.yaml
+  - runtimeclass.yaml
+  - serviceaccounts.yaml
+  - sigv4-proxy.yaml
+  - proxy-pdb.yaml
+  - sandbox-agent.yaml
+  - networkpolicy.yaml
```
izaitsevfb, line 23: please address for consistency

---

<details>
<summary>🔴 This revision removes the sandbox-agent Service, and every documented and tested way of reaching the sandbox still names it. (ai-generated section)</summary>

kustomization.yaml lists seven resources and none of them defines a Service for the worker; the only Service left in the module is sigv4-proxy's. Three things in this diff still assume the old one. deploy.sh's own summary says step 2 applies "the sandbox-agent worker + Service". Its final echo tells the operator to curl sandbox-agent.ai-sandbox.svc.cluster.local:8080. And test_worker_service_exists asserts exactly one Service named sandbox-agent in the namespace, so the smoke suite fails on any cluster this is deployed to. With no Service there is no ClusterIP and no cluster DNS record, so the worker is unreachable from arc-runners at all. Smallest fix: put the ClusterIP Service back on port 8080 targeting the http port. If the removal was deliberate, the callers and the smoke test need to change with it.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f9cc83d7a.</sub>

</details>

osdc/modules/agent-sandbox/deploy.sh
```diff
@@ -0,0 +1,100 @@
...
+  eks.amazonaws.com/role-arn="$SIGV4_ROLE_ARN" --overwrite
+kubectl rollout restart deployment/sigv4-proxy -n "$NAMESPACE"
+
+echo "[agent-sandbox] Waiting for rollouts..."
+kubectl rollout status deployment/sigv4-proxy -n "$NAMESPACE" --timeout=5m || true
+kubectl rollout status deployment/sandbox-agent -n "$NAMESPACE" --timeout=10m || true
+
+echo "[agent-sandbox] Deployed. The sandbox is callable from arc-runners like buildkitd:"
+echo "    curl -sf -X POST http://sandbox-agent.ai-sandbox.svc.cluster.local:8080/run \\"
+echo "      -d '{\"repo\":\"pytorch/pytorch\",\"ref\":\"main\",\"task\":\"...\"}'"
```
izaitsevfb, line 212: seems like a minor bug, worth addressing.

---

<details>
<summary>🟡 Both rollout waits end in "|| true", so the script prints "Deployed." and exits 0 whether or not either rollout finished. (ai-generated section)</summary>

They are the only commands in a set -euo pipefail script exempted from it, and they are the only commands that verify the deploy worked. Every sibling module with a Deployment — buildkit, monitoring, pypi-cache — calls rollout status bare and lets it fail the script. The step that matters most is the one just above them. On an upgrade from the earlier design the script strips the IRSA annotation off the worker's ServiceAccount and restarts it, and the restart is what actually drops the injected token from the running pod. Because the new pod has to become Ready before the old one goes away, a bad image or a crash-loop leaves the old credential-holding pod serving — and the operator is told the deploy succeeded. Smallest fix: drop || true from both.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f9cc83d7a.</sub>

</details>

osdc/modules/agent-sandbox/deploy.sh
```diff
@@ -0,0 +1,100 @@
...
+  kubectl annotate sa sandbox-agent -n "$NAMESPACE" eks.amazonaws.com/role-arn- >/dev/null
+  kubectl rollout restart deployment/sandbox-agent -n "$NAMESPACE"
+fi
+
+# --- Annotate the sigv4-proxy SA with its IRSA role, then restart it ---
+# The AWS credential is bound here, not to the agent: the pod-identity webhook
+# injects the web-identity token from this annotation.
+kubectl annotate sa sigv4-proxy -n "$NAMESPACE" \
+  eks.amazonaws.com/role-arn="$SIGV4_ROLE_ARN" --overwrite
+kubectl rollout restart deployment/sigv4-proxy -n "$NAMESPACE"
```
izaitsevfb: this looks like an important bug

---

<details>
<summary>🔴 The proxy's role is attached by an imperative annotate after apply, which forces an unconditional restart and leaves a first-deploy window with no credential. (ai-generated section)</summary>

kustomize applies the sigv4-proxy ServiceAccount without the role annotation; deploy.sh annotates it two steps later and then runs rollout restart. The restart is load-bearing because of that ordering rather than optional — but being unconditional it fires on every deploy, cycling both proxy replicas even when nothing changed, which partly defeats the kubectl_apply_if_changed upstream of it. And on a first deploy the proxy pods can be admitted and pass their tcpSocket readiness probe before the annotation exists, so the webhook injects no token and the proxy signs with nothing — while the worker has no retry, which is the argument the description itself makes for all the proxy availability work. Smallest fix: substitute the role ARN into serviceaccounts.yaml declaratively. deploy.sh already sed-substitutes three placeholders into the same kustomize output, so a fourth is free, and it removes both the window and the need to restart at all.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f9cc83d7a.</sub>

</details>

osdc/modules/agent-sandbox/kubernetes/base/sandbox-agent.yaml
```diff
@@ -0,0 +1,80 @@
...
+          # 12482 MiB are free for sandboxes (7.91 vCPU / 14624 MiB allocatable
+          # less the 0.92 vCPU / 2142 MiB the fleet's daemonsets request — both
+          # measured on a live node), and 7.00/2 = 3, 12482/4096 = 3.
+          # See modules/nodepools-agent-sandbox/defs/ai-sandbox.yaml.
+          resources:
+            requests:
+              cpu: "2"
+              memory: "4Gi"
+            limits:
+              cpu: "2"
```
izaitsevfb: an interesting catch. worth addressing in the final version.

---

<details>
<summary>🟡 Disk is the one resource dimension left uncapped, and it is the one the caller chooses. (ai-generated section)</summary>

The worker's resources block pins cpu and memory at requests == limits and sets no ephemeral-storage on either side. Each task clones a caller-named repo into a temporary directory on the container's writable layer, on a fleet node declared with node_disk_size: 100. A caller naming a large enough repo can therefore push the node into DiskPressure and get the pod evicted. The capacity table in the comment here and in the README is complete in vCPU and MiB and silent on disk, and arc-runners' own runner template sets ephemeral-storage on both requests and limits, so the convention exists in this repo. Rated below blocking because the fleet is dedicated and tainted, so no CI workload shares those nodes, and one replica runs. Smallest fix: add ephemeral-storage to requests and limits. Note a limit bounds the damage by evicting sooner; it does not prevent transient node pressure first.

Measured: Resource dimensions bounded by this spec.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f9cc83d7a.</sub>

</details>

osdc/modules/agent-sandbox/tests/smoke/test_agent_sandbox.py
```diff
@@ -0,0 +1,156 @@
...
+                f"sandbox-agent SA must NOT be able to {verb} {resource}; got {proc.stdout.strip()!r}."
+            )
+
+
+class TestAgentSandboxNetworkPolicies:
+    def test_default_deny_exists(self) -> None:
+        nps = run_kubectl(["get", "networkpolicies"], namespace=NAMESPACE)
+        names = {np["metadata"]["name"] for np in nps.get("items", [])}
+        assert "default-deny" in names, f"agent-sandbox must ship a default-deny NetworkPolicy; got {sorted(names)}."
+        assert len(names) >= 5, f"Expected the full NetworkPolicy set (>=5); got {sorted(names)}."
```
izaitsevfb, line 261: i agree with iz2 that this warrants a more comprehensive test coverage

---

<details>
<summary>🟡 The property this revision is built around — the worker can only talk to the proxy — is asserted only as a count of NetworkPolicy names. (ai-generated section)</summary>

The worker-egress half is covered only by test_default_deny_exists: default-deny is present, and there are at least five policies. Widening sandbox-agent-egress to 0.0.0.0/0, or loosening its podSelector, leaves all five names in place and every test green. Since the whole design rests on that allow-list, assert its contents: that sandbox-agent-egress permits the proxy pods and DNS and nothing else, and that whatever the clone needs is explicitly among them.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f9cc83d7a.</sub>

</details>

osdc/modules/agent-sandbox/kubernetes/base/networkpolicy.yaml
```diff
@@ -0,0 +1,143 @@
...
+  policyTypes:
+    - Ingress
+  ingress:
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: arc-runners
+      ports:
+        - protocol: TCP
+          port: 8080
```
izaitsevfb, line 49: curious if that's an intended limitation of the PoC?

---

<details>
<summary>⚪ One serial worker shared by every arc-runners pod means any caller can keep it unavailable to the others. (ai-generated section)</summary>

A second concurrent /run now gets an immediate 429 rather than queueing behind the listener, which is a real improvement — the refusal is visible instead of looking like a hang. What is still missing is any notion of who is calling. One request can hold the only worker for up to the clone timeout plus the invoke timeout, and nothing stops the same caller taking it again the moment it frees up, so a loop from any pod in arc-runners keeps every other consumer on 429. With one replica there is no second slot to fall back to. The README discloses the warm serial worker; the undisclosed half is that the caller set is unauthenticated and unbounded. Acceptable for a single-tenant prototype, worth naming before a second consumer exists. Smallest fix: say so in Limitations alongside the per-task reset gap, and treat a caller identity and per-request ephemerality as one Phase-2 item.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f9cc83d7a.</sub>

</details>

osdc/modules/agent-sandbox/tests/smoke/test_agent_sandbox.py
```diff
@@ -0,0 +1,156 @@
+"""Smoke tests for the agent-sandbox module.
+
+Validates the deployed standing infrastructure: namespace, the gvisor
+RuntimeClass, the two egress proxies (sigv4-proxy + sigv4-proxy), their Services,
+the IRSA-annotated sigv4-proxy SA, the credential-free sandbox-agent SA, and the
+NetworkPolicies. These check the security spine is wired — not that an agent run
+succeeds (that is the e2e test).
+"""
```
izaitsevfb, line 12: worth cleaning up

---

<details>
<summary>⚪ Several comments still describe the abandoned two-proxy design, two say the clone goes through a proxy, and one README bullet has lost its opening line. (ai-generated section)</summary>

The smoke suite's module docstring says it validates "the two egress proxies (sigv4-proxy + sigv4-proxy)" — the same name twice, where the second used to be the credential vault. The worker's egress policy is headed "ONLY the two proxies + DNS" although it permits one. Two places say the clone is proxied when it is not: the README overview claims the PoC proves "git-clone/Bedrock through proxies", and the Integration test section says the canary asserts "it clones a repo through the proxy". The Credential bullet has it right — public repos are cloned directly. Separately, the Limitations list has an orphaned fragment: "aws-sigv4-proxy:latest) — digest-pin before non-prototype use." sits under the propose/dispose bullet with no sentence of its own, so the image-pinning caveat has lost its opening clause and reads as part of the output-gate one.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f9cc83d7a.</sub>

</details>

osdc/modules/agent-sandbox/tests/smoke/test_agent_sandbox.py
```diff
@@ -0,0 +1,156 @@
...
+    def test_worker_ready(self, all_deployments: dict) -> None:
+        assert_deployment_ready(all_deployments, NAMESPACE, "sandbox-agent")
+
+    def test_worker_service_exists(self, all_services: dict) -> None:
+        svcs = filter_services(all_services, namespace=NAMESPACE, name="sandbox-agent")
+        assert len(svcs) == 1, f"Expected Service 'sandbox-agent' in '{NAMESPACE}'."
+
+    def test_worker_runs_under_gvisor(self, all_deployments: dict) -> None:
+        """The worker Deployment must request the gvisor RuntimeClass."""
+        dep = next(d for d in all_deployments["items"] if d["metadata"]["name"] == "sandbox-agent")
```
izaitsevfb: a minor test bug.

---

<details>
<summary>⚪ Three tests pick the Deployment by name across all namespaces, with no namespace filter and no default. (ai-generated section)</summary>

next(d for d in all_deployments["items"] if d["metadata"]["name"] == "sandbox-agent") appears three times, at lines 71, 84 and 98. all_deployments is cluster-wide, so a same-named Deployment in another namespace would be silently asserted against, and when the module is absent the bare next() raises StopIteration — still a failing test, but an error rather than the informative message the rest of the file is careful to provide. Smallest fix: a small helper that filters on namespace and name and asserts presence, in the shape filter_services is already used in with namespace=NAMESPACE.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f9cc83d7a.</sub>

</details>

izaitsevfb (dismissed):
there are too many nits, some of those look rather important. please do another pass before merging

osdc/modules/agent-sandbox/kubernetes/base/kustomization.yaml
huydhn: Oh darn, I should have landed this earlier, then iterate on this. Keeping this large stack open for a long time and unexpected change sneaks in

osdc/modules/agent-sandbox/kubernetes/base/sandbox-agent.yaml
huydhn: We haven't capped this anywhere else, but the suggestion looks reasonable. We can tweak / lift it later if needed

osdc/modules/agent-sandbox/kubernetes/base/sigv4-proxy.yaml
huydhn: Pin this to only Bedrock for now. We don't need anything else for the prototype

osdc/modules/agent-sandbox/deploy.sh
huydhn: Fixed

osdc/modules/agent-sandbox/kubernetes/base/networkpolicy.yaml
huydhn: No, this is not what I expected. I want the sandbox to scale up automatically when there are incoming requests. It looks like my agent interprets this as a TODO later, which is wrong

osdc/modules/agent-sandbox/kubernetes/base/kustomization.yaml
izaitsevfb: sorry! hopefully we can move faster now that I have this review pipeline

huydhn:
As testing looks ok now with multiple sandbox being able to run in parallel https://github.com/pytorch/pytorch-canary/actions/runs/32280319552/job/96157306388. I'm going to merge this now to unblock further iteration on the sandbox, which will surely come

PR #993, opened 2026-08-11
Add the agent-sandbox canary integration test

huydhn:
* #994
* __->__ #993
* #992
* #997
* #996
* #991
* #1005

`test-agent-sandbox`, gated by the AGENT_SANDBOX tag (requires arc-runners +
agent-sandbox), calls the sandbox from a normal runner exactly like a runner calls
buildkitd and asserts all three properties: the Service is reachable from
arc-runners with no RBAC, the repo clones with no token anywhere, and Bedrock
really answers.

Bedrock is a hard assertion — errors.bedrock, an empty report, or an ungrounded
prompt (empty top_level) all fail the job. Bedrock breakage was previously a WARN
that a green run hid.

Inert until a cluster enables the modules (next in the stack).


jeanschmidt (approved):

osdc/integration-tests/workflows/integration-test.yaml.tpl
```diff
@@ -1955,3 +1955,66 @@ jobs:
...
+  # anonymously with no token on the runner or in the agent, and (3) it really
+  # invokes Bedrock — through sigv4-proxy, which holds the AWS credential so the
+  # sandbox holds none at all.
+  test-agent-sandbox:
+    runs-on: { group: "{{RUNNER_GROUP}}", labels: ["{{PREFIX}}l-x86iamx-8-32"] }
+    container:
+      image: ghcr.io/actions/actions-runner:latest
+    env:
+      SANDBOX: http://sandbox-agent.ai-sandbox.svc.cluster.local:8080
+    steps:
```
izaitsevfb, line 1974: looks like blocking bug to me

---

<details>
<summary>🔴 The module no longer defines a `sandbox-agent` Service, so on a cluster deployed from this tree the health check fails on name resolution. (ai-generated section)</summary>

`kubernetes/base/sandbox-agent.yaml` carried a `Service` alongside the Deployment until the sigv4-proxy rework, which removed it. At this head the only `kind: Service` anywhere under `osdc/modules/agent-sandbox/` is `sigv4-proxy`, and no later PR in the stack adds one back — an already-deployed cluster could still hold an orphaned copy, since a kustomize apply does not prune, but a fresh namespace has nothing for `sandbox-agent.ai-sandbox.svc.cluster.local` to resolve to. Three things still expect it, which is why this reads as an accidental deletion rather than a redesign: `tests/smoke/test_agent_sandbox.py::test_worker_service_exists` asserts exactly one `sandbox-agent` Service in the namespace, the module README lists that Service on :8080 under Invocation, and `deploy.sh` prints the same URL as the way to call the sandbox. Restoring the Service object looks like the fix.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against c328c9965.</sub>

</details>

osdc/integration-tests/workflows/integration-test.yaml.tpl
```diff
@@ -1955,3 +1955,66 @@ jobs:
...
+          echo ""
+          echo "PASS: sandbox Service reachable from arc-runners"
+      - name: Run a task through the sandbox (anonymous clone + real Bedrock call)
+        shell: bash
+        run: |
+          set -euo pipefail
+          RESP=$(curl -fsS -m 300 -X POST "$SANDBOX/run" \
+            -H 'Content-Type: application/json' \
+            -d '{"repo":"pytorch/pytorch","ref":"main","task":"List the top-level files."}')
+          echo "Response: $RESP"
```
izaitsevfb, line 1993: flagging this race, wound need to fix this in the final version

---

<details>
<summary>🟡 Once a POST is in flight, a cancelled run leaves the worker busy and the next run's POST gets a 429 that fails the job with no reason in the log. (ai-generated section)</summary>

The rewritten worker takes its task lock without blocking and answers HTTP 429 with `busy: one task at a time` when a task is already running. This workflow sets `concurrency: cancel-in-progress: true`, which kills the client but not the task the worker is executing — that can continue for several more minutes. The replacement run's health check then passes, because the listener is threaded, and the POST is refused. `curl -f` discards the response body, so the log shows only `curl: (22) The requested URL returned error: 429` and never the word busy. Retrying a couple of times on 429 is the fix; `--fail-with-body` in place of `-f` would also surface the reason, whereas simply dropping `-f` would let the 429 body fall through to the `cloned` check and report a clone failure that never happened.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against c328c9965.</sub>

</details>

osdc/integration-tests/workflows/integration-test.yaml.tpl
```diff
@@ -1955,3 +1955,66 @@ jobs:
...
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends curl
+      - name: Health check (reachable from a regular runner, like buildkitd)
+        shell: bash
+        run: |
+          set -euo pipefail
+          curl -fsS "$SANDBOX/healthz"
+          echo ""
+          echo "PASS: sandbox Service reachable from arc-runners"
```
izaitsevfb, line 1985: flagging this for the future.

---

<details>
<summary>🟡 The health check is a single unretried attempt, where the in-cluster checks it models itself on both retry. (ai-generated section)</summary>

`curl -fsS "$SANDBOX/healthz"` gets one try. The buildkit connectivity test it names as its model loops 45 times at 15s intervals with a comment about outlasting a cold pool, and three pypi-cache jobs in this same template run `for attempt in $(seq 1 $MAX_RETRIES)`. The sandbox is harder to bring up than either: one replica on a tainted fleet whose nodes need a region-scoped custom gVisor AMI, plus `imagePullPolicy: Always`. Nothing gates the canary on readiness either — `deploy.sh` waits on both rollouts but behind `|| true`, and the orchestrator that opens the canary PR never deploys at all.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against c328c9965.</sub>

</details>

izaitsevfb (dismissed):
see the "module no longer defines a `sandbox-agent` Service" — have you check if the PoC works after the recent changes?

osdc/integration-tests/workflows/integration-test.yaml.tpl
huydhn: This is not what I expected, let me get this fix to scale up the sandbox fleet, and will add a tests here for multiple requests in parallel

osdc/integration-tests/workflows/integration-test.yaml.tpl
huydhn: Yup, the latest refactoring to support sigv4 broke this, and the deployment / test has become stale. I'll get this fixed before merging. Subsequent changes would be smaller in scope than this godzilla stack

osdc/integration-tests/workflows/integration-test.yaml.tpl
huydhn: I have added a new test to run `Parallel capacity (4 sandboxes at once, forces a fleet scale-up)` https://github.com/pytorch/pytorch-canary/actions/runs/32280319552/job/96157306388

PR #995, opened 2026-08-11
Build the agent image in deploy.sh instead of by hand

huydhn:
* __->__ #995
* #994
* #993
* #992
* #997
* #996
* #991
* #1005

The module was undeployable from a clean checkout: the operator had to run
buildx + crane by hand before `just deploy-module` would work, against a mutable
`:prototype` tag that needed imagePullPolicy: Always to roll at all.

deploy.sh now builds and pushes it to the in-cluster Harbor `osdc` project under a
content-hash tag, mirroring modules/zombie-cleanup — an unchanged tree skips the
build, a code change deploys a new immutable tag, and the default IfNotPresent
pull policy becomes correct. Drops agent_sandbox.agent_image from clusters.yaml.
Requires a local docker daemon, like the other image-building modules.

Verified on meta-staging-aws-ue1: first deploy built and pushed the content tag,
second skipped the build, and /run still returns a grounded Bedrock report.


jeanschmidt (approved):

osdc/modules/agent-sandbox/deploy.sh
```diff
@@ -44,7 +52,70 @@ if [[ -z "$BEDROCK_DEFAULT_MODEL_ID" ]]; then
...
+  docker save "ci-agent-sandbox:${TAG}" -o "$IMAGE_TAR"
+  crane push "$IMAGE_TAR" "localhost:8081/osdc/ci-agent-sandbox:${TAG}" --insecure
+  rm -f "$IMAGE_TAR"
+  IMAGE_TAR=""
+fi
+
+# `wait` reaps the job so bash doesn't print "Terminated" into the deploy log.
+{ kill "$PF_PID" && wait "$PF_PID"; } 2>/dev/null || true
+PF_PID=""
+AGENT_IMAGE="${IMAGE}:${TAG}"
```
izaitsevfb, line 116: flagging as a minor bug

---

<details>
<summary>🟡 A `crane manifest` failure for ANY reason is read as "the tag is not there", so a transient registry error rebuilds and pushes over a tag that already exists. (ai-generated section)</summary>

The skip decision is `if crane manifest "localhost:8081/osdc/ci-agent-sandbox:${TAG}" --insecure &gt;/dev/null 2&gt;&amp;1`. A 404, a connection reset, an auth failure and a 5xx all arrive as the same nonzero exit, and only the 404 means what the `else` branch assumes. On any of the others the script rebuilds and pushes over an existing tag, and nothing at the registry refuses that — the project is created with no tag-immutability rule, so this `if` is the only thing standing between the tag and a second push.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f7e0b14e9.</sub>

</details>

osdc/modules/agent-sandbox/deploy.sh
```diff
@@ -44,7 +52,70 @@ if [[ -z "$BEDROCK_DEFAULT_MODEL_ID" ]]; then
...
 fi
 
-# --- Read terraform outputs (sandbox-agent Bedrock IRSA role) ---
+# --- Build + push the agent image (mirrors modules/zombie-cleanup) ---
+# Content-hash tag over the image inputs, so a code change deploys a new immutable
+# tag and an unchanged tree skips the build entirely. Tests are excluded — they
+# never enter the image, and hashing them would rebuild on test-only edits.
+IMAGE="harbor:30002/osdc/ci-agent-sandbox"
+TAG=$(find "$MODULE_DIR/agent" \( -name '*.py' -o -name 'Dockerfile' \) \
+  ! -name 'test_*' ! -name 'conftest.py' -print0 | sort -z | xargs -0 cat | sha256sum | cut -c1-12)
```
izaitsevfb, line 61: also a minor bug

---

<details>
<summary>🟡 The hash input set is a hand-maintained name filter over an UNCURATED build context, so the next non-`.py` image input silently reuses a stale image. (ai-generated section)</summary>

Today the filter is exactly right: the Dockerfile copies in only `server.py`, so `*.py` + `Dockerfile` minus tests really is the set of local files the image consumes. What does not hold is the build context — it is `$MODULE_DIR/agent` itself, so a new image input needs no edit to deploy.sh, whereas zombie-cleanup, the module this claims to mirror, assembles its context with explicit `cp` lines that sit immediately below its identical filter, so adding an input there forces you past the filter.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f7e0b14e9.</sub>

</details>

osdc/modules/agent-sandbox/deploy.sh
```diff
@@ -44,7 +52,70 @@ if [[ -z "$BEDROCK_DEFAULT_MODEL_ID" ]]; then
...
+for i in $(seq 1 30); do
+  if curl -s -o /dev/null "http://localhost:8081/api/v2.0/health" 2>/dev/null; then
+    break
+  fi
+  if [[ "$i" -eq 30 ]]; then
+    echo "[agent-sandbox] ERROR: Harbor port-forward not ready after 30s" >&2
+    exit 1
+  fi
+  sleep 1
+done
```
izaitsevfb, line 84: and another minor bug :)

---

<details>
<summary>🟡 The readiness probe cannot tell "my port-forward came up" from "something else already owns 8081" — and then the script talks to the wrong Harbor. (ai-generated section)</summary>

`kubectl port-forward` is backgrounded, so `set -e` cannot see it fail. If 8081 is already bound it exits non-zero with "address already in use" while `curl http://localhost:8081/api/v2.0/health` succeeds against the pre-existing listener. Everything downstream then addresses that registry: `crane manifest` answers about it, so a tag present there makes the build skip, and the Deployment is pinned to a tag this cluster's Harbor does not have. The pods land in ImagePullBackOff.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f7e0b14e9.</sub>

</details>

osdc/modules/agent-sandbox/deploy.sh
```diff
@@ -44,7 +52,70 @@ if [[ -z "$BEDROCK_DEFAULT_MODEL_ID" ]]; then
...
+HTTP_CODE=$(curl -s -o /dev/null -w "%{http_code}" \
+  -X POST "http://localhost:8081/api/v2.0/projects" \
+  --netrc-file "$NETRC_FILE" \
+  -H "Content-Type: application/json" \
+  -d '{"project_name":"osdc","public":true}')
+case "$HTTP_CODE" in
+  201) echo "[agent-sandbox] Created Harbor project 'osdc'" ;;
+  409) : ;; # already exists
+  *) echo "[agent-sandbox] Warning: Harbor project creation returned HTTP $HTTP_CODE" ;;
+esac
```
izaitsevfb, line 96: flagging, because it would need to be fixed in the future, but obviously non-blocking for the PoC

---

<details>
<summary>🟡 HTTP 409 proves the `osdc` project exists, not that it is public — and the pull path is anonymous. (ai-generated section)</summary>

The comment one line above states the invariant the code needs: "Nodes pull anonymously through the harbor:30002 mirror, so the project must be public." `public: true` is only ever sent on creation. A pre-existing PRIVATE `osdc` returns 409, which this treats as success; the push then succeeds because it is authenticated as admin, and the anonymous node pull fails.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f7e0b14e9.</sub>

</details>

osdc/modules/agent-sandbox/kubernetes/base/sandbox-agent.yaml
```diff
@@ -33,10 +33,9 @@ spec:
       automountServiceAccountToken: false
```
izaitsevfb, line 33: https://github.com/pytorch/ci-infra/blob/f7e0b14e92cc2b0c74bac79d665e2fae238c09a6/osdc/modules/agent-sandbox/deploy.sh#L165-L167

also flagged on the previous PR in the stack

---

<details>
<summary>🔴 `kubectl rollout status … || true` turns every new pull failure this PR can cause into a successful deploy. (ai-generated section)</summary>

These lines are PRE-EXISTING and outside the diff; what the PR changes is their risk profile. The script now mints the tag and pins the Deployment to it, so a tag the cluster cannot pull surfaces as pods that never start, and `|| true` converts that into exit 0. CI reports the deploy green after the ten-minute timeout.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f7e0b14e9.</sub>

</details>

osdc/modules/agent-sandbox/deploy.sh
```diff
@@ -44,7 +52,70 @@ if [[ -z "$BEDROCK_DEFAULT_MODEL_ID" ]]; then
...
+
+HARBOR_ADMIN_PW=$(kubectl get secret harbor-admin-password -n harbor-system \
+  -o jsonpath='{.data.password}' | base64 -d)
+NETRC_FILE=$(mktemp)
+chmod 600 "$NETRC_FILE"
+cat >"$NETRC_FILE" <<EOF
+machine localhost
+login admin
+password ${HARBOR_ADMIN_PW}
+EOF
```
izaitsevfb, line 71: nit, should be an easy BE fix

---

<details>
<summary>⚪ The Harbor admin password goes into argv and then persists in the operator's docker config, right beside careful hygiene for the same secret. (ai-generated section)</summary>

The password reaches `crane auth login … -p "$HARBOR_ADMIN_PW"`, where it sits in the process table for the life of the call and is then written to the docker config with no logout — while the netrc holding the same password for curl is mktemped, `chmod 600`ed and removed by the trap.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f7e0b14e9.</sub>

</details>

osdc/modules/agent-sandbox/kubernetes/base/sandbox-agent.yaml
```diff
@@ -33,10 +33,9 @@ spec:
...
         - name: agent
+          # Content-hash tag from deploy.sh — immutable, so the default
+          # IfNotPresent pull policy is correct.
           image: __AGENT_IMAGE__
-          # Prototype uses a mutable :prototype tag; re-pull on every start so a
-          # fresh push is picked up.
-          imagePullPolicy: Always
           ports:
             - name: http
               containerPort: 8080
```
izaitsevfb, line 41: probably a stale working, worth updating

---

<details>
<summary>⚪ "Immutable content-hash tag" claims reproducibility the tag does not have — worth rewording, now that `imagePullPolicy: Always` is gone. (ai-generated section)</summary>

The tag hashes the tree, not the image: `FROM public.ecr.aws/docker/library/debian:12-slim` is a moving tag, `docker build` runs without `--pull`, and the `apt-get install` line is unpinned, so the same tag can legitimately name different bytes built at different times. The ask is only wording: have the comment and the README say what holds — a source-hash tag, treated as immutable by deploy.sh.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against f7e0b14e9.</sub>

</details>

izaitsevfb (approved):
nothing blocking for PoC, but please consider that those would need to be eventually fixed

PR #996, opened 2026-08-11
Add the sandbox agent worker

huydhn:
* #994
* #993
* #992
* #997
* __->__ #996
* #991
* #1005

The workload that runs inside the sandbox: a single-file HTTP worker (server.py)
plus its image. POST /run shallow-clones a public repo and asks Bedrock about it;
GET /healthz backs the readiness probe. stdlib only — no AWS SDK, because Bedrock
is reached as plain HTTP through the signing proxy added later in this stack, so
this process never holds a credential.

Three things worth review attention:

- Every stage's failure is captured in `errors` rather than raised, so a caller
  can see how far the path got (cloned but no Bedrock is a different diagnosis
  from neither).
- The prompt is grounded in the clone's real top-level listing, returned as
  `top_level`. Given only a file count the model invents a plausible listing,
  which would make "Bedrock returned a report" prove nothing.
- Threaded listener with an explicit task lock. Single-threaded it could not
  answer /healthz while a clone ran, so the readiness probe timed out and the pod
  left the Service mid-task. One task at a time is still enforced; a second /run
  gets 429 rather than hanging.

It binds :: (AF_INET6) — on the IPv6-only cluster an AF_INET listener is
unreachable from the Service and the probe.

29 unit tests at 100% coverage, without mocking what matters: the clone runs a
real git clone (redirected off the network with insteadOf), Bedrock runs against a
fake signing proxy on a real socket, and the HTTP surface against a real socket.
Nothing deploys it yet.


jeanschmidt (approved):

osdc/modules/agent-sandbox/agent/server.py
```diff
@@ -0,0 +1,223 @@
...
+            # Grounding is best-effort — a repo we can't list is still worth asking about.
+            result["errors"]["listing"] = getattr(exc, "stderr", None) or str(exc)
+            entries = []
+        result["top_level"] = entries
+
+        prompt = build_prompt(repo, ref, task, result["file_count"], entries)
+        try:
+            result["report"] = invoke_bedrock(model, prompt)
+        except (urllib.error.URLError, TimeoutError, KeyError, ValueError) as exc:
+            result["errors"]["bedrock"] = str(exc)
```
izaitsevfb, line 257: looks like a good catch (a corner case, but we might want it addressed eventually)

---

<details>
<summary>🟡 A proxy that disconnects mid-response makes `run_task` raise, so the caller gets a closed connection instead of the `errors` object this design exists to return. (ai-generated section)</summary>

The Bedrock call catches `urllib.error.URLError`, `TimeoutError`, `KeyError` and `ValueError`. A truncated response body raises `http.client.IncompleteRead` and a reset while reading the status line raises `http.client.RemoteDisconnected` — neither is a subclass of any of the four, so both escape `run_task`, whose docstring says it never raises. Reproduced against a stand-in proxy that sends a Content-Length and then closes: `run_task` raised `IncompleteRead(11 bytes read, 489 more expected)`. The pod survives and the task lock is released by the `finally`, so this is not a wedge; what the caller receives is the one answer this design is trying to avoid, because a closed connection cannot be told apart from the pod being gone. `except (OSError, http.client.HTTPException, KeyError, ValueError)` covers both, and since `URLError` and `TimeoutError` are already `OSError` subclasses the clause gets shorter rather than longer. Worth the same visit: `BEDROCK_TIMEOUT_S` is urllib's per-operation socket timeout, not a wall clock, so a proxy that keeps trickling bytes holds the single task slot well past 120 s — an overall deadline is a separate line of code.

<sub>Reviewed by codex gpt-5.6-sol (codex-cli 0.147.0) at xhigh effort, against f9432e424.</sub>

</details>

osdc/modules/agent-sandbox/agent/server.py
```diff
@@ -0,0 +1,223 @@
...
+            self._send(404, {"error": "not found"})
+            return
+        length = int(self.headers.get("Content-Length", "0"))
+        try:
+            spec = json.loads(self.rfile.read(length) or "{}")
+            if not spec.get("repo"):
+                raise ValueError("missing 'repo'")
+        except (ValueError, json.JSONDecodeError) as exc:
+            self._send(400, {"error": str(exc)})
+            return
```
izaitsevfb, line 321: looks like worth tightening

---

<details>
<summary>🟡 A caller can hold an unbounded number of handler threads open, because the request body is read before the task lock with no size cap, no read deadline, and no check that Content-Length is non-negative. (ai-generated section)</summary>

New in character with the threaded listener. `do_POST` does `self.rfile.read(length)` on a trusted declared length before it touches `_TASK_LOCK`, and `ThreadingHTTPServer` starts a thread per connection with no upper bound — so a caller that announces a body and sends it slowly, or never, parks one thread each for as long as it likes. `int()` also accepts a negative value, and `read(-1)` reads until end of file, so `Content-Length: -1` with a connection that is never closed parks a thread with no ceiling at all. Measured: 30 half-open POSTs left 30 handler threads blocked in the read. This is not the previous blocking issue returning — /healthz stayed responsive and the task lock stayed free — it is unbounded thread and memory growth in a 4 GiB pod, and the earlier report's guess that a one-slot guard would defuse it is wrong, because the read happens first. Fix: reject a Content-Length that is negative or above a few KB before reading, and set a socket timeout on the connection so a stalled read cannot outlive it.

<sub>Reviewed by codex gpt-5.6-sol (codex-cli 0.147.0) at xhigh effort, against f9432e424.</sub>

</details>

izaitsevfb, line 321: worth tightening at some point

---

<details>
<summary>🟡 Three ways to malform a /run request produce no HTTP response at all, and one of them makes `run_task` raise — which its docstring says it never does. (ai-generated section)</summary>

`do_POST` guards `json.loads` with `except (ValueError, json.JSONDecodeError)`, but (a) a body that is valid JSON and not an object reaches `spec.get("repo")` and raises AttributeError, (b) the `int(self.headers.get("Content-Length", "0"))` sits one line ABOVE the try, so a non-numeric header raises ValueError uncaught, and (c) `{"ref": null}` survives validation and `spec.get("ref", "main")` returns None for an explicit null, which reaches subprocess as a command argument. Measured at this head: bodies `[]`, `null` and `3` and a `Content-Length: abc` header each returned no response at all, and `run_task({"repo": …, "ref": None})` raised TypeError. The caller gets a closed connection instead of the 400 or the `errors` object, and a closed connection cannot be told apart from the pod being gone. Nothing wedges — the `try`/`finally` releases the task lock, and a following /run and /healthz both answered 200 — so the cost is diagnosability, not availability. Only the `{}` case is tested. Fix: move the Content-Length parse inside the try, require `isinstance(spec, dict)`, and type-check each field rather than truth-testing it, since `spec.get("ref") or "main"` still lets `0` or `[]` through to a subprocess argument.

<sub>Reviewed by codex gpt-5.6-sol (codex-cli 0.147.0) at xhigh effort, against f9432e424.</sub>

</details>

osdc/modules/agent-sandbox/agent/server.py
```diff
@@ -0,0 +1,223 @@
...
+        method="POST",
+        headers={
+            "Host": f"bedrock-runtime.{REGION}.amazonaws.com",
+            "Content-Type": "application/json",
+        },
+    )
+    with urllib.request.urlopen(req, timeout=BEDROCK_TIMEOUT_S) as resp:  # noqa: S310
+        payload = json.loads(resp.read())
+    content = payload.get("content") or []
+    return content[0]["text"] if content else ""
```
izaitsevfb, line 180: this looks like a bug, could you please confirm?

---

<details>
<summary>🟡 `model` is interpolated into the request path without percent-encoding, so an ARN — which Bedrock accepts as a model identifier and which contains a `/` — produces a different request than intended. (ai-generated section)</summary>

`invoke_bedrock` builds `http://{SIGV4_PROXY}/model/{model}/invoke` by string interpolation, and urllib neither escapes nor normalises the result. An inference-profile or foundation-model ARN is a documented model identifier and contains a `/`, so it lands on the wire as two path segments rather than one and the operation is no longer a model invoke. `urllib.parse.quote(model, safe="")` is the whole fix, and it is also what the Bedrock API asks for — the identifier is a single URL-encoded path segment. Two things ride along on the same line. `model` comes from the body of an endpoint with no authentication, so today the caller, not this worker, decides the path the proxy signs: measured, a `model` of `../../async-invoke#` leaves the wire as `POST /model/../../async-invoke`, the `#` discarding the `/invoke` suffix. What bounds that is outside this file — the Host header is set from AWS_REGION rather than from the request, and the proxy's role allows only InvokeModel and InvokeModelWithResponseStream — which is why the encoding bug is the part worth acting on and the rest is a question: what does the proxy do with a request path that is not a model invoke, reject it or sign and forward it?

<sub>Reviewed by codex gpt-5.6-sol (codex-cli 0.147.0) at xhigh effort, against f9432e424.</sub>

</details>

osdc/modules/agent-sandbox/agent/server.py
```diff
@@ -0,0 +1,223 @@
...
+    subprocess.run(
+        ["git", "clone", "--depth", "1", "--branch", ref, f"https://github.com/{repo}.git", dest],
+        check=True,
+        capture_output=True,
+        text=True,
+        timeout=CLONE_TIMEOUT_S,
+        # A private repo would otherwise make git prompt for a username and block
+        # until the clone timeout instead of failing.
+        env={**os.environ, "GIT_TERMINAL_PROMPT": "0"},
+    )
```
izaitsevfb, line 77: worth adding a FIXME if intended, we definitely want to fix that in the final version.

---

<details>
<summary>🟡 `ref` accepts only a branch or tag, never a commit sha — and a pinned revision is the natural thing for a canary to ask for. (ai-generated section)</summary>

`clone_repo` passes `ref` to `git clone --branch`, which resolves branches and tags only, so a caller pinning a sha gets a clone failure whose message does not explain the restriction — while the field is named `ref` and defaults to `main`, which reads as 'any revision'. The cheap answer is to document it as branch-or-tag in the module docstring and the endpoint comment; the fuller one is `git init` + `fetch --depth 1 origin &lt;ref&gt;` + `checkout FETCH_HEAD`, which accepts both — but confirm first that fetch-by-object-id is permitted for the repo in question, since that is a server-side setting, and validate `ref` there because it becomes a fetch refspec.

<sub>Reviewed by codex gpt-5.6-sol (codex-cli 0.147.0) at xhigh effort, against f9432e424.</sub>

</details>

osdc/modules/agent-sandbox/agent/server.py
izaitsevfb, line 257: worth surfacing the code for debuggability

---

<details>
<summary>⚪ When Bedrock refuses a request, `errors.bedrock` reads `HTTP Error 403: Forbidden` and nothing more — the AWS error code is in the response body, which is never read. (ai-generated section)</summary>

`str()` of a `urllib.error.HTTPError` is only the status line; the body carrying the AWS error code sits unread on the exception object. Measured: a stand-in proxy replying 403 with a Bedrock AccessDenied body produced exactly `HTTP Error 403: Forbidden`. The status code still separates 'not authorised' from 'throttled' and the proxy logs failed requests, which is why this is a diagnosability nit rather than a defect — but for an endpoint whose stated purpose is showing exactly how far the credential path got, a model not enabled for the account and a throttle are the two likeliest failures and now say least. Parse the `__type` or `message` field out of the body and return the AWS error code alone, rather than a slice of the raw body: an AccessDenied message names the assumed-role ARN, which is the proxy's identity now, and does not need to be in a response any network-authorised caller can read.

<sub>Reviewed by codex gpt-5.6-sol (codex-cli 0.147.0) at xhigh effort, against f9432e424.</sub>

</details>

izaitsevfb (approved):
solid PoC, thanks for doing it!

please take a look at the AI findings, they are not blocking, but they are worth remembering, as they would need to be addressed in the final version.

osdc/modules/agent-sandbox/agent/server.py
huydhn: Yeah, we need to follow up with a discussion on how best to handle the initial checkout inside the sandbox. This is a cheap way to unblock the prototype. At the very least, we would need to have `mitmproxy` in place too to handle private repo checkout with token.

osdc/modules/agent-sandbox/agent/server.py
huydhn: Good catch, it's a bug. I have been using `us.anthropic.claude-haiku-4-5-20251001-v1:0` as the default model and it fails to catch this case

osdc/modules/agent-sandbox/agent/server.py
huydhn: Done

osdc/modules/agent-sandbox/agent/server.py
huydhn: Done

osdc/modules/agent-sandbox/agent/server.py
huydhn: Fixed, I tighten this with `HTTPError` follow up with `OSError, http.client.HTTPException, KeyError, TypeError, ValueError` to try to catch them all

osdc/modules/agent-sandbox/agent/server.py
huydhn: This is an interesting attack vector, I haven't put much thought here yet (just assuming a trusted caller for the time being). In the mean time, I have asked the agent to do some basic checks here including:
* timeout so a stalled read can't park a thread forever
* Content-Length rejected when negative or above MAX_BODY_BYTES (64 KiB)

PR #997, opened 2026-08-11
Add a read-only Bedrock IRSA role for the sigv4 proxy

huydhn:
* #994
* #993
* #992
* __->__ #997
* #996
* #991
* #1005

The one AWS credential in the design, and it deliberately does not belong to the
sandbox: the role binds to the sigv4-proxy service account, which signs requests
on the agent's behalf. The agent gets no AWS identity at all, so a prompt-injected
agent has nothing to steal — and IMDS hop-limit 1 keeps the node role out of reach
too. Scoped to bedrock:InvokeModel and nothing else.

The resource list needs `arn:aws:bedrock:*::foundation-model/*`, not just this
region's, which is worth explaining: every Anthropic model in us-east-1 is
INFERENCE_PROFILE-only (the one still advertising ON_DEMAND is refused as
provider-legacy), so the call must go to a cross-region profile — and that is
authorized against the foundation-model ARN in whichever region the profile routes
to. Pinned to one region, invokes fail AccessDenied as soon as routing leaves it.
Still InvokeModel-only, and profiles stay scoped to this account.

Its own tofu root in the cluster's state; nothing consumes the output yet.


jeanschmidt (approved):

huydhn:
cc @izaitsevfb From what I see, implementing sigv4 proxy looks straightforward, so I decide to redo this PR to implement this right away. A couple of changes to call out:

* The Bedrock role will not be assigned to agent sandbox nodes, but to the those running sigv4 proxy service instead. They don't overlap
* Instead of using boto3, the `server.py` script (dispatcher skeleton) calls Bedrock via the sigv4 proxy https://github.com/pytorch/ci-infra/pull/996/changes#diff-f1355a658a71132ca080b4b2aceb5fd7d1f0305a37eb0b50f432d6c50cb6056dR82

The `mitmproxy` part looks more complex, so I will work on it later. One challenge is where and how to run git checkout, which will put CPU and memory pressure on the proxy at scale if we decide to route the traffic through it.

osdc/modules/agent-sandbox/terraform/main.tf
```diff
@@ -0,0 +1,133 @@
...
+  # (a `us.` profile can land in us-east-1/us-east-2/us-west-2, and AWS may add
+  # more). Hence the region wildcard on foundation-model: without it the invoke
+  # fails AccessDenied whenever routing leaves ${var.aws_region}.
+  #
+  # Still InvokeModel-only, and still scoped to this account's profiles. Tighten
+  # to specific model IDs once the agent's model is fixed.
+  bedrock_resources = [
+    "arn:aws:bedrock:*::foundation-model/*",
+    "arn:aws:bedrock:${var.aws_region}:${local.account_id}:inference-profile/*",
+  ]
```
izaitsevfb: not blocking here, but consider [an alternative](https://gist.github.com/izaitsevfb/3585e4ea15c724ce75c480dc5ffae56c
) for the final version

---

<details>
<summary>🟡 `foundation-model/*` is two wildcards, and the comment only argues for one of them. (ai-generated section)</summary>

The region wildcard is correct and the comment's reasoning for it holds. The `/*` on the model id has no stated justification, and the comment defers it ("tighten to specific model IDs once the agent's model is fixed") as though a narrower rule needed the exact model — it does not. Because the routed-region authorization uses the base model id, `foundation-model/anthropic.*` plus `inference-profile/us.anthropic.*` is narrower today, pins no exact model or version, and keeps the routing argument intact; it is also the shape AWS's own cross-region-inference policy example uses. Two consequences the current list has that the comment does not mention: the foundation-model statement also permits direct on-demand invocation of any matching model, independently of any inference profile, in every region; and the model id is chosen by the caller, not by config — `run_task` takes it from the request body and interpolates it straight into the proxy URL path. The narrowing also has a control that closes the direct-invoke half outright rather than just shrinking the list: condition the foundation-model statement on `bedrock:InferenceProfileArn`, leaving an unconditioned Allow on the profile ARN and a conditioned Allow on the foundation-model ARNs it routes to. A direct on-demand invoke carries no profile ARN in its request context, so the condition evaluates false and nothing else grants it — the profile becomes the only path, and it fails closed. AWS's own policy example, why dropping the foundation-model ARN outright breaks cross-region routing, the region-enumeration tradeoff and a two-call acceptance test: https://gist.github.com/izaitsevfb/3585e4ea15c724ce75c480dc5ffae56c

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 3174d084d.</sub>

</details>

osdc/modules/agent-sandbox/terraform/main.tf
```diff
@@ -0,0 +1,133 @@
...
+          "cloudwatch:GetMetricData",
+          "cloudwatch:GetMetricStatistics",
+          "cloudwatch:ListMetrics",
+          "logs:GetLogEvents",
+          "logs:FilterLogEvents",
+          "logs:DescribeLogGroups",
+          "logs:DescribeLogStreams",
+        ]
+        Resource = "*"
+      },
```
izaitsevfb, line 157: this looks like a contradiction, worth clarifying.

---

<details>
<summary>⚪ The new `ReadOnlyObservability` statement reads every CloudWatch metric and log group in the account, and nothing in the tree reads either. (ai-generated section)</summary>

The proxy this role is bound to runs with no `--name`, so the target AWS service comes from each request's `Host` header — its own manifest in #992 says "one proxy serves bedrock-runtime, cloudwatch, logs, etc.". The proxy therefore contributes no scope of its own, and this policy is the whole boundary between the sandboxed agent and AWS: `logs:FilterLogEvents` on `Resource = "*"` is every log group in the account, not just this module's. There is also no consumer for it — `server.py` only POSTs to `/model/&lt;id&gt;/invoke`, so the grant lands ahead of anything that needs it. The PR description still says the role is "Scoped to bedrock:InvokeModel and nothing else", which this statement contradicts, and the header's "and nothing else" now covers a much larger surface than it did. Two ways to close it: drop the statement until something needs it, or split it by resource form — the CloudWatch metric actions have no resource-level form and do need `*`, while the read side can name the log groups the agent is meant to see rather than all of them.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 3174d084d.</sub>

</details>

osdc/modules/agent-sandbox/terraform/main.tf
```diff
@@ -0,0 +1,133 @@
...
+data "terraform_remote_state" "base" {
+  backend = "s3"
+
+  config = {
+    bucket         = var.state_bucket
+    key            = "${var.cluster_id}/base/terraform.tfstate"
+    region         = "us-west-2"
+    dynamodb_table = "ciforge-terraform-locks"
+  }
+}
```
izaitsevfb, line 43: worth consolidating?

---

<details>
<summary>⚪ This is the third copy of the `us-west-2` + `ciforge-terraform-locks` pair that `scripts/state-config.sh` exists to prevent. (ai-generated section)</summary>

`state-config.sh` calls itself the "Single source of truth for the AWS region that hosts the shared state bucket(s) and the DynamoDB lock table", with the stated goal that a state-region migration be "a one-file edit instead of a sprawling search-and-replace". This block hardcodes both values; karpenter and hf-cache already do the same, so this makes three. The honest cost of fixing it: `terraform_remote_state` cannot read a shell variable, and both justfile recipes pass a fixed four-var set, so a real fix touches three roots and two recipes. Either do that or accept it as existing debt — a comment pointing at `state-config.sh` buys little.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 3174d084d.</sub>

</details>

izaitsevfb (approved):
lgtm overall, thanks for doing that! please check the findings below.

the CI coverage is worth addressing right away imo.

izaitsevfb:
this is important. let's make sure that it's covered by CI

---

<details>
<summary>🔴 CI parses this new root but nothing resolves it — no job runs `tofu validate` or `plan` on it. (ai-generated section)</summary>

Every check on this head is green and none of them resolves this root. The `lint` check does reach the file: `just lint` runs `tofu fmt -check -recursive` over `modules/`, which requires parseable, canonically formatted HCL, and that is all it requires. `just plan` iterates only the modules a cluster lists in `clusters.yaml`, `agent-sandbox` is in no cluster's list at this head, and the plan workflow's matrix is five prod clusters with no staging equivalent — the plan output posted back on this head names Base, karpenter and hf-cache and never this module. So the provider graph and the two `data.terraform_remote_state.base.outputs.oidc_provider*` reads have not been resolved by anything. Those two halves need different things: `tofu init -backend=false &amp;&amp; tofu validate` from the module directory closes the provider and schema half cheaply, but only a real plan against a cluster's state proves the remote-state reads resolve. Adding a staging cluster to the matrix does nothing on its own unless the module is also enabled there.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 3174d084d.</sub>

</details>

izaitsevfb:
worth a consideration (in the final version we should bound the cost)

---

<details>
<summary>For the PR that deploys the endpoint: what bounds Bedrock spend? (ai-generated section)</summary>

`POST /run` has no authentication and the NetworkPolicy admits the whole arc-runners namespace. The worker being one-task-at-a-time limits concurrency, not sustained request rate, and `max_tokens: 1024` caps output tokens, not input — so neither is a spend control. Nothing in this role bounds spend and nothing here needs to, but something should before the endpoint accepts traffic — alongside the question of who picks the model.

<sub>Reviewed by codex gpt-5.6-sol at xhigh effort, against 3174d084d.</sub>

</details>

osdc/modules/agent-sandbox/terraform/main.tf
huydhn: Yeah, it's badly named. Fixed

osdc/modules/agent-sandbox/terraform/main.tf
huydhn: It seems that each k8s module that use Terraform follows the same practice, leading to the duplication of `terraform_remote_state`. Let me file a BE task just in case we want to pick it up at some point https://github.com/pytorch/ci-infra/issues/1019

osdc/modules/agent-sandbox/terraform/main.tf
huydhn: This part somehow gives me a headache, I need to go read the history of how AWS comes up with the concept of inference profile in the first place. So, the changes are in place:

* Narrow this to only Anthropic model for the time being
* Allow Anthropic inference profile (they don't support anything else nowadays)
* Routing within us regions `ue1`, `ue2`, and `uw2`

huydhn:
> this is important. let's make sure that it's covered by CI
> 
> 🔴 CI parses this new root but nothing resolves it — no job runs `tofu validate` or `plan` on it. (ai-generated section)

Good catch, it's fixed now

PR #999, opened 2026-08-11
Add cluster RBAC for listener node reads

jeanschmidt:
* #1001
* #1003
* #1002
* #1000
* __->__ #999

----

- New ClusterRole arc-listener-node-reader granting get/list/watch on Nodes
- ClusterRoleBinding to the system:serviceaccounts:arc-systems group
- Enables the capacity monitor to read Node Spec.Unschedulable and Spec.Taints

Notes:
Node is cluster-scoped, so this grant must be cluster-wide, complementing the
namespaced pod/configmap grants in capacity-monitor-rbac.yaml. It binds the
arc-systems SA group rather than a named SA because each scale set gets a
dynamically named listener SA (<scale-set>-<hash>-listener) created by the ARC
controller. The placeholder schedulability check uses cordon/taint state to
decide whether a placeholder still holds a node a replacement runner could
schedule onto. This manifest is not yet wired into modules/arc/deploy.sh, which
currently applies only capacity-monitor-rbac.yaml.

### DIFF
--- a/osdc/modules/arc/kubernetes/listener-node-reader-rbac.yaml
+++ b/osdc/modules/arc/kubernetes/listener-node-reader-rbac.yaml
@@ -1,0 +1,39 @@
+# Cluster-scoped RBAC for the capacity monitor running in listener pods.
+# Grants read-only access to Node objects so the capacity monitor's placeholder
+# schedulability check can read Spec.Unschedulable (cordon) and Spec.Taints to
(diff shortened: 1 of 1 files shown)


huydhn:
Claude points out that this grant already exists in `modules/arc/kubernetes/capacity-monitor-rbac.yaml` already ships this in `arc-capacity-monitor-nodes`, so this might be redundant

PR #1000, opened 2026-08-11
arc-runners: per-org fan-out so one cluster serves additional GitHub orgs

jeanschmidt:
* #1001
* #1003
* #1002
* __->__ #1000
* #999

**Impact:** OSDC ARC runner generation; live only on uw1 (adds meta-pytorch runners beside pytorch). Byte-identical for every cluster without an `additional_orgs` block.
**Risk:** medium

## What
Teaches the arc-runners generator to emit an extra scale set per `additional_orgs` entry: each additional org gets an org-unique Kubernetes name (`<resource_slug>-<def>`), its own GitHub URL/secret/runner-group, and its own HUD-queue shard — while keeping the SAME GitHub runner label (`<prefix><def>`) as the primary org. uw1 is configured with a `meta-pytorch` org so it runs those runners alongside `pytorch`.

## Why
We need `meta-pytorch` (and potentially other orgs) to be served by the existing runner fleets without standing up separate clusters or duplicating every runner def. The fan-out is purely additive: when no `additional_orgs` are set, generation is unchanged, so existing clusters are unaffected.

The design keeps a few things honest:
- HUD-queue sharding is now org-scoped (`org_key`), so an org only shards against peers that actually serve it — a cluster carrying a solo additional org forms its own group of one instead of mis-sharding against clusters pointed at a different org.
- A preflight validates each `additional_orgs` entry before anything is written: required keys, DNS-1123 `resource_slug`, the 45-char chart name limit, and output-name collisions (primary vs org, org vs org). Misconfiguration fails at generate time with a clear message instead of a raw `KeyError` or an opaque helm-render failure at deploy, and leaves the last good render intact.

The smoke tests were reworked from def-name keying to a `GeneratedScaleSet` model that reads the generated files as the source of truth (indexed by the org-unique scale-set-name label). A def-keyed model would conflate the two orgs' listeners/placeholders/ConfigMaps that share a def, or flag every additional-org resource as stale.

# Notes
- Before uw1 deploy: the `meta-prod-aws-uw1-meta-pytorch` GitHub App/secret and the `meta-pytorch` runner-group must exist.
- uw1's `arc.chart_version` is pinned to `0.14.1-jeanschmidt.19` (the fork release carrying the `resourceName` decoupling this relies on).


huydhn:
There seems to be a latent bug here because CAPACITY_AWARE_HUD_API_URL today is hardcoded to pytorch-org https://github.com/pytorch/ci-infra/blob/31ba82a949bbfc5aea86f9dd53a8e979397e9d37/osdc/modules/arc-runners/templates/runner.yaml.tpl#L203

Also, I suspect that there might be double counting in the number of place holder pods with this setup because we have the pytorch+meta-pytorch queue showing up in both scale sets. Please verify this is not the case, I haven't looked into the new change in `0.14.1-jeanschmidt.19` closely yet, but will do it now

osdc/modules/arc-runners/scripts/python/generate_runners.py
```diff
@@ -483,12 +564,102 @@ def generate_runner(
...
+    for idx, entry in enumerate(additional_orgs):
+        entry = entry or {}
+        for key in ("github_config_url", "github_secret_name", "resource_slug"):
+            value = entry.get(key)
+            if not isinstance(value, str) or not value.strip():
+                errors.append(
+                    f"cluster '{cluster_id}' additional_orgs[{idx}]: '{key}' is required and must be a non-empty string"
+                )
+        slug = entry.get("resource_slug")
+        if not isinstance(slug, str) or not slug.strip():
```
huydhn, line 630: Why is this not an error too?

PR #1006, opened 2026-08-12
buildkit: allow m6id.12xlarge alongside m6id.24xlarge on the amd64 pool

huydhn:
## Problem

The amd64 BuildKit NodePool is pinned to a **single instance type**, on-demand only:

```yaml
- key: node.kubernetes.io/instance-type
  operator: In
  values: [m6id.24xlarge]
- key: karpenter.sh/capacity-type
  operator: In
  values: [on-demand]
```

That leaves Karpenter zero fallback. When us-east-2 ran out of m6id.24xlarge on-demand capacity, amd64 scale-up froze — the pool stalled at 12 nodes while KEDA asked for 37 replicas:

```
"skipping, nodepool requirements filtered out all instance types" NodePool=buildkit-amd64
```

~63 `InsufficientInstanceCapacity` rejections in a 13-minute window, evenly spread across 2a/2b/2c. 15 pods sat `Unschedulable`, and the prod ue2 smoke test failed on `buildkitd-amd64: 16/37 replicas ready` ([run](https://github.com/pytorch/ci-infra/actions/runs/31558547335/job/94268338341)).

## Change

Add `m6id.12xlarge` to the pool. It's exactly half a 24xlarge, so it holds **one** 42 vCPU / 155Gi pod where the 24xlarge holds two:

| type | vCPU/GiB | $/hr | pods | $/pod-hr | vCPU used |
|---|---|---|---|---|---|
| m6id.12xlarge | 48 / 192 | 2.848 | 1 | **2.848** | 88% |
| m6id.24xlarge | 96 / 384 | 5.695 | 2 | **2.848** | 88% |

Identical cost per pod, identical packing efficiency, identical 2850 GB NVMe per pod. Same CPU generation, so build throughput is unchanged. Karpenter gets two placement options instead of one.

16xl/32xl were considered and rejected — 42 vCPU doesn't tile into 64 or 128, so they land at 66% efficiency and a 33% premium per pod.

Smaller sizes also have deeper pools. In the same window Karpenter was failing every 48xlarge request while 12xl/24xl landed:

| type | launched | ICE'd |
|---|---|---|
| g5.48xlarge | 0 | 55 |
| g5.12xlarge | 3 | 9 |
| g6.48xlarge | 0 | 2 |
| g6.12xlarge | 2 | 3 |

## Implementation

- `--amd64-instance-type` / `--arm64-instance-type` accept a comma-separated list. **The first entry sizes the pod**, so the existing 42 vCPU / 155Gi spec is untouched.
- Generation fails loudly if any listed type can't fit one pod of the reference size.
- The per-size `instance-type` **label and taint are dropped from the NodePool** — they can't describe a multi-size pool. The `workload/buildkit-<arch>` taint already isolates it, and `nodeSelector` now pins arch via `kubernetes.io/arch`.
- The pod tolerates `instance-type` with **`operator: Exists`**, so it still schedules onto the existing nodes that carry the old `instance-type=m6id.24xlarge` taint. No forced node replacement.
- `m6id.12xlarge` added to `INSTANCE_SPECS` (`memory_mi` follows the table's existing ratio; `max_pods: 234`, confirmed against every 12xlarge node running in prod ue2).

Applied to both prod clusters and the `deploy.sh` default, so staging picks it up too.

## Verification

- Regenerated manifests: amd64 pod spec, NodePool limits (34560 CPU / 138240Gi) and **all arm64 output are byte-identical** to before.
- Generator reports `m6id.24xlarge: 2 pod(s) per node`, `m6id.12xlarge: 1 pod(s) per node`.
- `just test` — passes, 98.87% coverage.
- `just lint` — 12/13; the one failure is a pre-existing hadolint finding on `base/docker/runner-base/Dockerfile` present on an unmodified tree.

## Notes for review

Two things deliberately out of scope, both surfaced while debugging the same smoke test failure:

1. A node has been flapping `DiskPressure` since Aug 7 and has accumulated **193 Evicted** buildkitd pods. Likely orphaned `subPathExpr: $(POD_NAME)` cache dirs never being reclaimed — worth its own fix.
2. `dcgm-exporter` is OOMKilling fleet-wide (58/197 pods), the other half of that same smoke test failure.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


huydhn:
Superseded by a ghstack stack: #1007 (prod amd64 capacity fallback) → #1008 (staging small pool). Same first commit, no review had landed here.

PR #1007, opened 2026-08-13
buildkit: let a NodePool span instance sizes, add m6id.12xlarge to prod amd64

huydhn:
* #1008
* __->__ #1007

The amd64 pool was pinned to one instance type, on-demand only, so Karpenter
had no fallback: when us-east-2 ran out of m6id.24xlarge it logged "nodepool
requirements filtered out all instance types" against ~63 ICE rejections in 13
minutes across all three AZs, stalling at 12 nodes while KEDA asked for 37.

Instance types become a `type: pods_per_node` map in preference order, shared
via `defaults.buildkit`. m6id.12xlarge is half a 24xlarge — same pod spec, same
$/pod-hour, same packing:

    amd64_instance_types:
      m6id.24xlarge: 2
      m6id.12xlarge: 1

**Each entry becomes its own weighted NodePool** (100, then 90) sharing one
EC2NodeClass per arch, so the 24xlarge serves normal scale-up and the 12xlarge
is reached only on ICE or limits. One pool spanning both sizes would instead
pick per pending pod on price — and since the sizes are price-proportional,
trickle scale-up would yield a fleet of 1-pod 12xl nodes. Each pool carries the
full replica limit, since a spill absorbs the fleet alone.

One Deployment means one pod spec, so **the first entry sizes the pod**; later
entries must hold at least their declared count or generation fails. Order is
load-bearing and tested.

Both arches drop the per-size `instance-type` label and taint — they cannot
describe a pool set spanning sizes. `workload/buildkit-<arch>` already isolates
them and the pod tolerates `instance-type` with `Exists`, so it still schedules
onto existing nodes. Pod resources and `autoscaling.yaml` are unchanged.

Pool names are derived from config, so dropping or renaming an entry leaves
its NodePool behind — and an orphan is not inert, since the pod now selects on
`workload-type` and arch alone and tolerates `instance-type` with `Exists`. It
would keep provisioning a size no longer configured. Generated NodePools and
EC2NodeClasses are therefore stamped `osdc.io/module: buildkit` and `deploy.sh`
sweeps anything carrying that label but absent from the manifest, matching
`modules/nodepools`. This also makes them visible to `just remove-module`.

The EC2NodeClass `NodePool` tag is dropped for the same reason `InstanceType`
was: one class is now shared by every pool of an arch, so it cannot name the
pool that launched a node.

**Known limit:** both sizes are m6id and ICE is often family+AZ-wide, so a
family-wide crunch still stalls. A real fallback needs another NVMe family
(`m7id.*`), which the git cache and `instanceStorePolicy: RAID0` both require.


jeanschmidt:
## Findings

### Tier 1 — address before / around merge

#### [T1-1] Silently changes steady-state provisioning, not just the failure fallback
- **Severity:** Medium
- **Category:** Design / behavior
- **Location:** `osdc/modules/buildkit/scripts/python/generate_buildkit.py:451-464` (single unweighted NodePool, on-demand, no `weight`/`minValues`)
- **Issue:** One unweighted pool with two price-equal sizes + on-demand-lowest-price means
  Karpenter prefers the **cheaper 12xl** for any single pending pod, even when 24xl is plentiful.
  Under KEDA's default (trickle) scale-up, one-build-at-a-time arrivals produce a fleet skewed to
  1-pod 12xl nodes.
- **Impact:** Up to ~2× the node count of the old 24xl/2-pod packing under light load.
  Compute-cost-neutral and packing-neutral, but ~2× EBS root volumes, ~2× ENI/IP consumption, and
  ~2× git-cache warm-ups → more cold starts. `WhenEmpty` consolidation means the fleet does not
  self-correct until pods drain.
- **Suggested fix:** If the intent is "24xl normally, 12xl only on scarcity," express it with
  **two weighted NodePools** (24xl `weight: 100`, 12xl low weight) — Karpenter only spills to the
  low-weight pool on ICE/limits. Otherwise, consciously accept the re-shaping and note it in the PR.

#### [T1-2] No guard against a silent cluster-wide pod-shrink
- **Severity:** Medium
- **Category:** Footgun / robustness
- **Location:** `osdc/modules/buildkit/scripts/python/generate_buildkit.py` — `plan_pod_resources` (~:144-156)
- **Issue:** The pod takes the *smallest* entry. A one-character `clusters.yaml` edit — bumping
  `pods_per_node` on the smaller size — silently shrinks **every** build pod on that arch across
  all clusters. The only guard is `<1 vCPU / <1Gi` ("too small to schedule"): `m6id.12xlarge:8` →
  5 vCPU pods, `:40` → 1 vCPU, all still pass.
- **Impact:** All builds degrade invisibly; the only signal is a log line.
- **Suggested fix:** Assert the resulting size is ≥ what the primary/largest entry implies (or gate
  denser packing behind an explicit flag); add a smoke assertion on expected pod size.

### Tier 2 — cheap correctness / quality fixes

#### [T2-1] A code comment is factually wrong (NodePool limit)
- **Severity:** Medium
- **Category:** Correctness / misleading comment
- **Location:** `generate_buildkit.py:551-554` + `_nodepool_limits` (~:412-421)
- **Issue:** Limits are derived from the *first* map entry (`plan[0]`); the comment claims the
  total is "the same whatever mix of sizes Karpenter picks." False in general — only true here
  because 12xl is exactly half a 24xl. `{24xl:1, 12xl:1}` flips the limit by list order.
  (Flagged independently by 3 reviewers.)
- **Impact:** Latent trap the moment a non-proportional size is added; the comment misleads.
- **Suggested fix:** Fix the comment to "sized off the first entry," ideally compute the limit
  across all entries (max), add an ordering test.

#### [T2-2] "Single source" undercut by triple duplication
- **Severity:** Medium
- **Category:** DRY / single source of truth
- **Location:** `osdc/clusters.yaml:84-88` (defaults), `:425-429` (meta-prod-aws-ue1), `:477-481`
  (meta-prod-aws-ue2) — byte-identical maps; explanatory comment copy-pasted at `:79-83`,
  `:422-424`, `:474-476`. 4th copy in `osdc/docs/prod-cluster-ha-us-west-1.md:64-69`.
- **Issue:** The two prod clusters could inherit from `defaults` exactly like the 3 staging
  clusters already do; the overrides add only drift risk. Contradicts the PR's stated intent.
- **Impact:** Change the default packing and the two highest-traffic prod fleets silently keep
  stale values.
- **Suggested fix:** Delete the redundant prod overrides (true single source) **or** use a YAML
  anchor (`&buildkit_instances` on defaults, `*buildkit_instances` at each site).

#### [T2-3] The new integration seam has zero tests
- **Severity:** Medium
- **Category:** Test coverage
- **Location:** `osdc/scripts/cluster-config.py:135-138` (new dict → `"k:v,k:v"` branch)
- **Issue:** This is what wires `clusters.yaml` to the generator. Verified correct by hand
  (emits `m6id.24xlarge:2,m6id.12xlarge:1`, round-trips through `parse_instance_plan`) but nothing
  guards its shape/order.
- **Impact:** A wrong separator/order/empty-dict handling silently breaks the whole feature.
- **Suggested fix:** Add a unit test on the dict branch (assert the string, order preservation,
  empty-dict → "").

#### [T2-4] DRY + file size in the generator
- **Severity:** Low–Medium
- **Category:** DRY / structure
- **Location:** `generate_buildkit.py:86-93` (`compute_pod_resources`) vs `:162-167`
  (`pods_that_fit`); whole file now **749 lines** (repo hard limit 400).
- **Issue:** The usable-CPU/mem formula (spec lookup → ENI reserve → kubelet reserve → daemonset
  overhead) is duplicated across two functions. File exceeds the size recommended limit (no linter enforces it).
- **Impact:** Overhead/reserve model must change in lockstep or sizing and fit-checking disagree.
- **Suggested fix:** Extract `_usable_resources(instance_type) -> (cpu_m, mem_mi)` and call from
  both; split the YAML templates / resource helpers into a sibling module.

### Tier 3 — robustness (arguably out of scope, worth noting)

#### [T3-2] Same-family fallback + on-demand-only
- **Severity:** Low–Medium
- **Category:** Robustness / capacity
- **Location:** `generate_buildkit.py:458-460` (`capacity-type In ["on-demand"]`)
- **Issue:** Both amd64 sizes are `m6id`; AWS ICE is frequently family+AZ-wide, so both can be
  unavailable at once → same stall. On-demand-only is retained.
- **Impact:** A family-wide regional crunch still stalls.
- **Note:** On-demand-only is a **deliberate, correct** choice (warm NVMe cache + 135-min build
  drain make spot unsuitable). Multi-size on-demand *does* address the reported single-size ICE.
- **Suggested fix:** Consider a different amd64 family as the real fallback, or document the limit.

### Nits

- [x] **Migration-narrative comments** (repo comment rules): `generate_buildkit.py:280-281`
  ("pre-existing nodes still carry an instance-type taint from the single-size pool"),
  `test_generate_buildkit.py:233` ("no longer pinned…"), pre-existing `test:42` ("now imported…").
  Reword to the durable *why*, drop the history.
- [x] **"Pod too small" raises a raw traceback** instead of a clean `✗` error
  (`generate_buildkit.py:144-148`); still exits 1 under `set -e`, UX wart only.
- [x] **EC2NodeClass `InstanceType` AWS tag = first entry only** (`generate_buildkit.py:548`);
  a 12xl node is tagged `m6id.24xlarge`. Cosmetic.
- [x] **`instance_specs.py:206`** ENI comment `# 8 ENIs x 30 IPs - 8` reads as 232; the value 234
  is correct. Comment nit.
- [x] **`plan_pod_resources` log line** mixes `allocatable_*` from `sized[0]` with a possibly
  smaller-node-derived pod size (`generate_buildkit.py:149-156`); log-only, mildly misleading.

### Docs (most incomplete part — reconcile in the same PR)

- [x] **`osdc/.claude/skills/osdc-deployment/SKILL.md:294`** — "Defaults provide *only*
  `replicas_per_arch: 12`" now self-contradicts (this PR added instance-type maps +
  `autoscaling.enabled` to `defaults.buildkit`). Update prose + yaml example.
- [x] **`osdc-runners-nodepools/SKILL.md:80`** — "script default `m8gd.24xlarge`"; `deploy.sh`
  dropped its inline fallbacks. Default now comes from `defaults.buildkit` (arm64 `m7gd.16xlarge:4`).
- [x] **`osdc-runners-nodepools/SKILL.md:123-132`** — "BuildKit Pod Sizing Algorithm" still
  describes single-type sizing; add the smallest-entry rule.
- [x] **`osdc-runners-nodepools/SKILL.md:53-55`** — "every NodePool emits `instance-type` taint";
  buildkit no longer does. Qualify to runner/GPU pools only. (`docs/node-warmup-and-scheduling-gates.md:166`
  already scopes it correctly — no change there.)
- [x] **`osdc-runners-nodepools/SKILL.md:81`** — amd64 default is now a two-entry map; incomplete.
- [x] **`osdc-runners-nodepools/SKILL.md:88`** — stale cluster list (`arc-cbr-production`,
  `lf-prod-*` don't deploy buildkit). Actual: 3 staging + `meta-prod-aws-ue1` + `meta-prod-aws-ue2`.
- [x] **`docs/prod-cluster-ha-us-west-1.md`** — map format update is fine; module list still lists
  a buildkit module the live uw1 entry doesn't have (pre-existing planning-doc drift).

huydhn:
This fired again in prod today (2026-08-25), same failure mode this PR describes — recording the numbers here since it's a second data point from a cluster that doesn't have this change yet.

`meta-prod-aws-ue2` tripped the OSDC services-status workload panel at **48 unavailable replicas**. That number is entirely `buildkit/buildkitd-amd64` (66 desired / 18 available); every other Deployment and StatefulSet on the cluster was fully available and no DaemonSet crossed the 10% threshold (worst was `hf-cache-mount-gpu8` at 7.7%).

**Not an image-pull problem** — all 48 pods were `Pending`, never scheduled. Unrelated to the Harbor ECR pull-through work on #1035 / #1036.

### Karpenter

From the leader (`karpenter-b7f9f64f8-skfk8`). The container log only retains ~15 min at this volume, so this window is `19:21:37Z → 19:36:23Z`:

| | |
|---|---|
| `created nodeclaim` | 72 |
| `launched nodeclaim` | **1** |
| `failed launching nodeclaim` — all `InsufficientInstanceCapacity` | **71** |
| AZs cited | us-east-2a, 2b, 2c — 71 each |
| `skipping, nodepool requirements filtered out all instance types` | **178** |

```
InsufficientInstanceCapacity: We currently do not have sufficient m6id.24xlarge
capacity in the Availability Zone you requested (us-east-2a) ... (2b) ... (2c).
```

One successful launch out of 72 attempts. The NodePool's own limits were never in play — 960 of 34560 CPU.

### Impact

Stalled at **10 nodes**. At 2 pods/node, the 52 replicas KEDA was asking for needs 26. Oldest pending pod sat **70 minutes** and was still pending when I stopped watching; deployment was oscillating 66 → 56 → 52 with availability flat at 18–20.

Why it stays pinned rather than self-clearing: scale-up is `+100%/15s` with no stabilization, scale-down has a 1200s stabilization window, so desired stays high while real capacity can't follow. `meta-prod-aws-ue1` was healthy at 40/40 throughout.

### One caveat, re: the Known limit in the description

This event is consistent with the premise here, but it does **not** on its own show that this PR would have absorbed it. Every one of the 71 rejections is for `m6id.24xlarge` — no other m6id size was attempted, because there is no second pool to fall through to yet. So there's no evidence either way about whether `m6id.12xlarge` had capacity in those AZs at the same moment.

Given ICE is frequently family+AZ-wide, that's worth weighting: this PR is a strict improvement and clearly worth landing, but the second-family follow-up (`m7id.*`) may be what actually covers an event shaped like today's. Nothing to change here — just flagging that the recurrence doesn't validate the m6id-only fallback specifically.

huydhn:
Took all of it; everything below is in the branch now. Restacked onto `main` and dropped #1009 from the stack.

| | |
|---|---|
| **T1-1** | Two weighted NodePools per arch (`buildkit-<arch>` at 100, then −10 per fallback) sharing one EC2NodeClass. Spill is now explicit rather than a price tie. Each pool carries the **full** replica limit, since a spill absorbs the fleet alone. |
| **T1-2** | Inverted as discussed: entry 0 sizes the pod, every later entry is validated by `pods_that_fit` and fails generation otherwise. `m6id.12xlarge: 8` is now a hard error, not a silent fleet-wide shrink. |
| **T2-1** | Moot under the new design — limits are computed **per pool** from that pool's own type, so ordering can't flip them. Comment rewritten to match. Ordering test added. |
| **T2-2** | Both prod overrides deleted; the maps live only in `defaults.buildkit`. 4th copy in `prod-cluster-ha-us-west-1.md` fixed. Preferred deletion over a YAML anchor — an anchor that can't diverge is inheritance with extra syntax. |
| **T2-3** | Three tests on the dict branch: emitted string, key-order preservation, empty-dict → `""`. |
| **T2-4** | DRY taken — `allocatable_resources()` extracted and called from both sites. **Declining the file split:** there's no 400-line rule in this repo (grepped `CLAUDE.md`, `AGENTS.md`, every `SKILL.md`, `pyproject.toml`, the justfile), and siblings run 646–729 lines. Happy to do it as its own PR, but not folded in here. |
| **T3-2** | Documented, in the PR body and `osdc-runners-nodepools/SKILL.md`. A real fallback needs a second NVMe family (`m7id.*`) — the git cache and `instanceStorePolicy: RAID0` both require one. Separate PR. |

**Nits** — all taken. `InstanceType` tag dropped rather than corrected: a static tag can't describe a multi-size pool and Karpenter already sets `node.kubernetes.io/instance-type`.

One correction on `instance_specs.py`: the comment was wrong, but so was the suggested fix — `8 × 30 − 8` is 232, not 234. The actual formula is `8 ENIs × 29 IPs + 2`. Rather than restate it, #1008 makes it code (`eni_max_pods(enis, ips)`) and the prose comments are gone.

**Docs** — all seven fixed, including the `:88` cluster list (actual: `meta-staging-aws-{uw1,ue1,ue2}` + `meta-prod-aws-{ue1,ue2}`).

osdc/modules/buildkit/scripts/python/generate_buildkit.py
```diff
@@ -467,21 +550,28 @@ def _nodepool_block(arch, instance_type, cpu_limit, memory_limit_gi):
...
-    amd64_block = _nodepool_block("amd64", amd64_instance, amd64_cpu_limit, amd64_mem_limit)
+        Karpenter tries NodePools in weight order and only falls through when
+        the one above cannot provision, so the primary entry is what a normal
+        scale-up uses and the rest are reached on ICE or limits. A single pool
+        spanning both sizes would instead pick per pending pod on price, which
+        for price-proportional sizes means the smallest one every time.
+        """
+        blocks = []
+        for i, (instance_type, pods_per_node) in enumerate(plan):
+            name = "buildkit-" + arch if i == 0 else f"buildkit-{arch}-{instance_type.replace('.', '-')}"
```
jeanschmidt, line 569: worth to pay attention when deploying

---

<details>
<summary>🟠 BuildKit NodePool names are now derived from clusters.yaml but the module has no stale-NodePool cleanup, so removing a size orphans its pool (ai-generated section)</summary>

`_arch_blocks` derives a NodePool name per entry of `{arch}_instance_types`, so the set of NodePool objects a cluster owns is now config-dependent (`buildkit-amd64`, `buildkit-amd64-m6id-12xlarge`, ...) where it was a fixed pair. `modules/buildkit/deploy.sh:78` only pipes the generated file through `kubectl_apply_if_changed`, which creates and updates but never deletes, and the generated NodePools carry no labels at all (`generate_buildkit.py:432-433` emits only `metadata.name`). Nothing in the module reconciles removals. `osdc/scripts/kubectl-apply.sh:23-75` is diff-then-apply with no prune. The sibling module solves exactly this: `osdc/modules/nodepools/scripts/python/generate_nodepools.py:385-386` and `:438-439` stamp every generated NodePool and EC2NodeClass with `osdc.io/module: {module_name}`, and `osdc/modules/nodepools/deploy.sh:126-177` ("Step 3: Clean up stale NodePools and EC2NodeClasses") lists `kubectl get nodepools.karpenter.sh -l osdc.io/module=${MODULE_NAME}` and deletes anything absent from the generated set. `osdc/modules/pypi-cache/kubernetes/nodepool.yaml.tpl:10` carries the same label, and `osdc/justfile:576-588` (`remove-module`) sweeps `nodepool.karpenter.sh` / `ec2nodeclass.karpenter.k8s.aws` by that selector. `test_generate_buildkit.py:763` pins the derived name set: `{"buildkit-arm64", "buildkit-amd64", "buildkit-amd64-m6id-12xlarge"}`. This PR's first deploy is additive and safe. The next edit that drops or reorders an entry — the PR description itself says the real fallback is a different family (`m7id.*`), which would replace `m6id.12xlarge` — leaves `buildkit-amd64-m6id-12xlarge` live on all five buildkit clusters with weight 90, `limits.cpu: 34560`, and the `workload/buildkit-amd64` taint that the pod still tolerates. Karpenter keeps provisioning m6id.12xlarge nodes from a pool that no longer exists in clusters.yaml, and nothing in the repo or in `just remove-module` will delete it. Narrow fix: add `labels:\n osdc.io/module: buildkit` to the generated NodePool (`_nodepool_block`) and EC2NodeClass (`_nodeclass_block`) metadata, matching `generate_nodepools.py:385-386`, then append the stale-resource loop from `modules/nodepools/deploy.sh:130-177` to `modules/buildkit/deploy.sh` after the apply at line 78, using the names generated into `nodepools.yaml` as the expected set. The label alone is worth doing regardless: it is what wires the module into `just remove-module` (justfile:576-588), which today also leaves buildkit's NodePools behind.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 24d03baa9.</sub>

</details>

osdc/modules/buildkit/deploy.sh
```diff
@@ -24,16 +24,17 @@ source "$UPSTREAM_ROOT/scripts/mise-activate.sh"
...
+# type to its pods-per-node; several sizes give Karpenter a capacity fallback.
+# pods_per_node stays the default for any entry that omits its count.
 REPLICAS=$(uv run "$CFG" "$CLUSTER" buildkit.replicas_per_arch 4)
-ARM64_INSTANCE=$(uv run "$CFG" "$CLUSTER" buildkit.arm64_instance_type m8gd.24xlarge)
-AMD64_INSTANCE=$(uv run "$CFG" "$CLUSTER" buildkit.amd64_instance_type m6id.24xlarge)
+# No inline fallback: defaults.buildkit in clusters.yaml always supplies these,
+# and a missing value should abort rather than silently deploy a guess.
+ARM64_INSTANCE=$(uv run "$CFG" "$CLUSTER" buildkit.arm64_instance_types)
+AMD64_INSTANCE=$(uv run "$CFG" "$CLUSTER" buildkit.amd64_instance_types)
 PODS_PER_NODE=$(uv run "$CFG" "$CLUSTER" buildkit.pods_per_node 2)
```
jeanschmidt, line 35: maybe worth a small cleanup

---

<details>
<summary>🟡 deploy.sh keeps reading the inert buildkit.pods_per_node key while the two per-arch keys that actually worked are now read by nothing (ai-generated section)</summary>

This change deletes the `buildkit.amd64_pods_per_node` / `buildkit.arm64_pods_per_node` reads and the `--{amd64,arm64}-pods-per-node` arguments from deploy.sh, but leaves the `buildkit.pods_per_node` read on this line and still passes `--pods-per-node`. For any map-shaped value — the only shape the docs describe — `cluster-config.py:138` emits `type:count` for every entry, and `parse_instance_plan` (generate_buildkit.py:123) only consults `default_pods_per_node` for an entry that omits its count. So the surviving knob cannot change the output, while the two knobs that could are now read by nobody. `buildkit.pods_per_node` does not exist anywhere in `osdc/clusters.yaml` (grep for `pods_per_node` in that file returns only the two new `*_instance_types` maps at lines 89 and 92), so `PODS_PER_NODE` is always the inline `2`. Grepping `amd64_pods_per_node|arm64_pods_per_node` across the repo now finds hits only inside `generate_buildkit.py` (the argparse flags at lines 659-660 and the `x if x is not None else pods_per_node` normalisations at 188-189 and 404-405) and in tests — no caller supplies them. Meanwhile `osdc/.claude/skills/osdc-deployment/SKILL.md` still carries the unchanged line "Per-arch override keys are separate (`amd64_*` and `arm64_*`) — there is no flat `pods_per_node` key.". A maintainer who wants to change packing density reads deploy.sh, sees `buildkit.pods_per_node`, sets it (or re-adds `buildkit.amd64_pods_per_node`, which worked before this PR) in clusters.yaml, redeploys, and the generated manifests are byte-identical — no error, no warning, no effect. They lose an afternoon before finding that the count now lives inside the `*_instance_types` map. Delete the dead plumbing: drop the `PODS_PER_NODE` read and the `--pods-per-node` argument from deploy.sh, drop `--amd64-pods-per-node` / `--arm64-pods-per-node` (generate_buildkit.py:659-660) and the `pods_per_node` / `{arch}_pods_per_node` parameters from both generator signatures, and make `parse_instance_plan` require an explicit count. If the bare-type form is worth keeping for hand runs, keep `--pods-per-node` but have deploy.sh stop reading `buildkit.pods_per_node` and instead abort when a cluster still sets `buildkit.{amd64,arm64}_pods_per_node`, so a stale key fails loudly rather than silently.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 24d03baa9.</sub>

</details>

osdc/modules/buildkit/scripts/python/generate_buildkit.py
```diff
@@ -467,21 +550,28 @@ def _nodepool_block(arch, instance_type, cpu_limit, memory_limit_gi):
     Name: "CLUSTER_NAME_PLACEHOLDER-buildkit-{arch}"
     ManagedBy: "karpenter"
     NodePool: "buildkit-{arch}"
```
jeanschmidt: a non-required cleanup that might be worth considering

---

<details>
<summary>🟡 The EC2NodeClass `NodePool` tag is hardcoded to `buildkit-&lt;arch&gt;` and no longer names the pool that launched a fallback-size node (ai-generated section)</summary>

`_nodeclass_block` now emits one EC2NodeClass shared by every NodePool of an arch, but the `NodePool` tag it stamps on launched instances is still the fixed string `buildkit-{arch}`. Nodes provisioned by the fallback pool `buildkit-amd64-m6id-12xlarge` (named at line 566) will carry `NodePool=buildkit-amd64`, a value that does not correspond to the NodePool that created them. The same hunk removes `InstanceType: "{instance_type}"` for exactly this reason, and `test_nodeclass_has_no_instance_type_tag` (test_generate_buildkit.py:799) pins that removal, but `NodePool` was left behind. Before this change the mapping was one NodePool to one EC2NodeClass per arch, both named `buildkit-&lt;arch&gt;`, so the tag was accurate; `_arch_blocks` (line 555) now appends N NodePools plus a single `_nodeclass_block(arch)`, and `test_every_pool_shares_one_nodeclass_per_arch` (test_generate_buildkit.py:782) asserts that sharing. The sibling generator keeps this tag truthful by emitting one node class per pool: `osdc/modules/nodepools/scripts/python/generate_nodepools.py:496` writes `NodePool: "{name}"` using the pool's own name. Nothing in this repo reads the tag, so the impact is confined to AWS-side tooling. Anyone filtering EC2 instances or a cost report by `tag:NodePool` sees every amd64 buildkit node attributed to `buildkit-amd64`, so `tag:NodePool=buildkit-amd64-m6id-12xlarge` matches nothing and a spill onto the fallback size is invisible in tag-based views — which is precisely the event this PR exists to make happen. Remove the `NodePool: "buildkit-{arch}"` line from `_nodeclass_block`; Karpenter already stamps `karpenter.sh/nodepool=&lt;pool name&gt;` on every instance it launches, and that value is accurate for all pools. (Keeping a truthful custom tag would require one EC2NodeClass per NodePool, which defeats the sharing this change introduces.).

<sub>Reviewed by claude-opus-5[1m] at max effort, against 24d03baa9.</sub>

</details>

osdc/modules/buildkit/scripts/python/test_generate_buildkit.py
```diff
@@ -681,3 +687,118 @@ def test_arch_mismatch_fails(self, tmp_path):
 
             result = generate_buildkit.main()
         assert result == 1
+
+
+class TestMultipleInstanceTypes:
+    """A NodePool spanning several instance sizes, with one fixed pod size."""
```
jeanschmidt: nit

---

<details>
<summary>🟡 `TestMultipleInstanceTypes`'s docstring describes one NodePool spanning several sizes, the design the change deliberately rejected (ai-generated section)</summary>

The class docstring states the generator produces "A NodePool spanning several instance sizes". The implementation does the opposite: `_arch_blocks` (generate_buildkit.py:555-571) emits one NodePool per instance type, each pinned to a single-value `node.kubernetes.io/instance-type` requirement (line 465), differentiated only by `weight`. The class's own tests assert that. `test_one_weighted_nodepool_per_type` at test_generate_buildkit.py:763 asserts `set(pools) == {"buildkit-arm64", "buildkit-amd64", "buildkit-amd64-m6id-12xlarge"}` and at 772-773 that each pool lists exactly one type. The `_arch_blocks` docstring at generate_buildkit.py:560-562 names the spanning pool as the wrong approach: "A single pool spanning both sizes would instead pick per pending pod on price, which for price-proportional sizes means the smallest one every time.". The docstring is the first thing a reader of this test class sees, and it names the exact configuration the change exists to avoid. A later maintainer 'simplifying' the generator toward one pool with two values in `node.kubernetes.io/instance-type` would reintroduce the price-based per-pod selection that yields a fleet of 1-pod m6id.12xlarge nodes, with these tests' stated purpose appearing to endorse it. Replace with something matching the code, e.g. `"""One weighted NodePool per instance size, all sharing a single pod spec."""`.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 24d03baa9.</sub>

</details>

osdc/scripts/test_cluster_config.py
```diff
@@ -431,6 +435,39 @@ def test_bool_false_field_output(self, capsys):
...
+        """Dict values print as key:value pairs for generate_buildkit's parser."""
+        code = run_main("production", "buildkit.amd64_instance_types")
+        assert code == 0
+        out = capsys.readouterr().out.strip()
+        assert out == "m6id.24xlarge:2,m6id.12xlarge:1"
+        # Must not look like a Python dict repr
+        assert "{" not in out
+        assert "'" not in out
+
+    def test_dict_value_preserves_yaml_order(self, capsys):
```
jeanschmidt, line 448: maybe a test?

---

<details>
<summary>🟡 test_dict_value_preserves_yaml_order patches out load_config, so the YAML parse step that the load-bearing ordering depends on is untested (ai-generated section)</summary>

The description says "Order is load-bearing and tested." This test is named and documented as proving that clusters.yaml document order survives, but at line 460 it replaces `cluster_config.load_config` with a lambda returning a hand-built Python dict, so `yaml.safe_load` is never called. It verifies that `resolve()` and the new dict branch preserve an already-ordered dict; it does not verify that YAML parsing produces that order in the first place. Line 460: `with patch.object(cluster_config, "load_config", lambda *a, **k: reversed_cfg):`. The only tests that go through a real file are in `TestWithRealYaml` (line 477), whose fixture at line 483 writes `yaml.dump(FAKE_CONFIG, default_flow_style=False)` — `sort_keys` defaults to True, which would emit `m6id.12xlarge` before `m6id.24xlarge` — and that class never reads the instance-type map. The parse itself is at osdc/scripts/cluster-config.py:34-36. If the map order is ever reordered anywhere in the clusters.yaml -&gt; cluster-config.py -&gt; deploy.sh -&gt; generate_buildkit.py chain, generation still succeeds silently: a 12xlarge-first plan sizes the pod at 42 vCPU / 156Gi, the 24xlarge fallback holds 2 of those so `plan_pod_resources` passes, and `buildkit-amd64` at weight 100 becomes the 12xlarge pool — exactly the "fleet of 1-pod 12xl nodes" outcome the PR says it is designed to avoid, with no test failing. Make the test go through YAML: write `reversed_cfg` to a tmp file with `yaml.dump(reversed_cfg, sort_keys=False)`, point `cluster_config.CONFIG_PATH` at it via monkeypatch (the pattern already used at line 484), drop the `load_config` patch, and keep the same output assertion. That covers the parse step end to end.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 24d03baa9.</sub>

</details>

jeanschmidt (approved):
## Review summary

🟠 1 major · 🟡 4 minor

- 🟠 BuildKit NodePool names are now derived from clusters.yaml but the module has no stale-NodePool cleanup, so removing a size orphans its pool — `osdc/modules/buildkit/scripts/python/generate_buildkit.py:566`
- 🟡 deploy.sh keeps reading the inert buildkit.pods_per_node key while the two per-arch keys that actually worked are now read by nothing — `osdc/modules/buildkit/deploy.sh:35`
- 🟡 The EC2NodeClass `NodePool` tag is hardcoded to `buildkit-&lt;arch&gt;` and no longer names the pool that launched a fallback-size node — `osdc/modules/buildkit/scripts/python/generate_buildkit.py:552`
- 🟡 `TestMultipleInstanceTypes`'s docstring describes one NodePool spanning several sizes, the design the change deliberately rejected — `osdc/modules/buildkit/scripts/python/test_generate_buildkit.py:693`
- 🟡 test_dict_value_preserves_yaml_order patches out load_config, so the YAML parse step that the load-bearing ordering depends on is untested — `osdc/scripts/test_cluster_config.py:448`

<details>

**What this changes**

The buildkit generator stops taking one instance type per arch and takes an ordered `type: pods_per_node` map instead, hoisted into `defaults.buildkit` in `osdc/clusters.yaml` so all five buildkit clusters inherit it. Each map entry becomes its own Karpenter NodePool — the first named `buildkit-&lt;arch&gt;` at weight 100, later ones suffixed with the instance type at weight 100 minus 10 per rank — and all pools for an arch now share a single EC2NodeClass. One Deployment still means one pod spec: the first entry sizes it, and `plan_pod_resources` fails generation if a later entry declares more pods than it can hold. Both arches drop the per-size `instance-type` node label and taint, the pod's toleration becomes `Exists`, and `m6id.12xlarge` joins `INSTANCE_SPECS`/`ENI_MAX_PODS`. `cluster-config.py` grows a dict branch rendering `key:value,key:value`.

**What it gets right**

The pod-sizing invariant is guarded where it can actually be violated: `plan_pod_resources` rejects a fallback that over-declares, with an error naming both remedies, and rejects a primary split so far it yields a sub-1-vCPU pod. `test_pod_size_comes_from_first_type` pins the thing most likely to break silently — that adding the 12xlarge leaves the existing 42 vCPU / 155Gi amd64 pod byte-identical. `test_each_pool_limit_absorbs_the_full_replica_count` encodes the reason each pool carries the whole replica limit. The toleration change to `Exists` is the correct guard for pods that must still land on already-tainted nodes.

**Where the risk is**

The findings concentrate in two lines of `generate_buildkit.py`. Four separate reviewers landed on line 566, the pool-naming loop: NodePool names are now config-derived, `deploy.sh` only applies and never prunes, and the generated pools carry no `osdc.io/module` label, so `just remove-module` cannot see them either — those four are one edit, and the sibling `modules/nodepools` module already has exactly this cleanup to copy. Four more landed on line 552, where the shared EC2NodeClass still tags instances `NodePool: "buildkit-&lt;arch&gt;"` — the same staleness the change fixed one line down by deleting the `InstanceType` tag. Three converged on line 678, where `main()` pre-validates the plan with `--pods-per-node` while the generators parse it with the per-arch override.

One class of mistake recurs: an object that used to be one-per-arch became N-per-arch, and several places kept the old assumption — the tag, the missing prune, the test class docstring that still describes a single spanning pool. A second recurs in the config plumbing left half-migrated: `deploy.sh` still reads the now-inert `buildkit.pods_per_node`, the per-arch `*_pods_per_node` keys have no reader at all, and the arm64 argparse help still documents a bare instance type.

Two findings compound. Preference order is carried only by YAML mapping key order, and the two current keys sort in exactly the wrong order alphabetically; an accidental sort would both flip which size is primary and — with no prune — strand the old fallback pool at weight 90.

**Before merge, and after**

Fix now, because each is small relative to its blast radius: the stale-NodePool cleanup in `osdc/modules/buildkit/deploy.sh` (the label plus the sweep from `modules/nodepools/deploy.sh`), and the rollout timeout — "the next deploy serially rolls every BuildKit pod" turns on `--timeout=15m` at `deploy.sh:121` sitting under an 8100s termination grace, which is a one-line change. Settle the schema question now too, even if the answer is "keep the map": "instance preference order is carried by YAML mapping key order" in `osdc/clusters.yaml` is a design decision that gets more expensive after the format ships.

Reasonable as follow-up: the hardcoded `NodePool` tag in `_nodeclass_block`, the `main()` validation default, the dead `pods_per_node` plumbing and orphaned per-arch keys, the duplicated usable-resource derivation in `pods_that_fit`, the arm64 help text, the test-class docstring and inline YAML parsing, and the ENI derivation comments dropped from the unrelated pypi-cache entries in `instance_specs.py`. The abort for a downstream `clusters.yaml` lacking the new keys is a judgment call about whether out-of-repo consumers exist; it fails loudly and before any cluster mutation.

</details>

jeanschmidt:
overall I believe those are all nits, feel free to decide to act or not on them

huydhn:
Thanks — all five taken. Two of them turned out to matter more than the labels suggest.

**🟠 Stale NodePools.** Confirmed, and worse than orphaned objects: this PR also drops the pod's per-size pin, so the nodeSelector is now `workload-type` + `kubernetes.io/arch` and the `instance-type` toleration is `Exists`. An orphaned pool isn't inert — it stays a valid provisioning target and keeps launching a size that is no longer configured. #1008 makes that concrete: it replaces staging's amd64 plan, so `buildkit-amd64-m6id-12xlarge` would survive at weight 90 and could put 48-vCPU nodes under staging's 7-vCPU pods, on exactly the ICE path the fallback exists for.

Fixed as you suggested: generated NodePools and EC2NodeClasses carry `osdc.io/module: buildkit`, and `deploy.sh` sweeps anything with that label absent from the manifest. Buildkit emits one combined `nodepools.yaml` rather than a file per pool, so names are extracted from the manifest instead of from filenames; the sweep refuses to run on an empty expected set, so a generator failure cannot delete the module. This also makes the pools visible to `just remove-module`. New test asserts every generated object carries the label.

**🟡 `NodePool` tag** — dropped rather than made per-pool, same reasoning as `InstanceType`: one class is shared by every pool of an arch, so it cannot name one. Pinned by `test_nodeclass_has_no_nodepool_tag`.

**🟡 Docstring** — replaced with your wording.

**🟡 `pods_per_node` plumbing** and **🟡 the YAML-order test** — agreed on both, taking them as follow-ups; neither changes generated output. On the ordering one, your point about the two keys sorting the wrong way is the sharp part: it works today only because PyYAML preserves document order, and nothing pins that. Better fixed with a test that goes through a real parse than with a comment.

osdc/modules/buildkit/scripts/python/test_generate_buildkit.py
huydhn: Fixed

osdc/scripts/test_cluster_config.py
huydhn: Fixed

osdc/modules/buildkit/scripts/python/generate_buildkit.py
huydhn: Yeah, this has been fixed

osdc/modules/buildkit/scripts/python/generate_buildkit.py
huydhn: Yeah, this looks fixed. I ask Claude to verify that on staging

PR #1008, opened 2026-08-13
buildkit: run staging on a small instance pool

huydhn:
* __->__ #1008

Staging has sat at its warm baseline for 48 days with KEDA `ACTIVE=false`
throughout — one m6id.24xlarge plus one m7gd.16xlarge per cluster, three
clusters, ~$20k/month of idle capacity.

The integration test builds `FROM alpine:3.21`, so pod size is irrelevant to
it; it only needs 8 active backends per arch inside an 8 minute deadline. So
staging takes prod's big+half shape on much smaller instances:

    amd64_instance_types:      arm64_instance_types:
      m6id.4xlarge: 2            m7gd.4xlarge: 2
      m6id.2xlarge: 1            m7gd.2xlarge: 1

That is a **7 vCPU / 25Gi** pod, two per 4xlarge and one per 2xlarge, so the
8-pod burst is 4 nodes per arch — prod's count for amd64, double for arm64
(prod fits 4 on a m7gd.16xlarge). **Idle cost across the three clusters drops
from ~$19,956 to ~$4,183 a month.**

Baseline is 2 pods per arch, not 0: KEDA scales on
`haproxy_backend_current_sessions`, and with no pods the backend has no
servers, so the metric can never leave 0 and the pool would never wake. The
second pod is free on both arches — the primary pool provisions a 2-pod
4xlarge either way — and it matters beyond utilisation: at 1 replica,
`maxSurge: 0` / `maxUnavailable: 1` takes the backend to zero on every
pod-template rollout.

`max_pods` is no longer transcribed: entries are `eni_max_pods(enis, ips)`, and
737 is that helper too — `eni_max_pods(15, 50)`, the topology every
16xlarge-and-larger shares — not a separate ceiling as an earlier revision of
this comment claimed.

`{amd64,arm64}_max` go to 12. At max == the test's target of 8 there was zero
headroom, so one lagging pod left 7/8 and failed on the deadline.

Staging still has no KEDA `fallback` — real, but orthogonal to the pool shape
this PR changes, so it goes in its own PR.

**Tradeoff:** staging no longer exercises prod's 42 vCPU / 155Gi pod on a real
24xlarge. arm64 also needs 4 cold nodes against prod's 2, through a
single-replica Harbor with no pre-mirrored buildkit image — if the deadline
turns out flaky, a denser arm64 primary is the fix, at roughly +$2k/month.


jeanschmidt:
## Findings

### Tier 1 — turn these knobs before merge (all cheap; they protect the CI signal)

#### [T1-1] The arm64 8-minute deadline is flaky by construction
- **Severity:** High
- **Category:** CI reliability
- **Location:** `osdc/clusters.yaml:195-197,206` (staging arm64 map + `arm64_max:8`); `integration-tests/workflows/build-image.yaml` (`DEADLINE=480s`, target 8 backends)
- **Issue:** Five factors stack against arm64: (a) it needs **4 fresh nodes** (2 pods/4xlarge) vs
  prod's **2** (4/m7gd.16xlarge); (b) buildkit/alpine images aren't pre-mirrored → every cold node
  pulls through staging's **1-replica Harbor**, both arches bursting at once (up to 8 cold nodes on
  one Harbor); (c) buildkit pods don't tolerate the node-init startup taints → each fresh node must
  finish DaemonSet init + the cold image-pull race first; (d) **`max=8` equals the target of 8 →
  zero headroom**, so one lagging/ImagePullBackOff pod leaves 7/8 and the build fails at the
  deadline; (e) `min=1` means *every* run is cold. Median cold path ≈ 6–6.5 min (fits), but the tail
  realistically breaches 480s. `"4 small ≈ 2 large"` holds for raw EC2 provisioning but **not** for
  per-node init + image-pull through a single Harbor.
- **Impact:** Intermittent red on the buildkit PR check → merge-queue friction; recurs every run.
- **Suggested fix (most reliable first):** `arm64_min: 2` (warmer start → fewer cold nodes per
  run; also fixes T1-1), and/or a denser arm64 type so 8 pods fit in ≤2 nodes, and/or pre-mirror
  the buildkit image / add a Harbor replica to cut the cold-pull tail. Raising `arm64_max` above 8
  helps only if KEDA's desired-replica count can transiently exceed the 8-backend target (retry
  spikes) — the metric math otherwise caps desired at ~8, so verify the ScaledObject `metricType`
  before relying on a max bump.

#### [T1-2] "The formula is unit-tested" is overstated; a missing ENI entry is uncaught
- **Severity:** High (test quality; constants themselves are correct)
- **Category:** Test coverage
- **Location:** `osdc/scripts/python/instance_specs.py:212-215`; `generate_buildkit.py:86,164` (`ENI_MAX_PODS.get(type, vcpu)`); `test_generate_buildkit.py` (TestStagingSmallPool)
- **Issue:** The `ENIs×(IPs−1)+2` formula lives only in a comment; `234`/`58` are hand-typed
  literals. Mutation testing: changing the big-node value `234→58` still passes all tests; deleting
  the new `ENI_MAX_PODS` entries entirely also still passes (code silently falls back to `vcpu`,
  which rounds to the same 25Gi). The most likely future mistake — add a type to `INSTANCE_SPECS`,
  forget the ENI entry — is uncaught.
- **Impact:** No real protection against a wrong big-node `max_pods` or a missing ENI entry.
- **Suggested fix:** add a consistency guard (`assert set(buildkit_types) <= ENI_MAX_PODS.keys()`)
  plus a real derivation test that recomputes the formula rather than asserting the literal.

### Nits

- [x] **`clusters.yaml:191` (also `:279,:369`)** — comment "4 nodes per arch, the same node count
  prod's pool uses today" is **false for arm64** (prod reaches 8 in 2 nodes; staging needs 4), and
  it's wrong on exactly the risky arch. The PR *body* gets this right; the comment doesn't.
- [x] **`clusters.yaml:83`** — stale `defaults.buildkit` comment "staging inherits" (staging now
  pins explicitly).
- [x] **No KEDA `fallback` on staging** (prod sets 32/8) — a brief LB metrics-endpoint outage
  freezes the HPA at min=1 with no self-heal (distinct from T1-1, which is a legit 0 read).
- [x] **`test_generate_buildkit.py:793`** — temporal assert message "…as prod's pool does today".
- [x] **`instance_specs.py:205`** — pre-existing `r7i.2xlarge=56` inconsistent with the correct
  `*.2xlarge=58` (out of scope; the PR's own values are right).
- [x] **Pre-existing doc drift** #1008 didn't fix: `osdc-deployment/SKILL.md:294,302,305`,
  `docs/observability-estimates.md:242` (`replicas_per_arch:5` vs actual 12).

huydhn:
Both taken. Rebased onto the rewritten #1007 and onto `main`.

**T1-1** — `arm64_min: 1 → 2` (free: with weighted pools the warm pod already provisions a 4xlarge and the second fits on it), both maxes `8 → 12`, and `fallback: 4` on both arches.

I verified the caveat you attached to the max bump, and it holds up. The ScaledObject sets no `metricType`, so KEDA defaults to `AverageValue` with `targetValue: 1` — desired ≡ backend session count. The test runs **9** concurrent builds, not 8: the 8-replica `scale` matrix plus the `build` job, which the workflow calls out as deliberate over-subscription. So desired is 9, `max: 8` was genuinely the binding constraint, and at 12 a lagging pod still leaves 8 active servers to satisfy `TARGET: 8`. Server-templates (`arm 40`, `amd 400`) clear the new max.

**Not** taking the denser arm64 type yet — `m7gd.8xlarge:4` would match prod's 2 nodes but spends ~$2k/month of a ~$15.8k saving on a risk we haven't observed. Written into the PR body as a one-line change if the deadline proves flaky.

**T1-2** — your mutation test reproduced. Values are now computed by `eni_max_pods(enis, ips_per_eni)` rather than transcribed, plus the guard you asked for: every BuildKit type must have an entry, a derivation test that recomputes, and one pinning that the 4xlarge and 2xlarge values differ. `737` stays a literal — it's the AMI's prefix-delegation ceiling, not a product of the formula.

**Nits** — comment rewritten to name arm64 and the single-replica Harbor cold-pull path; `defaults.buildkit` "staging inherits" already went with #1007's dedup; KEDA fallback added; temporal assert reworded. `r7i.2xlarge: 56` left alone as out of scope, but it now carries a one-line note that it deviates from the formula so the next reader doesn't 'fix' it.

**Docs** — `observability-estimates.md` corrected here (`replicas_per_arch: 5 → 12`, noting KEDA makes the live count a range); the `osdc-deployment/SKILL.md` ones are in #1007.

huydhn:
One walk-back on the above: I've **removed the KEDA `fallback: 4`** from this PR.

Your nit stands and I'm not disputing it — staging genuinely has no fallback, and a brief LB metrics-endpoint outage will freeze the HPA at min with no self-heal. But it's orthogonal to what this PR actually changes (the instance pool shape), and I'd rather keep this one minimal. It'll come as its own PR.

Nothing else changed — the delta is 6 lines of `clusters.yaml`. `{amd64,arm64}_max: 12` and `arm64_min: 2` stay, since those are the T1-1 deadline fix and are tied to the pool shape.

osdc/clusters.yaml
```diff
@@ -190,12 +190,23 @@ clusters:
       scheduler_name: bin-pack-scheduler
       capacity_aware_fresh_multiplier: 1.0
     buildkit:
+      # Prod's shape on small instances; the integration test only builds
+      # alpine, so pod size is irrelevant. arm64 is the one to watch: 4 cold
+      # nodes to prod's 2, all pulling through a single-replica Harbor.
+      amd64_instance_types:
```
jeanschmidt, line 196: note for the deployment, we might need some manual cleanup or a script

---

<details>
<summary>🟠 Replacing staging's amd64 instance types strands the buildkit-amd64-m6id-12xlarge NodePool, which the buildkit deploy path never deletes (ai-generated section)</summary>

The three staging clusters stop inheriting `defaults.buildkit.amd64_instance_types` (`m6id.24xlarge: 2`, `m6id.12xlarge: 1`) and declare `m6id.4xlarge`/`m6id.2xlarge` instead. `generate_buildkit.py::_arch_blocks` names every non-primary pool after its instance type, so the generated set changes from {buildkit-amd64, buildkit-amd64-m6id-12xlarge} to {buildkit-amd64, buildkit-amd64-m6id-2xlarge}. `modules/buildkit/deploy.sh` only applies the generated file; it has no stale-resource sweep, so `buildkit-amd64-m6id-12xlarge` stays live in each cluster with nothing in the repo describing it any more. Naming: `osdc/modules/buildkit/scripts/python/generate_buildkit.py:566` — `name = "buildkit-" + arch if i == 0 else f"buildkit-{arch}-{instance_type.replace('.', '-')}"`. No pruning: `osdc/modules/buildkit/deploy.sh:78` pipes into `kubectl_apply_if_changed`, and `osdc/scripts/kubectl-apply.sh:23-75` runs only `kubectl diff` + `kubectl apply` (no `--prune`); grepping `osdc/justfile` for "buildkit" returns no recipe. The sibling module does prune: `osdc/modules/nodepools/deploy.sh:136-156` lists `kubectl get nodepools.karpenter.sh -l osdc.io/module=${MODULE_NAME}` and issues `kubectl delete nodepool.karpenter.sh` for every name not in the generated set. That staging previously produced the 12xlarge pool follows from `osdc/scripts/cluster-config.py:39-49`, whose `resolve()` falls back per-key to `defaults`, plus `osdc/clusters.yaml:89-91`. The stranded pool still matches the buildkitd-amd64 pods exactly — template label `workload-type: buildkit`, taint `workload/buildkit-amd64`, and `nodeClassRef: buildkit-amd64` (still generated) — at weight 90, the same weight `_arch_blocks` gives the new `buildkit-amd64-m6id-2xlarge`; its limits were computed from the old `amd64_max: 8` at 1 pod/node, i.e. `max_nodes = 16`, `cpu: "768"`, `memory: "3072Gi"` per `_nodepool_limits` at generate_buildkit.py:410-424. Each of meta-staging-aws-uw1/ue1/ue2 keeps a Karpenter NodePool that no file in the repository describes and that no deploy will ever remove. It remains a valid provisioning target for buildkitd-amd64, so whenever m6id.4xlarge cannot be provisioned (insufficient-capacity — the exact condition the fallback pool exists for) Karpenter can fall through to it and launch a 48-vCPU m6id.12xlarge to host a single 7-vCPU pod, up to 16 of them, which is the idle spend this change is removing. It surfaces only as an unexplained instance type on a staging bill or in `kubectl get nodes`, with no config to trace it back to. Narrow: after deploying, run `kubectl delete nodepool.karpenter.sh buildkit-amd64-m6id-12xlarge` once against each of meta-staging-aws-{uw1,ue1,ue2}, and record it as a deploy step in the PR. Broader, and what I would pick: port the stale sweep from `osdc/modules/nodepools/deploy.sh:126-156` into `osdc/modules/buildkit/deploy.sh` right after the NodePool apply, adding `labels: {osdc.io/module: buildkit}` to the NodePool metadata in `_nodepool_block` so the deployed set can be listed the same way, then deleting any labelled pool whose name is not in the freshly generated `nodepools.yaml`. That also covers the next instance-type change instead of only this one.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 275501027.</sub>

</details>

osdc/clusters.yaml
```diff
@@ -190,12 +190,23 @@ clusters:
...
+        m7gd.4xlarge: 2
+        m7gd.2xlarge: 1
       autoscaling:
         enabled: true
-        amd64_min: 2  # warm baseline: 2 pods (1x 24xlarge or 2x 12xlarge)
-        amd64_max: 8
-        arm64_min: 4  # 1x m7gd.16xlarge (4 pods/node)
-        arm64_max: 8
+        # Not 0: the session metric can never leave 0 with no backends behind it.
+        amd64_min: 1
```
jeanschmidt: worth having a look if not overbearing staging

---

<details>
<summary>🟠 Staging's amd64 warm baseline of 1 pod strands half of the m6id.4xlarge the primary NodePool provisions, in all three clusters (ai-generated section)</summary>

In each staging cluster the amd64 warm baseline is set to `amd64_min: 1`, but the amd64 primary NodePool is pinned to a single instance type, `m6id.4xlarge`, which this same change sizes at 2 pods of 7 vCPU / 25Gi. Karpenter's `consolidationPolicy: WhenEmpty` never right-sizes a non-empty node, and the pool's `node.kubernetes.io/instance-type In ["m6id.4xlarge"]` requirement offers no smaller node, so the second pod slot on that node is provisioned, paid for, and permanently empty. Three lines below, `arm64_min: 2` applies the opposite reasoning to the structurally identical m7gd.4xlarge. The same three lines appear at clusters.yaml:291 (meta-staging-aws-ue1) and clusters.yaml:392 (meta-staging-aws-ue2). `amd64_min: 1` (clusters.yaml:205) sits directly under `amd64_instance_types: m6id.4xlarge: 2` (clusters.yaml:197). The PR's own new test asserts the node holds two: `assert pods_that_fit(instance_type, res["cpu"], res["memory_gi"]) == declared` for `("m6id.4xlarge", 2)` at test_generate_buildkit.py:821, with the pod fixed at (7, 25) by the assertion on line 819. Recomputing through `allocatable_resources`/`compute_pod_resources` (generate_buildkit.py:73-107) with the new `ENI_MAX_PODS["m6id.4xlarge"] = eni_max_pods(8, 30) = 234` (instance_specs.py:219): kubelet reserve is 110m / 2929Mi, giving 15,590m usable CPU and 57,251 MiB usable memory after the 300m/440Mi DaemonSet overhead. One 7 vCPU / 25Gi pod consumes 7,000m / 25,600 MiB, leaving 8,590m and 31,651 MiB — 55% of the node — which nothing else can use because the pool taints it `workload/buildkit-amd64=true:NoSchedule` (generate_buildkit.py:472-475) and only buildkitd-amd64 tolerates it. `consolidationPolicy: WhenEmpty` is at generate_buildkit.py:442 and the instance-type pin at generate_buildkit.py:463-465. The adjacent `arm64_min: 2 # both warm pods fit the 4xlarge the primary pool provisions anyway` (clusters.yaml:208) states exactly this argument, applied. Each of the three staging clusters permanently runs a m6id.4xlarge at 45% allocation to hold its single warm amd64 backend — about 8.6 vCPU and 31 GiB of committed, unschedulable capacity per cluster, roughly a quarter of the ~$4,183/month baseline the description quotes — and gets no warm capacity in return. At burst the first additional amd64 backend must also be created rather than already being up. Narrow (what I would take): set `amd64_min: 2` at clusters.yaml:205, 291 and 392, matching `arm64_min: 2` — the m6id.4xlarge is provisioned either way, so the second warm backend costs nothing and does not touch the cold-start budget. Broader: make `m6id.2xlarge: 1` the primary and `m6id.4xlarge: 2` the fallback, so a lone warm pod sits on a 2xlarge and the amd64 baseline halves; the price is 8 cold 2xlarges instead of 4 cold 4xlarges at the 8-backend burst, against the 8-minute deadline the description already names as the risk.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 275501027.</sub>

</details>

osdc/clusters.yaml
```diff
@@ -190,12 +190,23 @@ clusters:
...
+        # Headroom over the test's 8-backend target, not a capacity estimate.
+        amd64_max: 12
```
jeanschmidt, line 210: maybe we should fix integration testing to excercice scaling and avoid being surprised by a future regression?

---

<details>
<summary>🟠 Raising staging's buildkit max from 8 to 12 falsifies the integration test's "9 builds contend for a max-8 pool" comments (ai-generated section)</summary>

`osdc/integration-tests/workflows/build-image.yaml` documents the whole design of its `scale` job in terms of a pool that caps at 8. Line 89-90 says "Runs concurrently with `build`, so 9 builds contend for a max-8 pool: the odd one out waits for HAProxy to add capacity, exercising the same autoscaling path", and lines 56-57 justify the 45-attempt retry loop with "when the pool is over-subscribed (9 builds &gt; 8 pods)". Raising `amd64_max`/`arm64_max` to 12 on all three staging clusters (clusters.yaml:207, 209, 293, 295, 394, 396) removes the over-subscription entirely — 9 concurrent builds now fit in a 12-pod pool. clusters.yaml:207/209 (and the ue1/ue2 copies at 293/295 and 394/396) change `amd64_max`/`arm64_max` from 8 to 12. `osdc/integration-tests/workflows/build-image.yaml:89` reads "so 9 builds contend for a max-8 pool" and :57 reads "when the pool is over-subscribed (9 builds &gt; 8 pods)". `osdc/integration-tests/scripts/python/phases.py:43` gates the BUILDKIT block on `["arc-runners", "buildkit"]`, which the three staging clusters satisfy (clusters.yaml:210-219, 305-319, 407-423). The prod clusters already run 360/30 (clusters.yaml:536-538), so staging was the last place the "max-8" comment was true. Anyone debugging a BuildKit integration-test failure reads that the 9th build must wait for scale-up and that the 45×15s retry loop exists to outlast that contention. Neither is true on staging any more, so the retry sizing looks unexplained and the stated "exercises the same autoscaling path" rationale for running `scale` concurrently with `build` is silently gone. Update `osdc/integration-tests/workflows/build-image.yaml:89-90` to say 9 builds against a pool that now caps at 12 (so the burst is bounded by scale-up latency, not by the cap), and drop or reword the "(9 builds &gt; 8 pods)" clause at :57 so the retry budget is justified by cold-start time rather than by over-subscription. Narrow alternative: state the cap as "the per-arch `*_max` in clusters.yaml" instead of a literal 8, so the comment stops drifting.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 275501027.</sub>

</details>

osdc/docs/observability-estimates.md
```diff
@@ -239,7 +239,7 @@ From `clusters.yaml` defaults plus `meta-prod-aws-ue2` overrides:
 | Harbor exporter / jobservice / portal / db / redis | 1 each | chart defaults |
 | node-compactor | 1 | single replica |
 | Monitoring Alloy | 2 | Hard-coded in `modules/monitoring/helm/alloy-values.yaml` (`controller.replicas: 2`, `clustering.enabled: true`) — there is no `monitoring.alloy_replicas` config knob |
-| BuildKit daemon | 10 | `defaults.buildkit.replicas_per_arch: 5` × 2 archs |
+| BuildKit daemon | 24 | `defaults.buildkit.replicas_per_arch: 12` × 2 archs; every buildkit cluster runs KEDA, so the live count varies between the per-arch min and max |
```
jeanschmidt, line 242: snakk doc nit

---

<details>
<summary>🟡 observability-estimates.md now states two different BuildKit replica counts for the same clusters.yaml default, 24 at line 242 and 10 at line 8 (ai-generated section)</summary>

The change corrects the input-data table row from 10 (5/arch) to 24 (12/arch), but the stale-estimates callout at line 8 of the same file still asserts "Current `clusters.yaml` defaults give: ... BuildKit 10 (5/arch × 2)". The two statements describe the same fact — `defaults.buildkit.replicas_per_arch` — and now disagree, and line 8 explicitly presents itself as the refreshed, authoritative summary ("The input-data table below has been refreshed"). New line 242 says `defaults.buildkit.replicas_per_arch: 12` × 2 archs = 24. Line 8 of the same file still says "BuildKit 10 (5/arch × 2)". `osdc/clusters.yaml:82` sets `replicas_per_arch: 12`, so line 8 is wrong on both the per-arch figure and the total. Before this change the two lines agreed (both said 10/5-per-arch); the change fixed one and left the other. A reader re-validating the meta-prod-aws-ue2 sizing hits the correction note first, takes 10 as the current default, and either re-"fixes" line 242 back to 10 or carries the wrong figure into a new cardinality estimate. The document's own maintenance note (line 14) makes these numbers load-bearing for Grafana Cloud cost planning. Update line 8 to "BuildKit 24 (12/arch × 2)" to match line 242 and `clusters.yaml:82`. If you also want the downstream C_fixed row at line 261 (`| BuildKit daemon | ~9 | 10 | ~90 |`) to stop contradicting the input table, bump its replica column to 24 and recompute the subtotal — though line 8 already warns that downstream subtotals are stale, so the line 8 fix is the one that matters.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 275501027.</sub>

</details>

osdc/modules/buildkit/scripts/python/test_generate_buildkit.py
```diff
@@ -802,3 +802,61 @@ def test_nodeclass_has_no_instance_type_tag(self):
...
+                assert pods_that_fit(instance_type, res["cpu"], res["memory_gi"]) == declared
+
+    def test_reaches_integration_test_target(self):
+        """The buildkit integration test needs 8 active backends per arch."""
+        res = plan_pod_resources([("m6id.4xlarge", 2), ("m6id.2xlarge", 1)])
+        per_node = pods_that_fit("m6id.4xlarge", res["cpu"], res["memory_gi"])
+        assert math.ceil(8 / per_node) == 4, "8 amd64 pods should need 4 nodes"
+
+    def test_nodepool_limits_cover_the_smallest_type(self):
+        """At max replicas an all-2xlarge fleet must stay inside the limits."""
```
jeanschmidt: small docstring nit

---

<details>
<summary>⚪ The new test's docstring claims it checks the max-replica case against the 2xlarge pool; it runs at 8 replicas and reads the 4xlarge pool's limits (ai-generated section)</summary>

Two claims in this one-line docstring do not match the body. (1) "At max replicas": line 831 calls `generate_nodepools_yaml(self.ARM64, self.AMD64, 8, 2)` with `replicas=8`, while the max this PR configures is 12 (`clusters.yaml:207`); the sibling test at line 824 already identifies 8 as the integration test's target, not the max. (2) "an all-2xlarge fleet ... inside the limits": line 834 filters to the NodePool named `buildkit-amd64`, which `_arch_blocks` generates for the *primary* entry `m6id.4xlarge` (generate_buildkit.py:566); the pool that would actually host a 2xlarge fleet is `buildkit-amd64-m6id-2xlarge`, and the loop `return`s before ever seeing it. generate_buildkit.py:565-568 names the i=0 pool `buildkit-&lt;arch&gt;` and each later entry `buildkit-&lt;arch&gt;-&lt;type&gt;`, and pins each pool to one instance type via `node.kubernetes.io/instance-type In ["{instance_type}"]` (lines 463-465), so 2xlarge nodes can only be provisioned by `buildkit-amd64-m6id-2xlarge`. `_nodepool_limits` (generate_buildkit.py:410-424) sizes each pool independently from its own `pods_per_node`, so the 4xlarge pool's limit says nothing about the 2xlarge pool's. clusters.yaml:207/209 set the real max to 12, not 8. The test reads as a guard on the max-replica ceiling and on the smallest fallback pool, and is neither. Raising `amd64_max` above 12, or lowering the 2xlarge fallback pool's limits, leaves this test green while the case it is named for goes unchecked. Either make the docstring match the body ("At the integration test's 8-backend target, the primary amd64 pool's CPU limit covers an all-2xlarge node count"), or make the body match the docstring: pass 12 for replicas and filter on `buildkit-amd64-m6id-2xlarge` instead of `buildkit-amd64`. I would take the second — it is the check the name promises and the one that actually catches a mis-sized fallback pool.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 275501027.</sub>

</details>

osdc/scripts/python/instance_specs.py
```diff
@@ -112,15 +112,27 @@
...
+def eni_max_pods(enis: int, ips_per_eni: int) -> int:
+    """AWS max-pods for an instance: one IP per ENI is the ENI's own, plus 2 host pods."""
+    return enis * (ips_per_eni - 1) + 2
+
+
 # ---------------------------------------------------------------------------
 # EKS max pods per instance type (from ENI limits).
 # Source: awslabs/amazon-eks-ami eni-max-pods.txt
 # The kubelet memory reservation formula uses max_pods, NOT vCPU count.
+# 737 is the AMI's prefix-delegation ceiling, not a product of eni_max_pods().
```
jeanschmidt: Need to verify the claim, might require a comment fix

---

<details>
<summary>⚪ The added comment says 737 is not a product of eni_max_pods(), but eni_max_pods(15, 50) is exactly 737 (ai-generated section)</summary>

The new comment tells the reader that the 737 entries are outside the formula the new `eni_max_pods` helper implements. They are not: `eni_max_pods(15, 50)` = 15 × (50 − 1) + 2 = 737, which is the standard 15-ENI × 50-IP value for the .16xlarge/.24xlarge instances that carry it here. It is also not a prefix-delegation number — the comment two lines above (line 133) correctly cites `awslabs/amazon-eks-ami eni-max-pods.txt`, which is the non-prefix-delegation table. `eni_max_pods` at instance_specs.py:126-128 returns `enis * (ips_per_eni - 1) + 2`; 15 * 49 + 2 = 737. Every other literal in the dict is likewise a product of the same formula: 234 = eni_max_pods(8, 30) (used explicitly at lines 213-214, 218-219, 221), 58 = eni_max_pods(4, 15) (lines 220, 222), 345 = eni_max_pods(7, 50) at line 201, 198 = eni_max_pods(4, 50) at lines 209-210. The 737 entries (lines 216-217, 223-225) are m8gd.24xlarge, m6id.24xlarge, c7gd.16xlarge, m7gd.16xlarge, m8gd.16xlarge — all 15-ENI × 50-IP types. The next person adding a large BuildKit or runner instance type reads that 737 is un-derivable and hand-transcribes it rather than writing `eni_max_pods(15, 50)`, which is precisely the transcription slip the new `test_eni_max_pods_is_derived_not_transcribed` test (test_generate_buildkit.py:850) was added to prevent. A wrong max_pods silently mis-sizes the pod via `kubelet_reserved` (generate_buildkit.py:78). Replace the comment with something true, e.g. `# 737 = eni_max_pods(15, 50) — the 15-ENI x 50-IP ceiling on the largest types.`, and optionally express those five entries as `eni_max_pods(15, 50)` so they are derived like the rest of the dict.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 275501027.</sub>

</details>

jeanschmidt (commented):
We need to check the staging usage and make sure the integration tests are covering any future regressions...

## Review summary

🟠 3 major · 🟡 1 minor · ⚪ 2 nit

- 🟠 Replacing staging's amd64 instance types strands the buildkit-amd64-m6id-12xlarge NodePool, which the buildkit deploy path never deletes — `osdc/clusters.yaml:196`
- 🟠 Staging's amd64 warm baseline of 1 pod strands half of the m6id.4xlarge the primary NodePool provisions, in all three clusters — `osdc/clusters.yaml:205`
- 🟠 Raising staging's buildkit max from 8 to 12 falsifies the integration test's "9 builds contend for a max-8 pool" comments — `osdc/clusters.yaml:207`
- 🟡 observability-estimates.md now states two different BuildKit replica counts for the same clusters.yaml default, 24 at line 242 and 10 at line 8 — `osdc/docs/observability-estimates.md:242`
- ⚪ The new test's docstring claims it checks the max-replica case against the 2xlarge pool; it runs at 8 replicas and reads the 4xlarge pool's limits — `osdc/modules/buildkit/scripts/python/test_generate_buildkit.py:830`
- ⚪ The added comment says 737 is not a product of eni_max_pods(), but eni_max_pods(15, 50) is exactly 737 — `osdc/scripts/python/instance_specs.py:135`

<details>

**What this changes**

Three staging clusters (meta-staging-aws-uw1/ue1/ue2) stop inheriting `defaults.buildkit.{amd64,arm64}_instance_types` and declare their own small-instance plans in `osdc/clusters.yaml` — `m6id.4xlarge:2` + `m6id.2xlarge:1` and `m7gd.4xlarge:2` + `m7gd.2xlarge:1` — which the generator turns into one weighted NodePool per instance type. KEDA bounds move from min 2/4 and max 8/8 to min 1/2 and max 12/12. `osdc/scripts/python/instance_specs.py` gains the four new instance types, a new `eni_max_pods(enis, ips)` helper, and seven max-pods entries rewritten as calls to it. A `TestStagingSmallPool` class is added to `test_generate_buildkit.py`, and one row of `osdc/docs/observability-estimates.md` goes from 10 replicas to 24.

**What it gets right**

`test_declared_packing_holds` pins the derived pod at (7, 25) and then asserts `pods_that_fit` returns the declared count for each instance type in both plans — that is the exact precondition the generator enforces at deploy time, checked for the new shape rather than assumed. The change reuses the existing big+half plan format instead of adding a staging-specific mechanism, so the fallback-pool machinery, weights and limits all come out of the same code path prod uses. The reasoning behind `amd64_min: 1` rather than 0 (the session metric cannot leave 0 with no backends) is recorded inline, next to the value it explains.

**Where the risk is**

The findings concentrate in three places and one recurring habit. Around `clusters.yaml:196` (and its twins at 282 and 383), four reviewers independently reached the same conclusion about orphaned NodePools: pool names are derived from instance type, so replacing the amd64 plan renames the fallback pool, and `osdc/modules/buildkit/deploy.sh:78` applies without pruning — the sibling `osdc/modules/nodepools/deploy.sh` does sweep stale pools, but only by a label buildkit never stamps. Those four share `c-001` and one edit settles all of them.

Around the new test class, two more clusters converge. Four findings target `test_nodepool_limits_cover_the_smallest_type` (`c-005`) — it filters on `buildkit-amd64`, the pool pinned to the 4xlarge, and passes 8 where the configured max is now 12 — and one edit fixes the pool filter, the replica count and the docstring together. Four more target `test_every_buildkit_type_has_an_eni_entry` (`c-006`), whose "every" is a four-string transcription of `clusters.yaml`.

That is the recurring class: this change replaces transcribed max-pods literals with a derived helper, and then introduces fresh transcriptions — the plan strings in the test, the 8 that should be 12, and the 24 in `observability-estimates.md` that now disagrees with the 10 nineteen lines below it and the 10 at line 8 (`c-004`, four findings, one edit). The false claim that 737 is outside `eni_max_pods` (`c-007`, three findings) sits in the same file as the helper and actively steers the next editor back toward transcribing.

One pair compounds. `amd64_min: 1` strands half the m6id.4xlarge the primary pool provisions, which alone is a utilization point; combined with the generated Deployment's unconditional `maxSurge: 0`, the same value takes staging amd64 to zero backends on every future pod-template rollout. Both are `c-002`, and setting min to 2 resolves both.

**Before merge, and after**

Before: decide the disposition of the stranded `buildkit-amd64-m6id-12xlarge` pool in `osdc/clusters.yaml` — once this merges, nothing in the repo names it, so either the one-time delete is recorded as a rollout step or the prune lands in `osdc/modules/buildkit/deploy.sh`. Fix the "737 is the AMI's prefix-delegation ceiling" comment in `osdc/scripts/python/instance_specs.py`; it is one line and it is arithmetically false. Settle `amd64_min: 1` in `osdc/clusters.yaml`, given it strands half a node and interacts with `maxSurge: 0`. Correct the BuildKit replica rows in `osdc/docs/observability-estimates.md` so the document does not state three counts for one service.

After: the test-hygiene group — the wrong pool filter and hardcoded 8 in `test_nodepool_limits_cover_the_smallest_type`, deriving the ENI check from `clusters.yaml`, dropping `test_small_and_large_types_do_not_share_a_max_pods_value`, and relocating the `eni_max_pods` tests next to the module they exercise. Also the stale "9 builds contend for a max-8 pool" comments in `osdc/integration-tests/workflows/build-image.yaml`, which now describe an environment that no cluster runs.

</details>

jeanschmidt (approved):
approving to unblock

huydhn:
Taken, with one correction of mine and one pushback.

**🟠 amd64_min: 1.** You're right, and the `maxSurge: 0` interaction is the part that decides it — at 1 replica the backend goes to zero on every pod-template rollout, for up to the 15m rollout timeout. Now 2 on both arches. It is free: the primary pool provisions a 2-pod 4xlarge either way, so the cost figure is unchanged.

**⚪ 737.** You're right, and I introduced this. `eni_max_pods(15, 50)` is exactly 737 — it is the plain formula for the 15-ENI × 50-IP topology, not a separate ceiling. The comment is corrected and the BuildKit entries now use the helper. I've left the ~30 runner/GPU 737s as literals: converting them would assert an ENI topology for instance types this PR doesn't own, which is the same transcription risk pointed the other way.

**🟠 Stranded `buildkit-amd64-m6id-12xlarge`.** Fixed at the root in #1007 — module label plus a prune sweep in `deploy.sh` — so no manual rollout step is needed. Agreed it had to be settled before merge.

**⚪ Limits test.** Raising this above nit: it wasn't untidy, it was incoherent. It filtered `buildkit-amd64` (the 4xlarge pool) while measuring 2xlarge node counts against it, at 8 replicas when max is 12, so it passed for the wrong reason. Now filters the 2xlarge pool and takes the replica count from a named `AMD64_MAX = 12`.

**🟡 observability-estimates.md.** Line 8 still said 10; reconciled with line 242.

**🟠 "9 builds contend for a max-8 pool".** Comments fixed, but I'd push back on major. The test still passes — what's lost is the *incidental* over-subscription coverage, since 9 sessions now get 9 pods instead of queueing. Worth flagging that this was deliberate test design and raising max removes it, so if we want the queueing path covered it should be an explicit test rather than a side effect of a too-low ceiling. Taking that as a follow-up.

Remaining follow-ups: deriving the ENI check from `clusters.yaml`, relocating the `eni_max_pods` tests next to the module they exercise, and the `pods_per_node` plumbing in #1007.

osdc/clusters.yaml
huydhn: Fixed

osdc/clusters.yaml
huydhn: Fixed

osdc/clusters.yaml
huydhn: Fixed

PR #1009, opened 2026-08-13
image-cache-janitor: reclaim orphaned BuildKit cache directories

huydhn:
* __->__ #1009
* #1008
* #1007

BuildKit pods cache to <cache>/<pod-name> on the instance store. The
directory outlives the pod and no surviving buildkitd knows about it, so
buildkit's own GC can never reclaim it. Two 8-day-old buildkit nodes were
holding 880 GB and 1194 GB there against 2 GB of containerd images — the
janitor was pruning the 2 GB and ignoring the rest.

The janitor already runs on these nodes and already has the containerd
socket, so live pod names come from `crictl pods` with no new RBAC. A
directory is orphaned when no pod of that name is running and it is older
than a grace period.

An empty pod list is treated as "unknown", not "nothing is running": the
janitor is itself a pod on the node, so crictl always reports at least
one, and without that floor a transient containerd fault would read as
"every directory is orphaned".

The cache is read through /proc/1/root rather than a hostPath mount,
which would have created the path on all ~786 nodes this DaemonSet covers
rather than the handful running BuildKit. Directory sizes are memoized so
a multi-hundred-GB orphan is measured once, not on every 300s cycle.

Ships with BUILDKIT_ORPHAN_DRY_RUN=true: it reports orphan count and bytes
as metrics but deletes nothing, so the sweep can be confirmed against real
nodes before it is allowed to remove anything.

### DIFF
--- a/osdc/base/kubernetes/image-cache-janitor/configmap.yaml
+++ b/osdc/base/kubernetes/image-cache-janitor/configmap.yaml
@@ -55,7 +55,12 @@ data:
     # BuildKit pods cache to <BUILDKIT_CACHE_DIR>/<pod-name> on the instance
     # store. A pod's directory outlives the pod, and no surviving buildkitd
     # knows about it, so buildkit's own GC can never reclaim it.
-    BUILDKIT_CACHE_DIR = os.environ.get("BUILDKIT_CACHE_DIR", "/mnt/k8s-disks/0/buildkit-cache")
+    # Read through /proc/1/root (hostPID + privileged) rather than a hostPath
+    # mount: a DirectoryOrCreate mount would create this path on all ~786 nodes
+    # the DaemonSet covers, not just the handful running BuildKit.
+    BUILDKIT_CACHE_DIR = os.environ.get(
+        "BUILDKIT_CACHE_DIR", "/proc/1/root/mnt/k8s-disks/0/buildkit-cache"
+    )
     BUILDKIT_ORPHAN_GRACE_SECONDS = int(os.environ.get("BUILDKIT_ORPHAN_GRACE_SECONDS", "1800"))
     BUILDKIT_ORPHAN_DRY_RUN = os.environ.get("BUILDKIT_ORPHAN_DRY_RUN", "true").lower() == "true"
 
@@ -170,7 +175,14 @@ data:
         A directory is orphaned when no pod of that name is running on the
         node. The grace period covers the window where kubelet has created
         the directory but the sandbox is not yet visible to crictl.
+
+        An empty live_pod_names is treated as "unknown", not "nothing is
+        running": the janitor is itself a pod on this node, so crictl always
+        reports at least one. Returning nothing keeps a transient containerd
+        hiccup from being read as "every directory is orphaned".
         """
+        if not live_pod_names:
+            return []
         return [
             d for d in dirs
             if d.name not in live_pod_names and (now - d.mtime) >= grace_seconds
@@ -313,6 +325,11 @@ data:
             return False
 
 
+    # (dir name, mtime) -> size. Sizing a multi-hundred-GB cache is expensive,
+    # and under dry run the same orphans persist across every cycle.
+    _SIZE_CACHE: dict[tuple[str, float], int] = {}
+
+
     def _dir_size_bytes(path: str) -> int:
         """Best-effort directory size via du. Returns 0 if it fails or times out."""
         try:
@@ -353,8 +370,14 @@ data:
         orphans = select_orphan_cache_dirs(
             _scan_cache_dirs(), live, time.time(), BUILDKIT_ORPHAN_GRACE_SECONDS,
         )
+        live_keys = {(d.name, d.mtime) for d in orphans}
+        for stale in [k for k in _SIZE_CACHE if k not in live_keys]:
+            del _SIZE_CACHE[stale]
         for d in orphans:
-            d.size = _dir_size_bytes(os.path.join(BUILDKIT_CACHE_DIR, d.name))
+            key = (d.name, d.mtime)
+            if key not in _SIZE_CACHE:
+                _SIZE_CACHE[key] = _dir_size_bytes(os.path.join(BUILDKIT_CACHE_DIR, d.name))
+            d.size = _SIZE_CACHE[key]
 
         total = sum(d.size for d in orphans)
         metrics.set("image_cache_janitor_buildkit_orphan_dirs", float(len(orphans)))
--- a/osdc/base/kubernetes/image-cache-janitor/configmap.yaml
+++ b/osdc/base/kubernetes/image-cache-janitor/configmap.yaml
@@ -29,6 +29,7 @@ data:
     import json
     import logging
     import os
+    import shutil
     import socket
     import subprocess
     import sys
@@ -50,6 +51,13 @@ data:
     IMAGE_CACHE_TARGET_GI = int(os.environ.get("IMAGE_CACHE_TARGET_GI", "400"))
     CHECK_INTERVAL_SECONDS = int(os.environ.get("CHECK_INTERVAL_SECONDS", "300"))
     METRICS_PORT = 9103
+
+    # BuildKit pods cache to <BUILDKIT_CACHE_DIR>/<pod-name> on the instance
+    # store. A pod's directory outlives the pod, and no surviving buildkitd
+    # knows about it, so buildkit's own GC can never reclaim it.
+    BUILDKIT_CACHE_DIR = os.environ.get("BUILDKIT_CACHE_DIR", "/mnt/k8s-disks/0/buildkit-cache")
+    BUILDKIT_ORPHAN_GRACE_SECONDS = int(os.environ.get("BUILDKIT_ORPHAN_GRACE_SECONDS", "1800"))
+    BUILDKIT_ORPHAN_DRY_RUN = os.environ.get("BUILDKIT_ORPHAN_DRY_RUN", "true").lower() == "true"
 
     # Convert GiB to bytes
     GI = 1024 ** 3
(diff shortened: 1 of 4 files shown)


jeanschmidt:
I believe the fix for this problem should be fixing how pods reuse the local cache, not keep cleaning (using janitor) or create a trow away local cache per pod. After further investigation I noticed a fundamental problem on buildkit that escaped all our reviews interactions and this is costing money and performance.

So this is the result of chat with my agent:

# Design review — PR #1009 "image-cache-janitor: reclaim orphaned BuildKit cache directories"

| | |
|---|---|
| PR | https://github.com/pytorch/ci-infra/pull/1009 (ghstack top: #1007 ← #1008 ← #1009) |
| Author | huydhn |
| Status | Open, CI green, ships in dry-run (`BUILDKIT_ORPHAN_DRY_RUN=true`) |
| This doc | Design position: the disk problem is real, but the janitor is the wrong long-term fix. Fix the cache design instead. |
| Date | 2026-08-18 |

## TL;DR

- **The disk problem is real and near-term** — orphaned BuildKit cache dirs are the *only* unbounded consumer on the buildkit nodes' NVMe (~110–150 GB/day/node); a long-lived node **wedges in 2–5 weeks** and kubelet's disk-pressure eviction can't save it. Something must fix this.
- **The janitor is a symptom patch.** It's a destructive `rm -rf` sweeper on a cluster-wide (~786-node) DaemonSet that cleans up churn a per-pod cache design keeps creating — forever — and it delivers **zero** build-speed benefit (every replacement pod still starts cold).
- **The root cause is the cache layout**, and fixing it also unlocks the thing we actually want: **a warm node reusing decompressed layers instead of re-pulling them.**
- **Recommended fix: a per-node "slot" cache picked by file lock** (details below). It makes orphans impossible *and* gives cross-generation warm reuse — so the janitor becomes unnecessary.

## The problem

BuildKit pods mount their cache from a node hostPath with a **per-pod subdirectory**:

```yaml
# modules/buildkit/scripts/python/generate_buildkit.py (deployment template)
env:
  - name: POD_NAME
    valueFrom: { fieldRef: { fieldPath: metadata.name } }   # unique per pod
volumeMounts:
  - name: buildkit-cache
    mountPath: /var/lib/buildkit
    subPathExpr: $(POD_NAME)
volumes:
  - name: buildkit-cache
    hostPath: { path: /mnt/k8s-disks/0/buildkit-cache, type: DirectoryOrCreate }
```

Effective path per pod: `/mnt/k8s-disks/0/buildkit-cache/buildkitd-<arch>-<rs-hash>-<rand>`. The directory **outlives the pod** and nothing reclaims it — buildkitd's GC is scoped to its own live root, and **kubelet never GCs hostPath**. Observed: two 8-day nodes holding **880 GB and 1194 GB** of orphaned cache against ~2 GB of images.

### Why it's urgent

- **NVMe:** m6id.24xlarge = 5.7 TB, m7gd.16xlarge = 3.8 TB, RAID0 into one volume at `/mnt/k8s-disks/0`, shared by containerd + kubelet + all buildkit pod caches + git-cache.
- **Leak:** ~110–150 GB/day/node; orphans are the *sole* unbounded component.
- **The safety valve fails:** because orphaned space isn't owned by a live pod, kubelet `DiskPressure` eviction evicts the *live* buildkit pods and frees almost nothing → node carries the taint (can't schedule), isn't empty (so Karpenter `WhenEmpty` won't recycle it), and **wedges**; buildkitd then hits ENOSPC. Time-to-wedge ~3–5 weeks (amd64) / 2–4 weeks (arm64).
- **Nodes never self-heal:** NodePools use `WhenEmpty` + `consolidateAfter: 5m`, **no `expireAfter`**; the warm KEDA baseline keeps ≥1 pod on the node so it never goes empty. Every scale event / rolling update mints a new pod name → new empty dir → *multiplies* orphans.

## Root cause: the cache is keyed on a value that never recurs

Verified against BuildKit source:
- **OCI worker, containerd worker disabled** (`kubernetes/base/configmap.yaml`: `[worker.oci] enabled = true`, `[worker.containerd] enabled = false`) → every pulled/decompressed layer lives under the pod's private `--root`; no shared node-level store.
- **buildkitd takes an exclusive flock on `<root>/buildkitd.lock`** (`cmd/buildkitd/main.go`) → the two pods on a 24xlarge **cannot** share one dir. The per-pod subdir is a *collision-avoidance requirement*, not a sharing mechanism.
- **`POD_NAME = metadata.name` is unique per pod.** A replacement pod gets a new name → a new **empty** dir → it re-pulls and re-decompresses; the old dir is orphaned.

The `subPathExpr:$(POD_NAME)` construct was clearly *reaching for* a persistent per-identity cache — but it's wired to a `Deployment`, whose pod names are random and never recur. So today's design delivers **neither reuse nor cleanup** — the worst of both, and the source of both the leak and the cold rebuilds.

## Why I'm pushing back on the janitor

Not because it's badly built — it's carefully guarded. Because it's the wrong thing to build:

1. **It manages a symptom forever.** The design keeps minting orphans on every scale/rollout; the janitor just keeps sweeping. We never stop *creating* the garbage — we sign up to keep cleaning it indefinitely.
2. **It buys zero build-speed.** Every replacement pod still starts cold and re-pulls/re-decompresses the huge CUDA base layers. The janitor reclaims disk but leaves the real prize — warm nodes — on the table.
3. **Destructive, with a fleet-wide blast radius.** It's `rm -rf` driven from a privileged DaemonSet spanning **~786 cluster nodes** to fix a handful of buildkit nodes. The dry-run default, 30-min grace, "empty crictl list = unknown" floor, and size memoization all exist *because* the operation is dangerous — that's a lot of safety scaffolding for a self-inflicted problem.
4. **Wrong layer / coupling.** It puts buildkit-module-specific knowledge (the on-disk cache path) into a `base/` cluster-wide component. The cache layout now lives in two places — a single-source violation.
5. **Ongoing cost.** ~250 lines of privileged runtime Python to own, test, and operate, plus a metrics/rollout process — versus a small change in the component that actually owns the cache.
6. **It concedes the design flaw instead of fixing it.** The right move is to make the cache dir *recur and get reused*; then there's nothing to sweep.

## The fix: a per-node "slot" cache, picked by file lock

Instead of one dir per (random) pod name, use a **fixed, tiny set of dirs per node** — `slot-0 … slot-(N-1)` — and have each starting pod claim the first slot no live buildkitd is holding. Two ordinary primitives: the Linux `flock` (which buildkitd already takes on its root) + fixed numbered directories. It lives in the pod's own entrypoint (~a dozen lines), not in a fleet DaemonSet.

A slot is "in use" only while a live buildkitd holds its lock. When that pod dies, the kernel releases the lock automatically (crash/OOM/kill included), but **the slot's directory — with its warm cache — stays on disk.** The next pod on that node grabs it and reuses the warm layers.

**Walk-through (one 24xlarge, 2 slots):**
```
t0  Pod A starts → slot-0 free → takes it → pulls CUDA layers (cold) → slot-0 WARM
    Pod B starts → slot-0 locked → takes slot-1 (cold) → slot-1 WARM
t1  Rolling update: Pod A terminates → slot-0 lock released; slot-0 dir + warm cache stay on disk
t2  Pod A' (replacement) lands on the node → slot-0 free → takes it
    → reuses warm layers → NO re-download/decompress          ← the win
```
Only two dirs ever exist on this node, reused forever. No orphans, no leak, and warm restarts.

### Why this is the approach to pursue

- **Fixes the root cause:** a fixed, reused set of slot dirs → orphans can't accumulate → **no sweeper needed, ever.**
- **Delivers the actual goal:** warm reuse across pod generations on the same node → faster builds (no re-pull/re-decompress of multi-GB CUDA bases). The janitor gives none of this.
- **Self-healing:** `flock` releases on process death — no cleanup logic, no grace period, no dry-run, no "is this dir dead?" heuristics.
- **Non-destructive:** slots are reused *in place*. The only deletion is a one-time migration sweep of legacy dirs (see below).
- **Right layer:** the logic lives in the buildkit pod's own entrypoint — the component that owns the cache — not smeared across a 786-node DaemonSet.

## Implementation guidance (for Huy)

1. **Mount the cache *root*, not a per-pod subpath.** Replace the `subPathExpr:$(POD_NAME)` mount at `/var/lib/buildkit` with a mount of the whole `/mnt/k8s-disks/0/buildkit-cache` hostPath into the pod (e.g. at `/cache`). Drop the `POD_NAME` env — no longer needed.

2. **Add an entrypoint wrapper that picks a slot.** Iterate `slot-0 … slot-(SLOTS-1)`, starting buildkitd on the first free one; advance when buildkitd exits with the lock error. Illustrative:
   ```sh
   for i in $(seq 0 $((SLOTS-1))); do
     root="/cache/slot-$i"; mkdir -p "$root"
     buildkitd --root "$root" "$@"        # runs in foreground if it wins the slot
     # got here => it failed fast because the slot is busy → try the next one
   done
   echo "no free slot on node" >&2; exit 1  # fail loudly, never double-lock/corrupt
   ```

3. **Let buildkitd's own lock arbitrate — do not pre-flock in the wrapper.** A second `flock` on the same `buildkitd.lock` inode conflicts even within one process, so grabbing a test-lock and then `exec`-ing buildkitd would make buildkitd fail to lock. Just start buildkitd and treat its `"another instance running?"` exit as "slot busy, try next." This also closes any TOCTOU window, since the real lock *is* buildkit's own. Two pods racing the same slot are serialized safely by the kernel — the loser advances.

4. **Set `SLOTS` = max `pods_per_node` for that arch's pool.** amd64 packs 2 (24xl) or 1 (12xl) per node; arm64 packs 4 (m7gd.16xl). Use the max (2 for amd64, 4 for arm64) so the densest node has enough slots; smaller nodes simply never touch the higher slots (the picker always starts at 0). The generator already computes `pods_per_node` per instance type (from #1007's map) — emit its max as an env.

5. **Budget the disk for N warm roots sharing one volume.** With fixed slots, up to `SLOTS` full caches coexist on the NVMe. buildkitd's GC (`reservedSpace`/`maxUsedSpace`/`minFreeSpace` in `buildkitd.toml`) is *per root*; tune it so `SLOTS × per-root cap` + headroom fits (5.7 TB amd64 / 3.8 TB arm64). This is the one genuinely new consideration — but it's *bounded*, unlike today's unbounded leak.

6. **Keep it a `Deployment` — don't switch to StatefulSet.** A StatefulSet's ordinals are Set-global and *not* pinned to nodes (a recreated `-0` can land anywhere), so it wouldn't give node-local reuse and would add ordering/headless-Service overhead for nothing. The slot is chosen at runtime by the lock, which is exactly what you want.

7. **One-time migration of legacy dirs.** After cutover, existing `<pod-name>` dirs won't match `slot-*` and won't be reused. Because the rollout is `maxSurge:0 / maxUnavailable:1`, the old pod on a node is gone before the new slot-pod starts, so the legacy dirs are dead. Clear them once — either a one-shot script (like the existing `scripts/cleanup-git-cache.sh`) **or**, pragmatically, let #1009 do a *single* enabled pass as the migration broom and then retire it. After that, no ongoing deletion is needed.

8. **Add reuse observability.** Log/emit the chosen slot index and whether it was warm (dir non-empty / cache size > 0) at startup, so you can confirm in practice that replacement pods are hitting warm slots.

## Open question (out of tree)

The production build client is `pytorch/pytorch`'s `.ci/docker/build.sh` under `REMOTE_BUILDKIT` (per `modules/buildkit/README.md`), not in this repo. If it passes `--export-cache/--import-cache` to a registry, some cross-node reuse happens *there* independently — worth confirming to know what production relies on today. (The `git-cache` is a separate, genuinely node-shared read-only hostPath — git repos, not image layers, and not the orphan source.)

huydhn:
Potentially doesn't need this anymore, moving this back to draft.

PR #1010, opened 2026-08-13
image-cache-janitor: reclaim orphaned BuildKit cache directories

huydhn:
* __->__ #1010
* #1008
* #1007

BuildKit pods cache to <cache>/<pod-name> on the instance store. The
directory outlives the pod and no surviving buildkitd knows about it, so
buildkit's own GC can never reclaim it. Two 8-day-old buildkit nodes were
holding 880 GB and 1194 GB there against 2 GB of containerd images — the
janitor was pruning the 2 GB and ignoring the rest.

The janitor already runs on these nodes and already has the containerd
socket, so live pod names come from `crictl pods` with no new RBAC. A
directory is orphaned when no pod of that name is running and it is older
than a grace period.

An empty pod list is treated as "unknown", not "nothing is running": the
janitor is itself a pod on the node, so crictl always reports at least
one, and without that floor a transient containerd fault would read as
"every directory is orphaned".

The cache is read through /proc/1/root rather than a hostPath mount,
which would have created the path on all ~786 nodes this DaemonSet covers
rather than the handful running BuildKit. Directory sizes are memoized so
a multi-hundred-GB orphan is measured once, not on every 300s cycle.

Ships with BUILDKIT_ORPHAN_DRY_RUN=true: it reports orphan count and bytes
as metrics but deletes nothing, so the sweep can be confirmed against real
nodes before it is allowed to remove anything.

### DIFF
--- a/osdc/base/kubernetes/image-cache-janitor/configmap.yaml
+++ b/osdc/base/kubernetes/image-cache-janitor/configmap.yaml
@@ -29,6 +29,7 @@ data:
     import json
     import logging
     import os
+    import shutil
     import socket
     import subprocess
     import sys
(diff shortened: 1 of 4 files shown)


huydhn:
Created by mistake — an amend dropped the ghstack trailers. The change is on #1009.

PR #1018, opened 2026-08-18
Pin GOMAXPROCS=2 on dcgm-exporter to stop 8-GPU node OOMKills

jeanschmidt:
**Impact:** dcgm-exporter DaemonSet on GPU nodes
**Risk:** low

## What
Adds `GOMAXPROCS=2` to the dcgm-exporter DaemonSet env, plus a doc explaining the root cause and the reasoning. One env line, no memory-limit change.

## Why
dcgm-exporter was OOMKilled fleet-wide (`osdc#1011`) — ~48% of 8-GPU pods, 0% on 1-/4-GPU nodes. The pinned Go 1.24 image has no cgroup-aware `GOMAXPROCS` (a Go 1.25 feature), so it defaults to the host vCPU count (up to 192 on the largest 8-GPU nodes), letting per-scrape allocation bursts transiently overshoot `GOMEMLIMIT` and breach the hard 512Mi cgroup ceiling. Pinning `GOMAXPROCS=2` — the value Go 1.25 would compute for the pod's `200m` CPU limit — caps that concurrency without touching memory limits.

# Notes
- This is a prod-soak-only change: staging has no GPU nodes, so it can't be exercised before merge. Validate post-rollout via `container_memory_working_set_bytes` and OOM attribution per instance type (see `docs/dcgm-exporter-memory-tuning.md`).
- The pin becomes redundant (but harmless) once the image is bumped to a Go 1.25+ build (`4.6.0-4.8.3` already ships on Go 1.26).
- If 8-GPU nodes still OOM afterward, native `libdcgm` per-GPU memory is the likely driver — next zero-memory lever is trimming the 23-metric CSV. Do not lower `GOMEMLIMIT` further; it was never the binding constraint.

### DIFF
--- a/osdc/CLAUDE.md
+++ b/osdc/CLAUDE.md
@@ -80,3 +80,4 @@ Reference documentation in `docs/`:
 | `docs/arc-fork-build-deploy.md` | ARC fork (jeanschmidt/actions-runner-controller) build/release workflow and chart publishing |
 | `docs/pypi-package-cache.md` | PyPI wheel cache architecture, slug naming, S3 layout, runner integration |
 | `docs/h100-fabric-handles-imex-channels.md` | Enabling CUDA fabric handles (`CU_MEM_HANDLE_TYPE_FABRIC`) on H100 runners — IMEX channel creation, the gated fabric runner, and verification |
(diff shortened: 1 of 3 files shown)


huydhn (approved):
The logic sounds reasonable to me, not sure how best to test this but deploy the change and see how it goes in staging and then prod

PR #1023, opened 2026-08-20
Raise the hf-cache 1-GPU rclone reserve to 1Gi

huydhn:
Raises the `hf-cache-mount-gpu1` tier 640Mi → 1Gi, with the matching 1Gi trim on the 1-GPU runner defs.

## Why
640Mi OOM-killed 39 gpu1 mounts on `meta-prod-aws-ue2` in a day (baseline 1-3) and is still climbing — 55 flagged at 22:00Z. An rclone OOM drops the FUSE node-wide → spurious `LocalEntryNotFoundError` in jobs.

It started when the GPU AMI rolled to nvidia `1.32-v20260818` (kernel 6.18.39 → 6.1.180); pinning that glob is tracked separately.

**The mechanism is not identified.** A matched-pair diff of the two AMIs (same instance type, same GPU count) found no configuration difference: same cgroup v2, same `memory.max`/`high`/`oom.group`, same `vm.dirty_*` and reclaim sysctls, same THP, same xfs-on-NVMe mount, same kubelet reserves and eviction thresholds, MGLRU disabled on both. Only the kernel binary and kubelet version differ, and kubelet writes identical cgroup knobs.

What *is* established is that the reserve is the only variable separating the OOMing tier from the clean ones, under the same kernel and the same timm workload:

| tier | reserve | AMI | OOM rate |
|---|---|---|---|
| cpu | 640Mi | old | 0% (700 mounts) |
| gpu1 | 640Mi | new | **19.4%** (206 mounts) |
| gpu4 | 2Gi | new | 0% (81) |
| gpu8 | 4Gi | new | 0% (65) |

`GOMEMLIMIT` doesn't substitute — it was already set to 576MiB on this tier when the kills happened, and it's a soft, heap-only ceiling. An old-AMI gpu4 mount was observed pinned at 1.98Gi of its 2Gi limit having reclaimed 73M pages with zero OOM kills, so this workload saturates whatever memcg it's given and survives only while reclaim keeps up. More headroom buys more room for that.

## Runner defs trimmed 1Gi
The reserve is `request == limit`, and the 8xl 1-GPU nodes have only 280Mi (a10g/l4) and 168Mi (t4) of margin. `just analyze-utilization` drops both node types to **0 pods** with the bump alone, so a10g/l4 go 113→112Gi and t4 115→114Gi. Labels keep their `-NNN-`. Verified back to 1 pod/node.

## Notes
- `buffer-size` stays `0`.
- Mitigation, not a root-cause fix. Follow-ups: pin the AMI glob; and `memory.high` is currently `max` on these cgroups, so there's no proactive reclaim before the hard OOM — setting it may be a better lever than raising the ceiling, and would also test the reclaim hypothesis.
- #1022 pins `GOMAXPROCS` on the same DaemonSets — complementary, bounds the heap side.

Deploy: redeploy `hf-cache` + `arc-runners`.

### DIFF
--- a/osdc/modules/arc-runners/defs/l-x86aavx2-29-113-a10g.yaml
+++ b/osdc/modules/arc-runners/defs/l-x86aavx2-29-113-a10g.yaml
@@ -1,12 +1,12 @@
 # ARC runner definition: l-x86aavx2-29-113-a10g
 # Instance type: g5.8xlarge — 1-GPU A10G, 32c/128Gi node
-# Actual K8s capacity ~118.4Gi. Max job memory after all overhead: 113Gi.
+# Actual K8s capacity ~118.4Gi. Max job memory after all overhead: 112Gi (label keeps -113-).
 runner:
   name: l-x86aavx2-29-113-a10g
   instance_type: g5.8xlarge
   disk_size: 150
   vcpu: 29
-  memory: 113Gi
+  memory: 112Gi
   gpu: 1
   proactive_capacity: 0
   hud_failure_base_capacity: 30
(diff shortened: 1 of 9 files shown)


jeanschmidt:
did you validate if this won't cause packing issues? AFAIR we don't have this headroom on most instances, so this should probably be paired with an adjustment on memory size for the runners...

huydhn:
#1024 has fixed the OOM. This is not needed anymore.

PR #1024, opened 2026-08-20
Cap the hf-cache rclone per-socket TCP receive buffer

huydhn:
Caps `net.ipv4.tcp_rmem` max at 512Ki in the rclone pod's own netns.

## Why
The gpu1 mounts are being OOM-killed on `meta-prod-aws-ue2` and now `meta-prod-aws-ue1`. The kernel OOM report shows it's **socket memory** — not the Go heap, not page cache. Six sampled kills, near-identical:

```
memory: usage 655772kB, limit 655360kB
  sock   492-502 Mi   ← 76% of the limit
  anon   130-143 Mi   ← well under GOMEMLIMIT=576MiB
  file         0 Mi
```

Under cgroup v2 socket buffers are charged to the container. Polling gpu1/`g5.16xlarge` mounts on **ue1, where both AMIs run side by side**, isolates the cause:

| AMI | kernel | connections | peak sock | per conn |
|---|---|---|---|---|
| 20260803 | 6.18.39 | 245–292 | **233 Mi** | 0.69–0.91 Mi |
| 20260817 | 6.1.180 | 45–254 | **472 Mi** | 1.47–1.92 Mi |

Connection count is unchanged — the old AMI peaked *higher*. The new kernel charges ~2× the socket memory for the same traffic, which consumes the tier's entire margin:

```
old:  233 sock + 135 anon + ~20  ≈ 390 Mi of 640  ✓
new:  472 sock + 135 anon + ~20  ≈ 627 Mi of 640  ✗
```

rclone 1.69.1 has no flag to cap VFS read concurrency (`--s3-upload-concurrency` is upload-only, `--transfers` gates uploads), so bound the per-socket ceiling instead: 512Ki × ~290 conns ≈ 145 Mi.

## Safety
- Writes to the **pod's own netns** (not `hostNetwork`) — cannot affect the node or other pods. Smoke test asserts both.
- Verified writable from the privileged container via a no-op write.
- Non-fatal; effective value logged. Tunable via `hf_cache.tcp_rmem_max`.
- No throughput cost: intra-region S3 BDP is far below 512Ki.

## Related
#1023 (640Mi → 1Gi) is complementary headroom — 627 Mi fits with margin. #1022 (`GOMAXPROCS`) targets `anon`, which is ~135 Mi on both kernels; I'd close it. Pinning the GPU AMI glob is a separate discussion.

Deploy: `just deploy-module <cluster> hf-cache`

### DIFF
--- a/osdc/modules/hf-cache/deploy.sh
+++ b/osdc/modules/hf-cache/deploy.sh
@@ -30,6 +30,10 @@ RCLONE_IMAGE=$(uv run "$CFG" "$CLUSTER" hf_cache.rclone_image "rclone/rclone:1.6
 TAINT_REMOVER_IMAGE=$(uv run "$CFG" "$CLUSTER" hf_cache.taint_remover_image \
   "public.ecr.aws/amazonlinux/amazonlinux:2023")
 VFS_CACHE_MAX_SIZE=$(uv run "$CFG" "$CLUSTER" hf_cache.vfs_cache_max_size "75%")
+# Per-socket TCP receive-buffer ceiling for the rclone pod netns; socket memory is what
+# OOM-kills the mount (see mount-daemonset.yaml.tpl). 512Ki x ~280 conns keeps sock well
+# inside the smallest tier; intra-region S3 BDP needs far less.
+TCP_RMEM_MAX=$(uv run "$CFG" "$CLUSTER" hf_cache.tcp_rmem_max "524288")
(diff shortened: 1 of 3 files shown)


jeanschmidt:
humm, I am slightly skeptical on this one, as this problem existed for pypi-cache and a similar solution did not provide any relief, but we can try, worse case is less performance

jeanschmidt (approved):

PR #1026, opened 2026-08-21
[gpu-dev] Add PyPI publish workflow

georgehong:
* #1027
* __->__ #1026

Ports wdvr/osdc's publish.yml to the monorepo layout, and also splits the workflow into a build and publish step as per recommendation of pypa. 

- Tag scoped to `gpu-dev/v*`
- Split build/publish; the credentialed job only downloads and uploads
- Artifact/attest/publish actions ignore defaults.run.working-directory,
  so those paths are explicit
- Guards: tag on main, version matches pyproject, version not already published

Test plan: 
- Top of stack: builds, able to connect to server, basic function check
- Once committed, will attempt to publish by pushing the new version bump with tag `gpu-dev/v0.7.16` to main. Environment and publishers have been updated already.


huydhn (approved):
LGTM! https://github.com/pytorch/ci-infra/actions/runs/32520551644 looks correct

PR #1032, opened 2026-08-22
OSDC: Add S3 gateway VPC endpoint to keep in-region S3 off the NAT gateways

huydhn:
`NatGateway-Bytes` is the largest single line on the OSDC bill — $149,257 net across the fleet in July 2026 (5,631 TB), ue1 and ue2 at 99.8% of the volume. Regressing daily NAT bytes on daily S3 GET count for ue1 (Jul 7–31) gives **r² = 0.9465 at 64.3 MB/GET with an intercept of −6,186 GB/day**, i.e. zero: it is all same-region S3, at the size profile of an image layer blob. Direction agrees — 2,720 TB in vs 257 TB out.

Harbor's `disableredirect: false` sends containerd straight to a presigned S3 URL (correct — it keeps multi-TB/day off the registry pods), but no OSDC VPC ever had an S3 gateway endpoint (the legacy ALI VPCs do), so those bytes matched the `0.0.0.0/0` NAT route at $0.045/GB. Gateway endpoints are free and in-region EC2→S3 transfer is free, so this takes them to $0. The redirect stays on — with the endpoint it beats both alternatives.

**Rollout:** staging-uw1 → prod-uw2 while it is still paused → ue2 → ue1. Confirm a second `tofu plan` is clean per cluster: endpoint routes come from `ModifyVpcEndpoint`, not `CreateRoute`, so they cannot be inline `route` blocks on the private route tables.

**Tested:** `just test` OK; `tofu fmt -check -recursive` and `tofu validate` clean. `just lint` fails only on a pre-existing hadolint finding at `base/docker/runner-base/Dockerfile:36`, reproducible on a clean tree.

Details, risks reviewed, and verification commands in `docs/s3-gateway-endpoint.md`.

### Testing

Deploy on ue1 staging, run the integration tests https://github.com/pytorch/pytorch-canary/actions/runs/32757852501/job/97529425214

### DIFF
--- a/osdc/CLAUDE.md
+++ b/osdc/CLAUDE.md
@@ -71,6 +71,7 @@ Reference documentation in `docs/`:
 | `docs/observability-estimates.md` | Per-unit cost estimates for metrics cardinality and log volume (Grafana Cloud billing) |
 | `docs/operations.md` | Operational prerequisites — AWS CLI, mise, working directory setup for cluster management |
 | `docs/ipv6-cluster-recreation.md` | Operator runbook for destroying and recreating an OSDC cluster as IPv6-only (accepted data losses: Harbor S3, EFS pypi-cache) |
+| `docs/s3-gateway-endpoint.md` | Why every VPC needs an S3 gateway endpoint — NAT data-processing analysis, the Harbor redirect interaction, and the inline-route Terraform gotcha |
 | `docs/loki_query.md` | CLI queries against Grafana Cloud Loki when kubectl logs is unavailable |
 | `docs/mimir_query.md` | CLI queries against Grafana Cloud Mimir (Prometheus metrics, no in-cluster Prometheus) |
 | `docs/runner_naming_convention.md` | Runner label format and the ~42 character name limit (ARC/K8s/Cilium constraints) |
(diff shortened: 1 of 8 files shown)


huydhn:
Testing on staging ue1

```
---

  Verified on meta-staging-aws-ue1 — endpoint vpce-0f69a3fe78511b6cc, created 2026-08-24
  17:07:51Z.

  Route is in place and winning:

  meta-staging-aws-ue1-vpc-private-1  (rtb-0509345c8cbcfb160)
      0.0.0.0/0       -> nat-0daed370e11da9e18    active
      pl-63a5400a     -> vpce-0f69a3fe78511b6cc   active   <- more specific

  One private route table is expected here (single_nat_gateway: true on staging); all 3
  private subnets associate with it, so no AZ is uncovered. On prod this will be one table per
  AZ.

  Traffic confirmed bypassing NAT. During the integration test (run 32757852501, success), a
  cold node pulled ~15 GB from a same-region S3 bucket while the NAT gateway stayed at its ~16
  MB/5min idle floor for the entire window. Same-region S3 is off the NAT path.
```

zxiiro (approved):

jeanschmidt (approved):

PR #1034, opened 2026-08-25
Proxy the private ECR through Harbor on off-region clusters

huydhn:
Follow-up to #1032, which cut `meta-prod-aws-ue1`'s NAT traffic 71% and did nothing for `ue2` or `uw1`.

Every cluster pulls the PyTorch CI images from us-east-1 ECR, and ECR redirects layer downloads to an IPv4-only bucket **in its own region**. ue1 is us-east-1, so those bytes match its prefix list and take the gateway endpoint. Off-region clusters miss the prefix list and pay NAT plus inter-region on every pull.

So route those pulls through Harbor's proxy cache, as a seventh entry alongside the six upstreams it already handles.

## Validated on `meta-staging-aws-ue2`

| | |
|---|---|
| Harbor → private ECR auth | `aws-ecr` endpoint, `status: healthy` |
| Proxy + cache | `ecr-cache/pytorch/…`, bucket 6.2 → 8.77 GB |
| Cache **miss** (cross-region) | 8,763 MB on the NAT gateway — one-time per layer |
| Cache **hit**, cold node | **4.0 GiB pulled, NAT flat at 16.1 / 12.4 MB** (idle floor ~15 MB) |

4 GiB reached a node that had never held the image and the NAT gateway did not move.

Gated on `harbor.ecr_pullthrough`, default off, enabled only on `meta-staging-aws-ue2`. Does not help ue1 — its ECR pulls are already same-region.

Background and measurements: `docs/cross-region-ecr-pulls.md`.

**Tested:** `just test` OK (3 unit + 2 smoke cases). `tofu fmt`/`validate` clean. `just lint` fails only on a pre-existing hadolint finding at `base/docker/runner-base/Dockerfile:36`.

**Not yet exercised:** the containerd mirror path. Validation above pulled `harbor:30002/ecr-cache/...` directly, which bypasses the rewrite. Deploying to staging-ue2 and pulling by the real ECR reference is what takes this out of draft.

### DIFF
--- a/osdc/CLAUDE.md
+++ b/osdc/CLAUDE.md
@@ -77,6 +77,7 @@ Reference documentation in `docs/`:
 | `docs/current_runner_load_distribution.md` | Job counts and peak concurrency by runner type (pytorch/pytorch, from ClickHouse) |
 | `docs/node-utilization-optimization.md` | Runner-to-node packing efficiency analysis and instance type recommendations |
 | `docs/node-warmup-and-scheduling-gates.md` | Full node initialization sequence — taints, DaemonSets, init containers before job scheduling |
+| `docs/cross-region-ecr-pulls.md` | Private ECR pull-through via Harbor — why off-region clusters miss the S3 gateway endpoint, and the design that routes those pulls through the local cache |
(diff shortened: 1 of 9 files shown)


huydhn:
Superseded by #1035 — same change, resubmitted as a ghstack with #1036 stacked on top for the prod enablement.

PR #1035, opened 2026-08-25
Proxy the private ECR through Harbor on off-region clusters

huydhn:
* #1036
* __->__ #1035

Follow-up to #1032, which cut `meta-prod-aws-ue1`'s NAT traffic 71% and did nothing for `ue2` or `uw1`.

Every cluster pulls the PyTorch CI images from us-east-1 ECR, and ECR redirects layer downloads to an IPv4-only bucket **in its own region**. ue1 is us-east-1, so those bytes match its prefix list and take the gateway endpoint. Off-region clusters miss the prefix list and pay NAT plus inter-region on every pull.

So route those pulls through Harbor's proxy cache, as a seventh entry alongside the six upstreams it already handles.

## Validated on `meta-staging-aws-ue2`

| | |
|---|---|
| Harbor → private ECR auth | `aws-ecr` endpoint, `status: healthy` |
| Proxy + cache | `ecr-cache/pytorch/…`, bucket 6.2 → 8.77 GB |
| Cache **miss** (cross-region) | 8,763 MB on the NAT gateway — one-time per layer |
| Cache **hit**, cold node | **4.0 GiB pulled, NAT flat at 16.1 / 12.4 MB** (idle floor ~15 MB) |

4 GiB reached a node that had never held the image and the NAT gateway did not move.

Gated on `harbor.ecr_pullthrough`, default off, enabled only on `meta-staging-aws-ue2`. Does not help ue1 — its ECR pulls are already same-region.

Design: `docs/cross-region-ecr-pulls.md`.

**Tested:** `just test` OK (3 unit + 2 smoke cases). `tofu fmt`/`validate` clean. `just lint` fails only on a pre-existing hadolint finding at `base/docker/runner-base/Dockerfile:36`.

**Not yet exercised:** the containerd mirror path. Validation above pulled `harbor:30002/ecr-cache/...` directly, which bypasses the rewrite. Deploying to staging-ue2 and pulling by the real ECR reference is what takes this out of draft.

### DIFF
--- a/osdc/CLAUDE.md
+++ b/osdc/CLAUDE.md
@@ -78,6 +78,7 @@ Reference documentation in `docs/`:
 | `docs/current_runner_load_distribution.md` | Job counts and peak concurrency by runner type (pytorch/pytorch, from ClickHouse) |
 | `docs/node-utilization-optimization.md` | Runner-to-node packing efficiency analysis and instance type recommendations |
 | `docs/node-warmup-and-scheduling-gates.md` | Full node initialization sequence — taints, DaemonSets, init containers before job scheduling |
+| `docs/cross-region-ecr-pulls.md` | Private ECR pull-through via Harbor — why off-region clusters miss the S3 gateway endpoint, and the design that routes those pulls through the local cache |
 | `docs/arc-fork-build-deploy.md` | ARC fork (jeanschmidt/actions-runner-controller) build/release workflow and chart publishing |
 | `docs/pypi-package-cache.md` | PyPI wheel cache architecture, slug naming, S3 layout, runner integration |
 | `docs/h100-fabric-handles-imex-channels.md` | Enabling CUDA fabric handles (`CU_MEM_HANDLE_TYPE_FABRIC`) on H100 runners — IMEX channel creation, the gated fabric runner, and verification |
--- a/osdc/base/kubernetes/registry-mirror-config.yaml
+++ b/osdc/base/kubernetes/registry-mirror-config.yaml
@@ -56,18 +56,19 @@ data:
     MARKER_FILE="/var/lib/registry-mirror-config/.configured"
     HARBOR_PORT="${HARBOR_PORT:-30002}"
 
+    # Conditional entry must join the list, or enabling it later keeps skipping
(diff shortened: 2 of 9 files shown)


huydhn:
---

**Validation complete on `meta-staging-aws-ue2` (2026-08-25).** The containerd mirror path is now exercised: pulling by the real ECR reference on a cold node landed the image in Harbor's cache —

```
docker/registry/v2/repositories/ecr-cache/pytorch/pytorch-linux-focal-linter/
```

which only happens if containerd resolved to `harbor:30002/v2/ecr-cache`. Cache-miss cost 3,377 MB of cross-region NAT as expected (one-time per layer); the cache-hit case was measured separately at 4.0 GiB pulled with NAT flat at 16.1 / 12.4 MB.

huydhn:
@zxiiro After looking at the deployment results of https://github.com/pytorch/ci-infra/pull/1032, I realize that PR alone is not enough.

https://github.com/pytorch/ci-infra/pull/1032 by itself fixes the issue on `ue1` only because PyTorch ECR is on that region. I can see the drop in this region as expected. For other regions like `ue2`, the pull is cross region traffic and will go through the public NAT.

To fix that, we need to properly setup pull through cache for these regions. The flow there becomes `ECR (us-east-1) ──1st pull, cross-region──► Harbor cache in S3 (us-east-2) ──all later pulls, same region──► free`

zxiiro (approved):

zxiiro:
> @zxiiro After looking at the deployment results of #1032, I realize that PR alone is not enough.
> 
> #1032 by itself fixes the issue on `ue1` only because PyTorch ECR is on that region. I can see the drop in this region as expected. For other regions like `ue2`, the pull is cross region traffic and will go through the public NAT.
> 
> To fix that, we need to properly setup pull through cache for these regions. The flow there becomes `ECR (us-east-1) ──1st pull, cross-region──► Harbor cache in S3 (us-east-2) ──all later pulls, same region──► free`

I was going to raise this as well after doing some research yesterday I realized it only reduced gateway use for ue1. Looks like you took care of it. Let's see if this change improves costs in the other regions.

PR #1043, opened 2026-08-25
agent-sandbox: enforce the task-pod isolation contract in the API server

izaitsevfb:
✴️ iz2: Every isolation property of an agent task pod — gVisor, no service-account token, no volumes, the task image and nothing else — is currently enforced by one Python function, the dispatcher's `job_manifest()`. The API server accepts whatever that function emits, so a bug or a careless refactor there is enough to produce a task pod with no gVisor and a mounted token, and nothing in the cluster would notice.

RBAC cannot narrow this. The dispatcher legitimately holds `create jobs`, and "create a Job" is not a verb that can be expressed as "create a Job shaped like this".

This adds a `ValidatingAdmissionPolicy` that restates the same contract in the API server. It is deliberately redundant with `job_manifest()` — that is the point. Anything it rejects is either an attack or a dispatcher bug.

It is also the precondition for the next PR in this stack: it is what makes it safe to put a JWT parser inside the process that holds create-Job RBAC, rather than in a separate pod in front of it. The worst case of a parser compromise becomes arbitrary Job creation, and this bounds what an arbitrary Job may look like.

## Shape

- **Checked twice, on the Job and again on the Pod.** Validating only the Job template leaves a window: the Pod is created later by the Job controller, and any mutating webhook in the cluster gets to add a sidecar, a Secret volume or a projected token to it in between. The Pod pass is a second binding, label-scoped to `app: sandbox-task` so it does not catch the dispatcher and proxy pods.
- **The Job pass is scoped to the dispatcher's service account**, which is the threat this exists for. That is also what lets the namespace hold an ordinary maintenance Job without each one having to satisfy a contract written for untrusted workloads.
- **`failurePolicy: Fail`, `validationActions: [Deny]`.** An evaluation error rejects the request rather than silently unenforcing the contract, and a warning about a Job escaping the sandbox is a log line nobody reads.

The rules cover runtime class, service account, token automount, host namespaces, `nodeName`/`schedulerName`, volumes, resource claims, sysctls, image repository, container and init-container counts, `envFrom`/`valueFrom`, privilege and capabilities, `procMount`, host ports, resource limits and Guaranteed QoS, `terminationGracePeriodSeconds`, and Job-level `parallelism`/`completions`/`backoffLimit`/`activeDeadlineSeconds`.

[...]

PR #1044, opened 2026-08-25
agent-sandbox: split the dispatcher into one file per boundary

izaitsevfb:
✴️ iz2: **Stacked on #1043 — review that one first; this PR's diff is against it.**

No behaviour change. `dispatcher.py` becomes four modules, imports running strictly downward, each named for a boundary the process crosses:

```
__main__  ->  http_api  ->  tasks  ->  kube
```

| file | owns |
|---|---|
| `kube.py` | the Kubernetes API and the Job template — the task-pod security boundary, reviewable on its own |
| `tasks.py` | the task table and the run-one-Job-to-completion loop |
| `http_api.py` | parse, bound, dispatch, answer |
| `__main__.py` | wiring, and nothing else |

The split is preemptive rather than cosmetic. OIDC verification and caller authorization land in this process in the next PR, and a single file that already mixes the HTTP surface with the code holding create-Job RBAC is the wrong thing to add a JWT parser to. The rule the split encodes: every file names a boundary the system crosses and states its contract in one sentence that does not mention its callers. Corollary — if a realistic one-behaviour change needs more than two files open, the split is wrong.

Two calls worth your eye:

- **The Kubernetes code stays one unit.** Splitting `job_manifest()` from `create_job()` would put the gVisor runtime class in a different file from the call that submits it, which makes the security review worse, not better.
- **Modules import modules and call through them** — `kube.job_state(...)`, never `from kube import job_state`. A name bound at import time is a second binding that a test substituting the call would not reach.

The container entrypoint is now `python3 <dir>`, which runs `__main__.py` with `<dir>` prepended to `sys.path` — no package, no `PYTHONPATH`, no install step. `deploy.sh`'s content-hash tag already walks the directory recursively, so it picks up all four files.

## Test plan

- The same 55 tests, ported to the new module names, passing unchanged. Same count before and after, which is the claim.
- The entrypoint additionally smoke-tested the way the container invokes it: `python3 <dir>` → listening on `[::]`, `/healthz` 200 with the right capacity, unknown path 404.
- `__main__.py` is loaded by path under another name in the test, because `import __main__` resolves to whatever is running the process. It is still worth a test — it decides the listen address, and binding IPv4 on this IPv6-only cluster is a silent, total outage.


osdc/modules/agent-sandbox/dispatcher/Dockerfile
```diff
@@ -20,11 +20,15 @@ RUN apt-get update \
         python3 \
     && rm -rf /var/lib/apt/lists/*
 
-COPY dispatcher.py /usr/local/bin/dispatcher.py
-RUN chmod 0755 /usr/local/bin/dispatcher.py
+# The modules are listed rather than globbed so the image cannot pick up a test file,
+# and so adding one is a visible line in review.
+COPY __main__.py http_api.py tasks.py kube.py /usr/local/lib/sandbox-dispatcher/
```
jeanschmidt, line 25: yeah, a check/verification seems reasonable to avoid regressions

---

<details>
<summary>🟡 The Dockerfile's explicit module list and the directory deploy.sh hashes are two places that must agree, with nothing checking that they do (ai-generated section)</summary>

The image's module set is now enumerated in the Dockerfile, while deploy.sh derives the image tag by globbing the same directory. Before this change the dispatcher was one file and there was nothing to forget; with four files and more announced, adding a module updates the content hash (so a new tag is built and pushed) without updating this COPY line, and the new image simply lacks the module. deploy.sh:63-68 hashes the directory with `find "$1" \( -name '*.py' -o -name 'Dockerfile' \) ! -name 'test_*' ! -name 'conftest.py'`, so any new non-test .py changes DISPATCHER_TAG and triggers `build_and_push ci-agent-sandbox-dispatcher "$DISPATCHER_TAG" "$MODULE_DIR/dispatcher"` (deploy.sh:124). The Dockerfile lists the four modules by hand. The PR description states the next change adds OIDC verification to this process, i.e. at least one more module. The repo already has a pattern that avoids the hand-list: osdc/modules/zombie-cleanup/deploy.sh:107-113 builds a temp context, copies the sources, then `rm -f "$BUILD_CONTEXT/test_"*.py` and `rm -f "$BUILD_CONTEXT/conftest.py"`, so osdc/modules/zombie-cleanup/docker/Dockerfile:6 can safely say `COPY *.py ./`; osdc/base/node-compactor/docker/Dockerfile:10 and osdc/modules/harbor-cache-recovery/docker/Dockerfile:6 do the same. A module added without a matching COPY line produces a pushed image that dies at import. `kubectl rollout status deployment/sandbox-dispatcher -n ai-sandbox --timeout=10m` (deploy.sh:207) then fails ten minutes into a deploy with a CrashLoopBackOff whose only clue is a ModuleNotFoundError in the pod log; the old replicas keep serving, so nothing else signals it. Narrow, and the one I would pick because it keeps the "adding one is a visible line in review" property the comment above argues for: assert the two agree in deploy.sh before building, e.g. `for f in "$MODULE_DIR"/dispatcher/*.py; do b=$(basename "$f"); case "$b" in test_*|conftest.py) continue;; esac; grep -q "$b" "$MODULE_DIR/dispatcher/Dockerfile" || { echo "[agent-sandbox] ERROR: $b is not COPYed by dispatcher/Dockerfile" &gt;&amp;2; exit 1; }; done`. Broader: adopt the zombie-cleanup pruned-temp-context pattern (deploy.sh:107-113) for both agent-sandbox images and use `COPY *.py`, which removes the list entirely.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 502b2a0a9.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/http_api.py
```diff
@@ -0,0 +1,137 @@
+"""The HTTP surface: parse, bound, dispatch, answer. No task logic lives here.
```
jeanschmidt, line 1: :)

---

<details>
<summary>🟡 agent/sandbox.py's module docstring points at ../dispatcher/dispatcher.py, which this change deletes (ai-generated section)</summary>

This change deletes `dispatcher/dispatcher.py` and moves the HTTP surface into `dispatcher/http_api.py`, but the module docstring of `osdc/modules/agent-sandbox/agent/sandbox.py` (line 8, a file this diff does not touch) still says "The HTTP surface lives in the dispatcher (../dispatcher/dispatcher.py)". That relative path now resolves to nothing. `osdc/modules/agent-sandbox/agent/sandbox.py:8`: "The HTTP surface lives in the dispatcher (../dispatcher/dispatcher.py): it turns each request into one Job". The diff deletes `osdc/modules/agent-sandbox/dispatcher/dispatcher.py` in full, and `ls` of the directory now holds only `__main__.py`, `http_api.py`, `kube.py`, `tasks.py` and the two test files. A repo-wide grep for `agent-sandbox/dispatcher|dispatcher/dispatcher` returns this line as the only remaining path reference. Someone reading the untrusted task library to find where /run is served follows the path, finds no such file, and has to re-derive the layout from the Dockerfile. The pointer was written precisely so that reader would not have to. Edit `osdc/modules/agent-sandbox/agent/sandbox.py:8` to name the new file: "The HTTP surface lives in the dispatcher (../dispatcher/http_api.py): it turns each".

<sub>Reviewed by claude-opus-5[1m] at max effort, against 502b2a0a9.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/kube.py
```diff
@@ -0,0 +1,237 @@
...
+TASK_DEADLINE_S = int(os.environ.get("TASK_DEADLINE_S", "900"))
+# Finished Jobs are deleted as soon as their log is read; this is the backstop for a
+# dispatcher that died mid-wait, and has to outlast a plausible restart.
+JOB_TTL_S = 3600
+SLOT_CPU = os.environ.get("TASK_CPU", "2")
+SLOT_MEMORY = os.environ.get("TASK_MEMORY", "4Gi")
+SLOT_DISK = os.environ.get("TASK_EPHEMERAL_STORAGE", "20Gi")
+
+MAX_LOG_BYTES = 1024 * 1024
+# How much of a non-2xx body to quote back. Same number as the request-body cap in
```
jeanschmidt: nit, should be an easy fix

---

<details>
<summary>⚪ The MAX_ERROR_BYTES comment says it caps how much of an error body is quoted back, but api_request truncates the quote to 500 characters (ai-generated section)</summary>

The comment added above the new `MAX_ERROR_BYTES` constant opens by describing it as "How much of a non-2xx body to quote back", but at `kube.py:98` the value only bounds the `exc.read(...)`; what actually reaches the ApiError message is capped by the `[:500]` slice. The comment's own third clause ("this one bounds what we read FROM the API server") states it correctly, so the constant carries two descriptions and the first one is not what the code does. `kube.py:46` defines `MAX_ERROR_BYTES = 64 * 1024`; `kube.py:98` is `detail = exc.read(MAX_ERROR_BYTES).decode(errors="replace")[:500] if hasattr(exc, "read") else ""`, and `kube.py:99` raises `ApiError(f"{method} {path} -&gt; {exc.code}: {detail}")`. A 64 KiB API error body is therefore reported as 500 characters, not 65536. Someone who wants more of a 403/422 body in the dispatch error — the case the surrounding prose says the message exists for — raises MAX_ERROR_BYTES per this comment and sees no change, because the 500-character slice is what governs. Reword line 43 to describe the read rather than the quote: "# How much of a non-2xx body to read before quoting its first 500 characters back." A broader alternative is to hoist the 500 into a named constant beside this one so both bounds are visible in the same place; I would take the reword, since the split's premise is that the constant block states what each number does.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 502b2a0a9.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/tasks.py
```diff
@@ -0,0 +1,129 @@
...
+MAX_CONCURRENT_TASKS = int(os.environ.get("MAX_CONCURRENT_TASKS", "6"))
+# How long /status can still answer for a finished task before its result is dropped.
+RESULT_RETENTION_S = int(os.environ.get("RESULT_RETENTION_S", "3600"))
+POLL_INTERVAL_S = 2
+
+_TASKS: dict[str, dict] = {}
+_TASKS_LOCK = threading.Lock()
+
+
+def run_to_completion(task_id: str, spec: dict) -> dict:
```
jeanschmidt: nit, should be an easy fix

---

<details>
<summary>⚪ tasks.run_to_completion is public but skips the task table, so a caller pairing it with start_task() leaks a concurrency slot for the life of the pod (ai-generated section)</summary>

tasks.py now exposes three run-shaped names: `run_to_completion` (runs the Job, never touches `_TASKS`), `run_and_record` (runs it and calls `_finish`), and `run_in_background` (threads `run_and_record`). Only the latter two close the slot that `start_task()` opened. Every genuinely-internal helper in this file carries a leading underscore — `_running_locked`, `_prune_locked`, `_finish` — so the unprefixed `run_to_completion` reads as part of the module's public API, while it is in fact the one entry point that breaks the reserve/release protocol. `start_task()` (tasks.py:93-102) inserts `{"state": "running", "result": {}, "finished_at": 0.0}`. `_prune_locked()` (tasks.py:78) only deletes entries whose `state == "done"`, and `_running_locked()` (tasks.py:68) counts entries whose `state == "running"`. An entry that is never passed to `_finish()` is therefore counted forever and never pruned. The two in-repo callers in http_api.do_POST (http_api.py:118 and http_api.py:122) both go through the recording wrappers, so nothing is broken today; every test in test_dispatcher.py, however, calls `tasks.run_to_completion(...)` directly (lines 160, 168, 173, 178, 186, 194, 404, 409, 438), which is what makes it the obvious name to reach for. With MAX_CONCURRENT_TASKS defaulting to 6 (tasks.py:24), six such calls wedge the dispatcher. A future caller in http_api that reserves a slot and then calls `run_to_completion` instead of `run_and_record` — the plausible mistake when adding an authenticated /run path — permanently consumes a slot per request. After six requests every /run answers 429 with "at capacity: 6 tasks in flight" until the pod restarts, and /healthz reports in_flight=6 with no Job or pod in the namespace to explain it, so the debugging starts at the capacity code rather than at the caller. Narrow fix: rename to `_run_to_completion` so the leading-underscore convention this file already uses marks it as internal, and update the nine call sites in test_dispatcher.py. Broader fix: fold the slot reservation into the runner — have `run_and_record` take the spec and return `None` at capacity, so `start_task()` need not be public either. I would take the rename; it costs a test-only edit and removes the trap.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 502b2a0a9.</sub>

</details>

jeanschmidt (commented):
## Review summary

🟡 2 minor · ⚪ 2 nit

- 🟡 The Dockerfile's explicit module list and the directory deploy.sh hashes are two places that must agree, with nothing checking that they do — `osdc/modules/agent-sandbox/dispatcher/Dockerfile:25`
- 🟡 agent/sandbox.py's module docstring points at ../dispatcher/dispatcher.py, which this change deletes — `osdc/modules/agent-sandbox/dispatcher/http_api.py:1`
- ⚪ The MAX_ERROR_BYTES comment says it caps how much of an error body is quoted back, but api_request truncates the quote to 500 characters — `osdc/modules/agent-sandbox/dispatcher/kube.py:43`
- ⚪ tasks.run_to_completion is public but skips the task table, so a caller pairing it with start_task() leaks a concurrency slot for the life of the pod — `osdc/modules/agent-sandbox/dispatcher/tasks.py:33`

<details>

**What this changes**

`dispatcher/dispatcher.py` becomes four modules — `kube.py`, `tasks.py`, `http_api.py`, `__main__.py` — with imports running downward and every cross-module call made through the module object rather than a bound name. The Dockerfile stops copying one script into `/usr/local/bin` and copies four named files into a directory, with the entrypoint changed from a script path to `python3 &lt;dir&gt;`. The tests move to the new module names with their assertions intact, plus an importlib shim that loads `__main__.py` under another name. The move is behaviour-preserving; the two substantive edits carried inside it are `tasks.status()`, which builds the `/status` payload under the lock so the HTTP layer never touches `_TASKS`, and `MAX_ERROR_BYTES`, a separate constant for the error-body read holding the same value the shared cap had.

**What it gets right**

Keeping `job_manifest()` and `create_job()` in one file. The gVisor runtime class, `automountServiceAccountToken: False` and the requests-equal-limits block sit next to the call that submits them, so the task-pod boundary stays readable as one unit. The module-call convention is load-bearing rather than stylistic: the ported tests patch `kube.job_state` and `kube.api_request` directly, and a `from kube import` binding in `tasks.py` would have made those substitutions silently miss. `tasks.status()` is a real narrowing over the pre-split code, not just a move — the table and its lock are now unreachable from `http_api`. And the entrypoint gets a test despite being the one module that cannot be imported by name, which is the right call when binding IPv4 on an IPv6-only cluster fails totally and quietly.

**Where the risk is**

The dominant shape is description that no longer matches the code it points at. Several independent passes converged on the same two comments outside `dispatcher/` — `kubernetes/base/kustomization.yaml:9` and `agent/sandbox.py:8` — which still name the deleted `dispatcher.py`; they share a cluster and one two-line edit settles every one of them. That density of convergence is the signal. The kustomization line is the costlier of the two: it is the pointer from the admission policy to the Python that must satisfy it.

The same class recurs inside the new files, in prose written for this split. The `MAX_ERROR_BYTES` comment in `kube.py` describes a quote-back cap that the `[:500]` slice below it actually governs, and `http_api.py`'s "it never decides anything" is contradicted by the `AGENT_IMAGE` 500 twelve lines into `do_POST`. Both are small, and both matter more than usual here, because the split's stated premise is that each file's contract fits in one sentence.

The Dockerfile holds a second, separate risk at lines 25 and 34: a hand-written module list against a `deploy.sh` that hashes the directory by glob, and a directory entrypoint that diverges from every other Python image in the repo and is what forces the importlib shim in the test. Those two compound — a fix for either rewrites the same lines — and both are triggered by exactly the event the description announces, a fifth module arriving. `run_to_completion` being public while skipping the task table compounds with that too: the plausible caller who reaches for the wrong runner is the one adding the authenticated path.

**Before merge, and after**

Before: retarget the two references to the deleted `dispatcher.py` in `kubernetes/base/kustomization.yaml` and `agent/sandbox.py`. Two words, no design decision. With them, the `MAX_ERROR_BYTES` comment in `dispatcher/kube.py` and the "never decides anything" sentence in `dispatcher/http_api.py` — one line each, and both are contract prose the next change will be read against.

After: the `run_to_completion` name in `dispatcher/tasks.py`. Its label says minor, but the fix is a choice between a rename touching nine test call sites and folding slot reservation into the runner, and that choice is worth making deliberately. The two Dockerfile items — the explicit module list versus the hashed directory, and the directory entrypoint — are hazards for the next module rather than defects in what merges here, so follow-up is right, provided they land before the module that trips them.

</details>

huydhn (approved):
Stamped to unblock! Let's try this out in staging, I guess the current integration tests for agent sandbox will do, but maybe we can enhance it later to test all the HTTP surface (if we haven't done that already)

jeanschmidt (approved):

PR #1045, opened 2026-08-25
agent-sandbox: authenticate /run with GitHub OIDC and authorize by policy

izaitsevfb:
✴️ iz2: **Stacked on #1044, which is stacked on #1043. Review in that order; this PR's diff is against #1044.**

`/run` and `/status` are unauthenticated today: any pod in `arc-runners` can reach them, spend the Bedrock budget, and read anyone's results. This closes both.

## Shape

Two new files, each one boundary. `oidc.py` answers only *did GitHub sign this token, for us, recently*. `authorize.py` answers only *may this caller do this*, and imports nothing else in the process — a reviewer asking what a caller is allowed to make happen reads one file and stops.

**The request decides less than it looks like it does.** `authorize()` returns a frozen `Grant`, and the Job is built from the Grant alone, so the repository to clone and the model are policy rather than request. A caller naming a different repo — or a different model — is refused with a 403 rather than quietly given the policy's one. When the capability manifest lands it changes where the Grant's values come from and touches nothing downstream; `test_authorize.py` runs the same allow/deny table against the v1 constants *and* a stub loader carrying the v2 signature, so the seam is exercised now instead of promised.

**v1 policy, as a module constant rather than an env var**, because it is the answer to "who may spend our Bedrock budget" and belongs in git history and review. One caller — `pytorch/ciforge` — matched on `repository_id`/`repository_owner_id` (names can be renamed and re-registered), on a protected ref, on a github-hosted runner, on a non-pull-request event. Both `workflow_ref` and `job_workflow_ref` must be inside that repo, or an allowed repo could delegate its identity to a reusable workflow living anywhere.

**PyJWT + `cryptography` from apt, not a hand-rolled RS256 verifier.** The stdlib has no public-key crypto, and the fifty-first line of a from-scratch check is a Bleichenbacher forgery. The image already links OpenSSL for its Kubernetes client, so this opens no category that was closed; the untrusted sandbox image is built from `agent/` and gains nothing from it.

**Keys arrive as a mounted ConfigMap refreshed by a CronJob**, because the dispatcher is the process holding create-Job RBAC and therefore the one that must not reach the internet. The refresher's Role can patch that one named object and nothing more. It writes `fetched_at` *into* the document: a ConfigMap volume only updates when its content changes and GitHub rotates rarely, so an mtime-based staleness bound would age out a healthy key set while a refresher dead for a month looked identical.

**The manifest declares that ConfigMap with no `data`.** This is load-bearing, not tidiness. An earlier revision seeded it with an empty placeholder, which made every deploy blank the live signing keys — apply reasserts what the manifest declares, so the seed overwrote whatever the refresher had written. Note the tempting fix is a trap in the other direction: had the seed ever shipped, later removing the `data:` block would make apply prune the field it already owned. A test asserts the ConfigMap stays data-less, because re-adding a placeholder looks like an improvement.

## The one thing that needs a decision

`REQUIRE_AUTH` ships **`false`** and governs exactly one case — a request with no `Authorization` header. A presented token is always verified and authorized whatever the flag says.

Be clear about what that does and does not buy: **while it is false, authentication is optional**, and an unauthenticated caller can do *more* than one whose real token was denied, by omitting the header. That is a migration window, not a posture. It is tolerable only because `/run` is already reachable unauthenticated by the whole `arc-runners` namespace today — strictly no worse than the status quo, strictly better once flipped. **It must be flipped once the GHA client sends tokens.** Unrecognised values abort at startup rather than defaulting to off.

## Test plan

- `test_oidc.py` signs real RS256 tokens with generated keys and drives the actual verifier: `alg=none`, an HS256 token forged from the public key (assembled by hand, because PyJWT refuses to *encode* it), a `kid` naming one key with another's signature, expiry, audience, issuer, missing required claims, a stale key set, a key set timestamped in the future, and a rotation picked up without a restart.
- `test_dispatcher.py` covers the same over the real HTTP surface — including that a forged token is 401 *while `REQUIRE_AUTH` is still false*, that `/status` reports another caller's task as absent rather than forbidden, and that a token `/run` would deny is not an identity `/status` accepts.
- `test_jwks_refresh.py` covers the refresher's failure paths, which matter more than its success path: a bad fetch must leave the existing keys alone rather than overwrite good keys with unusable ones.
- `osdc/pyproject.toml` gains `pyjwt`/`cryptography` as **test-only** dependencies; the image gets them from apt. I could not regenerate `uv.lock` (no `uv` on my machine) — `just test` runs plain `uv sync`, which resolves, but you may want a refreshed lock committed.

## Known residual

Tokens are replayable until `exp`: `jti` is neither required nor consumed, because doing so needs state shared across dispatcher replicas that v1 does not have. The concurrency cap and namespace quota bound the damage meanwhile.

## Follow-up commits

Beyond the ConfigMap fix above: `model` was accepted and then silently dropped, so a caller could ask for a cheap model, be charged for the policy's, and never be told — it is now refused the way `repo` already was, and the README no longer documents either as a caller's choice. The deploy comment claimed a failed key refresh warns rather than aborts, but only the `kubectl wait` was guarded and the `kubectl create job` above it would have taken the whole deploy down under `set -euo pipefail`; both are guarded now. And `oidc.py` pointed at a manifest filename that does not exist.


osdc/modules/agent-sandbox/dispatcher/authorize.py
```diff
@@ -0,0 +1,173 @@
...
+# a stage that runs from the default branch after an untrusted stage finishes. See the
+# module README on the residual that leaves: the prompt is caller-controlled, so a
+# workflow that reads PR content can shape it. The Grant is what bounds the damage.
+DENIED_EVENTS = frozenset({"pull_request", "pull_request_target", "pull_request_review", "issue_comment"})
+
+# Verified 2026-08-25: every workflow in pytorch/ciforge runs on `ubuntu-latest`. Keeping
+# this to github-hosted means a token minted on a self-hosted runner — including anything
+# in the arc-runners fleet, which is the namespace that can already reach /run over the
+# network — is not accepted.
+ALLOWED_RUNNER_ENVIRONMENTS = frozenset({"github-hosted"})
```
jeanschmidt: seems like a bug / unintended invariant, please take a look

---

<details>
<summary>🟠 The policy admits only github-hosted runners, but only arc-runners pods can reach /run, so flipping REQUIRE_AUTH leaves no caller that can be authorized (ai-generated section)</summary>

`authorize()` denies any token whose `runner_environment` is not `github-hosted` (authorize.py:144-145). The set of callers that can open a TCP connection to `/run` is exactly the pods in the `arc-runners` namespace, and those are self-hosted Actions runners, whose tokens carry `runner_environment: "self-hosted"`. The two sets are disjoint, so once `REQUIRE_AUTH` is `true` every request is either 401 (no header) or 403 (self-hosted token) — there is no caller the shipped policy can admit. The premise is stated in the diff itself, authorize.py:65-68: "a token minted on a self-hosted runner — including anything in the arc-runners fleet, which is the namespace that can already reach /run over the network — is not accepted." Reachability: `kubernetes/base/dispatcher.yaml:148` is `type: ClusterIP`, and `kubernetes/base/networkpolicy.yaml:36-49` (`sandbox-agent-ingress`) admits ingress to `app: sandbox-dispatcher` only `from: namespaceSelector kubernetes.io/metadata.name: arc-runners`, on top of `default-deny` at networkpolicy.yaml:11-22. There is no Ingress, LoadBalancer or NodePort anywhere in `kubernetes/base/`, so a github-hosted runner on GitHub's infrastructure cannot reach the Service at all. The one in-repo consumer confirms the shape: `osdc/integration-tests/workflows/integration-test.yaml.tpl:1969` runs `test-agent-sandbox` on `runs-on: {group: RUNNER_GROUP, labels: [...l-x86iamx-8-32]}` (a self-hosted OSDC runner) and posts to `$SANDBOX/run` at lines 1990 and 2051 with no Authorization header. Meanwhile `kubernetes/base/dispatcher.yaml:94-96` says "FLIP THIS TO 'true' once the GHA client is sending tokens. It is the whole point of the OIDC work", and README:108-109 says "flipping it to `true` is the point of this work, not an optional extra". The migration the change exists to enable cannot be performed. An operator who follows the instruction in dispatcher.yaml and flips `REQUIRE_AUTH` to `true` takes `/run` from working-for-everyone to working-for-nobody: the canary job `test-agent-sandbox` fails on its first `curl -fsS`, and any future GHA client that mints a token on the OSDC fleet gets a 403 naming the runner environment. Recovering means either widening the policy (the thing authorize.py:65-68 says is deliberately excluded) or building a network path that does not exist in this module. Narrow: state the blocker where the flip is documented — add to README's "Enabling enforcement" section and to the dispatcher.yaml comment that no caller can currently satisfy `ALLOWED_RUNNER_ENVIRONMENTS`, so the flag must not be flipped until a caller path exists. Broader, and the one I would pick: decide the caller shape in this change. If the caller is a `pytorch/ciforge` workflow running on the OSDC self-hosted fleet, add `"self-hosted"` to `ALLOWED_RUNNER_ENVIRONMENTS` and rely on `repository_id`/`repository_owner_id` plus the `workflow_ref`/`job_workflow_ref` prefix to distinguish it from any other pod in `arc-runners` — those are the checks that actually do that work, not the runner-environment one. If the caller really is github-hosted, the change needs the ingress path that lets it reach the ClusterIP.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/README.md
```diff
@@ -97,17 +173,17 @@ just deploy-module meta-staging-aws-ue1 agent-sandbox             # IRSA + proxy
 # -m 900: the call waits for the task, and a cold fleet waits for a Karpenter node.
 curl -fsS -m 900 -X POST http://sandbox-agent.ai-sandbox.svc.cluster.local:8080/run \
   -H 'Content-Type: application/json' \
-  -d '{"repo":"pytorch/pytorch","ref":"main","task":"Summarize the build layout",
-       "model":"us.anthropic.claude-haiku-4-5-20251001-v1:0"}'
+  -d '{"ref":"main","task":"Summarize the build layout"}'
```
jeanschmidt, line 213: probably not intentional?

---

<details>
<summary>🟡 deploy.sh still prints a /run example sending `repo`, the copy of the same example in README.md was updated to drop it (ai-generated section)</summary>

The module keeps the same operator-facing curl example in two places: the README "Use it" block and the echo block deploy.sh prints when a deploy finishes. This change updated the README copy (README.md:176, :180) to drop `repo`, and left the deploy.sh copy untouched, so the two now describe different APIs. README.md:183-184 now states "The caller no longer picks the repository or the model... sending either is a 403". deploy.sh:232 still prints `-d '{"repo":"pytorch/pytorch","ref":"main","task":"..."}'` and deploy.sh:234 prints `-d '{"repo":"...","wait":false}'`. http_api.py:205-210 answers 403 when `spec.get("repo")` differs from `grant.clone_repo`, which is `authorize.V1_CLONE_REPO = "pytorch/pytorch"` (authorize.py:52) — so the literal second example is refused as printed. The last thing a deploy prints is a command that now returns `403 the repository to clone is set by policy (pytorch/pytorch), not by the request`, contradicting the README the same change just corrected; the operator has no way to tell from that text which copy is current. Update the echo block in deploy.sh:231-235 to match README.md:174-181: `-d '{\"ref\":\"main\",\"task\":\"...\"}'` and `-d '{\"wait\":false}'`. If you would rather not maintain the example twice, replace the deploy.sh block with a one-line pointer to the module README.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/deploy.sh
```diff
@@ -167,6 +167,26 @@ kubectl kustomize "$MODULE_DIR/kubernetes/base/" \
...
+# deploys do not collide.
+#
+# BOTH steps are guarded, and that is the point: under `set -euo pipefail` an unguarded
+# `kubectl create job` aborts the whole deploy, which is not what a failed key fetch
+# deserves — everything else has already applied and the CronJob retries on its own. An
+# earlier revision guarded only the wait and claimed in this comment that a failure
+# warns; it did not.
+JWKS_JOB="jwks-refresh-deploy-$(date +%s)"
+echo "[agent-sandbox] Fetching OIDC signing keys (${JWKS_JOB})..."
+if kubectl create job "$JWKS_JOB" --from=cronjob/jwks-refresh -n "$NAMESPACE"; then
```
jeanschmidt: this can become a problem very fast

---

<details>
<summary>🟡 Every deploy creates a jwks-refresh Job that nothing ever reaps, so Job and Pod objects accumulate in ai-sandbox without a ceiling (ai-generated section)</summary>

`kubectl create job --from=cronjob/...` builds a standalone Job from `cronJob.Spec.JobTemplate.Spec` with no ownerReference back to the CronJob, so the CronJob controller's `successfulJobsHistoryLimit: 1` / `failedJobsHistoryLimit: 3` (oidc.yaml:108-109) never see it, and the jobTemplate sets `backoffLimit` and `activeDeadlineSeconds` but no `ttlSecondsAfterFinished` (oidc.yaml:114-116), so the TTL controller never sees it either. Each `just deploy-module &lt;cluster&gt; agent-sandbox` therefore leaves one completed Job plus its Succeeded Pod behind permanently, with a unique `$(date +%s)` name so they never overwrite each other. deploy.sh:181 mints `JWKS_JOB="jwks-refresh-deploy-$(date +%s)"` and deploy.sh:183 creates it; nothing later in deploy.sh deletes it — the prune block at deploy.sh:201-203 only removes three named legacy objects. The jobTemplate at oidc.yaml:113-116 carries no `ttlSecondsAfterFinished`, unlike the task Jobs the same module creates (`kube.py:143` sets `ttlSecondsAfterFinished: JOB_TTL_S`, asserted at test_dispatcher.py:156) and unlike the dispatcher's own belt-and-braces cleanup (`kube.delete_job`, kube.py:237-246). Grepping the repo, `ttlSecondsAfterFinished` appears only in kube.py — this is the only `kubectl create job --from=cronjob` in the tree, so there is no sibling pattern that already handles it. One orphan Job object and one Succeeded Pod object accumulate in `ai-sandbox` per module deploy, per cluster, forever — after 200 deploys `kubectl get jobs,pods -n ai-sandbox` returns 200 entries of deploy history mixed in with live task pods, and an operator diagnosing a stuck task has to filter them out by hand. It does not become an outage: the ResourceQuota (quota.yaml:25) caps `pods: "20"`, and that counter excludes pods whose phase is Succeeded or Failed, and `kube.task_result` lists pods under a `job-name=` label selector (kube.py:218-221), so neither the quota nor the request path is affected. The cost is etcd objects and operator noise that grow with deploy count and never shrink. Narrow, and the one I would pick: add `ttlSecondsAfterFinished: 600` to `jobTemplate.spec` in oidc.yaml beside `backoffLimit: 2` (line 115). `kubectl create job --from=cronjob` copies that spec verbatim, so the one line reaps the manual deploy Jobs and the CronJob's own history alike. Broader, if you also want the deploy wall-clock back: guard the whole block on the keys being absent — `if ! kubectl get configmap oidc-jwks -n "$NAMESPACE" -o jsonpath='{.data.jwks\.json}' | grep -q .; then ... fi` — since after the first deploy the CronJob has refreshed within 6 h and `oidc.JWKS_MAX_AGE_S` is 24 h, so the create-and-wait-up-to-120s at deploy.sh:183-185 is redundant work on every deploy after the first.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/authorize.py
```diff
@@ -0,0 +1,173 @@
+"""Who may call, and what they get. The only file here that decides policy.
```
jeanschmidt: probably not intentional?

---

<details>
<summary>🟡 authorize.py's docstring says it is the only file that decides policy and a reviewer can read it and stop, but http_api.py decides the anonymous caller's grant (ai-generated section)</summary>

The module docstring claims this is "The only file here that decides policy" and that "A reviewer asking 'what is a caller allowed to make happen?' reads this file and stops". In the shipped configuration the most permissive decision is not made here: http_api.py:165-176 decides that a caller with no `Authorization` header gets a full v1 `Grant`, bypassing every check in this file. http_api.py:165-176 constructs `authorize.Grant(caller="unauthenticated", clone_repo=authorize.V1_CLONE_REPO, model=authorize.V1_MODEL, …)` without calling `authorize()`; `REQUIRE_AUTH` ships `false` (http_api.py:62, kubernetes/base/dispatcher.yaml `REQUIRE_AUTH: "false"`). Nothing in authorize.py mentions that path, so a reader who "reads this file and stops" concludes only `pytorch/ciforge` on a protected github-hosted ref may dispatch. A reviewer or auditor following this docstring's instruction reaches the wrong conclusion about who may spend the Bedrock budget in the configuration that actually ships, which is the exact question the file claims to answer completely. Add one sentence to the docstring: "While `REQUIRE_AUTH` is false, http_api.py grants a request with no Authorization header the same v1 values without consulting this file — see the REQUIRE_AUTH comment there." Alternatively move the migration-window Grant behind a named function in this module (e.g. `unauthenticated_grant()`) so the claim becomes true again; I would do both.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/authorize.py
```diff
@@ -0,0 +1,173 @@
...
+    {
+        "name": "pytorch/ciforge",
+        "repository_id": "1133856973",
+        "repository_owner_id": "21003710",  # the pytorch organisation
+    },
+)
+
+# The workflow files inside an allowed repo whose tokens are accepted. A prefix, because
+# GitHub spells these `owner/repo/.github/workflows/file.yml@refs/heads/main`.
+ALLOWED_WORKFLOW_PREFIX = "pytorch/ciforge/"
```
jeanschmidt: either we should document or find a way to have a single list

---

<details>
<summary>🟡 ALLOWED_WORKFLOW_PREFIX repeats the caller slug as a single global, so a second entry in ALLOWED_CALLERS would be denied at the workflow check (ai-generated section)</summary>

`ALLOWED_CALLERS` (authorize.py:37-43) is a tuple iterated by `_lookup_caller` (:95-102), i.e. a structure built to hold more than one caller, but the workflow checks at :138 and :141 compare against one module-global prefix that repeats the same repo slug as `ALLOWED_CALLERS[0]​["name"]`. The two copies must agree, and only one of them scales with the list. `"name": "pytorch/ciforge"` at authorize.py:39 and `ALLOWED_WORKFLOW_PREFIX = "pytorch/ciforge/"` at :47 are the same value written twice. test_authorize.py drives the entire allow/deny table from the single ciforge identity (GOOD_CLAIMS, test_authorize.py:24-32), so adding a caller fails no test. Onboarding a second caller looks like a one-line addition to ALLOWED_CALLERS; that caller then passes `_lookup_caller` and is refused at :138 with "workflow is not on the allow-list", a message that names neither the prefix nor the reason, with the whole suite still green. In the other direction, once two entries exist, caller A's token carrying a `workflow_ref` inside caller B's repo satisfies the global prefix. Move the prefix into each allow-list entry (`"workflow_prefix": "pytorch/ciforge/"`) and compare against `caller["workflow_prefix"]` at both :138 and :141. The narrower alternative is deriving `f"{caller['name']}/"` after the lookup; I would take the per-entry field, since it keeps the allow-list one readable table and closes the A-presents-B's-workflow case at the same time.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/authorize.py
```diff
@@ -0,0 +1,173 @@
...
+
+@dataclass(frozen=True)
+class Grant:
+    """Everything the run is permitted to use, decided before any Job exists.
+
+    Frozen, because the whole point is that nothing downstream may edit it. The Job
+    builder takes one of these and never sees the request body.
+    """
+
+    caller: str  # "pytorch/ciforge" — for the audit line, not for any decision
```
jeanschmidt: probably not intentional?

---

<details>
<summary>🟡 `Grant.caller` is the access-control key /status compares, while its comment says it is for the audit line and no decision (ai-generated section)</summary>

The field is labelled as carrying no decision, but it is the authorization subject for reads. `do_POST` passes it to `tasks.start_task(grant.caller)` (http_api.py:220) where it becomes the entry's `owner`, and `tasks.status()` (tasks.py:134) compares that stored owner against the identity recomputed on every `GET /status` to decide whether a result is revealed. So `Grant.caller` does two jobs — audit string and ownership key — and only the first is stated where the field is defined. http_api.py:220 `task_id = tasks.start_task(grant.caller)`; tasks.py:134 `if task is None or task.get("owner") != owner: return None`; http_api.py:156 recomputes the same value on each /status via `authorize.authorize(...).caller`. No test covers an authenticated /status read: in test_dispatcher.py, `test_status_does_not_expose_another_callers_task` (line 629) and `test_a_caller_can_read_its_own_task` (line 638) both use the hardcoded literal "unauthenticated" and send no token, and `test_status_applies_the_same_policy_as_run` (line 643) uses a token that authorize *denies*, so the success branch at http_api.py:156 never returns a value in the suite. A maintainer who makes the audit identity more specific — folding in `workflow_ref` or a run id, which http_api.py:227 currently appends separately precisely because `caller` is coarse — gives each freshly minted token a different owner. Every `wait: false` caller then gets `404 unknown task id` for its own result once REQUIRE_AUTH is flipped, and the entire suite stays green because no test does an authenticated POST followed by /status. Making it coarser instead silently widens who can read whose results. Narrow: restate the comment as "the task-ownership key /status compares — changing its shape changes who can read a result", and add a test that POSTs `{"wait": false}` under a real token and then GETs /status with a second token from the same caller. Broader: add a separate `audit: str` field for the log line and keep `caller` solely as the ownership key, so the two cannot be conflated. I would do the narrow one plus the test.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/authorize.py
```diff
@@ -0,0 +1,173 @@
...
+    # traffic while passing all of its unit tests, because two fixtures disagreed about
+    # this claim's type and each suite validated its own belief. Assert the string.
+    if claims.get("ref_protected") != "true":
+        raise Denied("only a protected ref may dispatch agent tasks")
+
+    clone_repo, model = policy(caller) if policy else (V1_CLONE_REPO, V1_MODEL)
+
+    # The request contributes the prompt and the commit to read. Everything else about
+    # the run is policy: a caller cannot name a repository to clone or a model to spend,
+    # and `repo` in the body is ignored rather than validated, so there is no version of
```
jeanschmidt: probably not intentional?

---

<details>
<summary>🟡 authorize.py says `repo` in the body "is ignored rather than validated"; http_api.py validates it and answers 403 on a mismatch (ai-generated section)</summary>

This comment asserts that `repo` is ignored rather than validated, "so there is no version of this where a validation slip lets one through". The shipped code does validate it: http_api.py:205-210 compares `spec["repo"]` against `grant.clone_repo` and answers 403. The same file's rule 1 at authorize.py:16 ("Nothing downstream of authorize() reads the request body") is falsified by the same two comparisons. http_api.py:205 `if spec.get("repo") and spec["repo"] != grant.clone_repo:` → 403, and http_api.py:212 does the same for `model`; http_api.py:131-134 states the opposite of this comment ("It is still type-checked, and still compared to the Grant below"); README.md:91-93 and test_dispatcher.py::test_the_caller_cannot_choose_the_repository also document validation-then-403. The comment appears to predate the follow-up commit that added the refusal. The file a reviewer is told to read for the policy states that the request's `repo` is never compared to anything, and reasons from that to conclude no validation slip is possible. Someone auditing for exactly that slip skips http_api.py:205-214, where the comparison actually lives. Change the sentence to: "and `repo` and `model` in the body never choose anything — http_api.py compares them to the Grant and refuses a disagreement with 403 rather than substituting." Also soften rule 1 at line 16 to "Nothing downstream of authorize() takes a value from the request body: the Job is built from the Grant.".

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/http_api.py
```diff
@@ -0,0 +1,249 @@
...
+            length = int(raw_length)
+        except ValueError:
+            raise ValueError(f"invalid Content-Length: {raw_length!r}") from None
+        if not 0 <= length <= MAX_BODY_BYTES:
+            raise ValueError(f"Content-Length must be between 0 and {MAX_BODY_BYTES}, got {length}")
+
+        spec = json.loads(self.rfile.read(length) or "{}")
+        if not isinstance(spec, dict):
+            raise ValueError("body must be a JSON object")
+        # 'repo' is no longer required, and no longer decides anything: the repository to
```
jeanschmidt, line 137: probably not intentional?

---

<details>
<summary>🟡 http_api.py's module docstring still documents `repo` as a required /run field and `model` as a caller choice, and still says the file never decides anything (ai-generated section)</summary>

Two sentences in the module docstring are falsified by this change. Line 5 declares `POST /run -&gt; body {"repo","ref","task","model"?,"wait"?}` — `repo` without `?`, i.e. required, and `model` as a caller-supplied field. Lines 10-11 declare "the rule for this one is that it never decides anything: it validates the shape of a request and hands it on", while the file now authenticates, authorizes, answers 403 on a repo/model mismatch, and builds the migration-window Grant itself. The added comment at line 131 says "'repo' is no longer required", and the `missing 'repo'` check was deleted from `_read_spec`; http_api.py:205-214 issues two policy 403s; http_api.py:169-176 constructs a `Grant` from `authorize.V1_CLONE_REPO`/`V1_MODEL`. README.md:62 was updated to `{"ref"?,"task"?,"wait"?}`; this docstring, which is the same contract stated 126 lines above the code, was not. The docstring is the first thing a reader of the dispatcher opens. It states a request contract (`repo` mandatory, `model` a caller choice) that returns 400/403 in practice, and a layering rule the file no longer keeps, so the next person to add a decision here has no signal that the rule already broke. Update line 5 to `POST /run -&gt; body {"ref"?,"task"?,"wait"?} ("repo"/"model" accepted only when they match policy)` and add the 401/403 outcomes; replace "it never decides anything" at lines 10-11 with "it decides only who is asking (oidc + authorize) and refuses a request that contradicts the Grant; task logic and policy values live elsewhere".

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/http_api.py
```diff
@@ -0,0 +1,249 @@
...
+        # Grant below — a caller that names a different repo is told so rather than
+        # quietly getting the policy's one.
+        for key in ("repo", "ref", "task", "model"):
+            if key in spec and not isinstance(spec[key], str):
+                raise ValueError(f"'{key}' must be a string")
+        if "wait" in spec and not isinstance(spec["wait"], bool):
+            raise ValueError("'wait' must be a boolean")
+        return spec
+
+    def _caller(self) -> str:
```
jeanschmidt, line 148: maybe a refactor here to avoid duplication?

---

<details>
<summary>🟡 _caller() restates the authentication preamble of _grant_for() rather than calling it, so /status and /run hold two copies of one rule (ai-generated section)</summary>

`_caller()` (http_api.py:142-156) and `_grant_for()` (:158-178) implement the same two-branch rule — no header plus `not REQUIRE_AUTH` means the `"unauthenticated"` identity, otherwise `authorize.authorize(oidc.verify(oidc.bearer_token(header)), ...)`. `_caller()` is exactly `self._grant_for({}).caller`: with `spec={}` the else branches are identical calls and the no-header branch produces a Grant whose `caller` is the same literal. The identity string `"unauthenticated"` is therefore also written twice (:151 and :170), and tasks.status compares the two for equality. http_api.py:150-151 and :165-176 are the two copies; the comment at :152-155 states the invariant explicitly ("The SAME check /run runs"). tasks.py:134 (`task.get("owner") != owner`) is what makes /status depend on both copies producing byte-identical identities. The module otherwise funnels one rule through one implementation: http_api.py:172-173 reads `authorize.V1_CLONE_REPO`/`V1_MODEL` instead of restating them, and tasks.py:141-149 routes both `run_and_record` and `run_in_background` through `run_to_completion` rather than each owning the loop. A later change to the no-header branch — tightening REQUIRE_AUTH, adding a rate limit, renaming the anonymous identity — applied to `_grant_for` alone leaves /status on the old rule, and no test covers that: test_dispatcher.py's `test_requiring_auth_refuses_a_request_with_no_token` exercises /run only, and `test_status_applies_the_same_policy_as_run` covers only the token-present path. Collapse the body of `_caller()` to `return self._grant_for({}).caller`, keeping the existing docstring and the comment about /status running the same check as /run. That removes the second copy of the REQUIRE_AUTH branch and the second `"unauthenticated"` literal in one edit.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/oidc.py
```diff
@@ -0,0 +1,179 @@
...
+def _check_age(fetched_at: float) -> None:
+    age = time.time() - fetched_at
+    if age > JWKS_MAX_AGE_S:
+        raise InvalidToken(f"signing keys are {int(age)}s old (max {JWKS_MAX_AGE_S}s) — the JWKS refresher has stopped")
+    if age < -FUTURE_SKEW_S:
+        raise InvalidToken(f"signing keys are timestamped {int(-age)}s in the future")
+
+
+def _load_keyset() -> PyJWKSet:
+    """The mounted JWKS, re-read when the file changes or the interval elapses.
```
jeanschmidt: probably not intentional?

---

<details>
<summary>🟡 `_load_keyset`'s docstring says the key set is re-read when the file changes; there is no file-change detection, only the interval (ai-generated section)</summary>

The docstring promises two reload triggers: a file change or the elapsed interval. The body implements only the second — the cached branch at oidc.py:103 returns the cached keyset whenever `now - _CACHE["loaded_at"] &lt; JWKS_RELOAD_INTERVAL_S`, without stat-ing the file or comparing its content, so a rotation lands up to 60 s late and never earlier. oidc.py:100-108: `now = time.monotonic()` … `if cached is not None and now - _CACHE["loaded_at"] &lt; JWKS_RELOAD_INTERVAL_S: … return cached` — no `st_mtime`, no re-read. test_oidc.py::test_the_key_set_is_not_re_read_on_every_request deletes the mounted file and asserts the second `verify()` still succeeds, which pins the absence of any file-change trigger. The constant's own comment at oidc.py:64-66 states the real behaviour ("re-read at most this often"). Someone debugging why a freshly written ConfigMap did not take effect trusts the docstring, concludes the mount or the refresher is broken, and looks in the wrong place for up to a minute; someone tuning `JWKS_RELOAD_INTERVAL_S` upward believes a content change would still be picked up promptly. Change the first line to "The mounted JWKS, re-read at most once per JWKS_RELOAD_INTERVAL_S seconds." (optionally noting that a rotation is therefore visible within one interval).

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/oidc.py
```diff
@@ -0,0 +1,179 @@
...
+            return cached
+
+        try:
+            document = json.loads(JWKS_PATH.read_text())
+        except OSError as exc:
+            raise InvalidToken(f"no signing keys available: {exc}") from None
+        except ValueError as exc:
+            raise InvalidToken(f"signing keys are unreadable: {exc}") from None
+
+        fetched_at = document.get("fetched_at")
```
jeanschmidt, line 129: better handle it, to avoid pain debugging in the future

---

<details>
<summary>🟡 A JWKS file whose top-level JSON is not an object raises AttributeError out of _load_keyset, dropping the connection instead of answering 401 (ai-generated section)</summary>

The try block at lines 110-115 catches `OSError` and `ValueError`, so it covers a missing file and malformed JSON — but valid JSON that is not an object (`[]`, `null`, `3`, `"x"`) parses fine and then `document.get` raises `AttributeError`. The same hole exists three lines later: `PyJWKSet.from_dict(document["jwks"])` at line 123 raises `AttributeError` if `document["jwks"]` is not a dict, and the except tuple at line 124 lists `ValueError, KeyError, TypeError, jwt.PyJWKSetError` but not `AttributeError`. `AttributeError` is neither `InvalidToken` nor `Denied`, the only two exceptions caught in http_api.py:96-103 (`do_GET`) and http_api.py:193-200 (`do_POST`), so it escapes the handler and `BaseHTTPRequestHandler` closes the connection with no response. kube.py:108-113 already documents and fixes this exact failure class for a different exception type: "ValueError is neither ApiError nor OSError, so it would slip past every caller's except clause, escape do_POST, and close the connection with no response — the one answer this endpoint exists to avoid." test_oidc.py's `test_an_unparseable_key_file_refuses` states the intended contract as "must fail closed, not raise something no caller catches", but only exercises `"{ not json"`, which is a ValueError. If the oidc-jwks ConfigMap ever holds a non-object under `jwks.json` — a hand-edit, a future writer, a restore from a bad backup — every /run and /status request that presents a token drops its connection with a traceback in the dispatcher log, instead of the 401 the rest of this file is written to produce. The refresher only ever writes an object, so this is not reachable from the normal path today. Add `if not isinstance(document, dict): raise InvalidToken("signing keys are unreadable: not a JSON object")` immediately after the `json.loads` at line 111, and add `AttributeError` to the except tuple at line 124 so a non-dict `document["jwks"]` is also converted into an InvalidToken.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/test_authorize.py
```diff
@@ -0,0 +1,173 @@
...
+        authorize_fn(claims(), {"task": {"$ref": "something"}}, policy)
+
+
+def test_the_allow_list_is_code_not_configuration(monkeypatch):
+    """It is the answer to "who may spend our Bedrock budget", so it belongs in git
+    history and code review. An env var would move that decision out of both."""
+    monkeypatch.setenv("ALLOWED_CALLERS", "attacker/repo")
+    monkeypatch.setenv("ALLOWED_WORKFLOW_PREFIX", "attacker/repo/")
+    with pytest.raises(Denied):
+        authorize_fn(claims(repository_id="999"), {}, None)
```
jeanschmidt: probably not intentional?

---

<details>
<summary>🟡 test_the_allow_list_is_code_not_configuration is denied by the module constants either way, so the two env vars it sets test nothing (ai-generated section)</summary>

The test sets `ALLOWED_CALLERS` and `ALLOWED_WORKFLOW_PREFIX` and then asserts `Denied` for `repository_id="999"` — claims that the module constants deny regardless. The setenv calls are inert (authorize.py imports only `dataclasses`; there is no `os` import and no `environ` reference anywhere in it), and even a future version that did read `ALLOWED_CALLERS` would not match id "999" against the value `"attacker/repo"`, so the assertion would still hold. The test cannot fail for the reason its name gives. authorize.py:23-25 is the whole import block: `from __future__ import annotations` and `from dataclasses import dataclass`. The assertion is the same call and the same expectation as the already-present parametrised case at test_authorize.py:88, `({"repository_id": "999"}, "a different repo in the same org")`, run through `authorize_fn(claims(**override), {}, policy)` at line 96. The guard the test advertises — "moving the allow-list into an env var must fail here" — does not exist. Someone doing exactly that refactor gets green CI plus a test whose name and docstring say the property is protected, which is worse than having no test at all. Make the environment value one that would grant if it were read, so the test fails when the constant becomes configuration: `monkeypatch.setenv("ALLOWED_CALLERS", json.dumps([{"name": "attacker/repo", "repository_id": "999", "repository_owner_id": "999"}]))` and call `authorize_fn(claims(repository_id="999", repository_owner_id="999"), {}, None)`. Note this only detects a JSON-shaped env implementation; the blunt alternative that detects any of them is to assert the source holds no environment lookup, e.g. `assert "environ" not in Path(authorize.__file__).read_text()`. I would take the first and accept its narrowness, or drop the test as redundant with line 88.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/kubernetes/base/oidc.yaml
```diff
@@ -0,0 +1,201 @@
...
+                  cpu: "50m"
+                  memory: "64Mi"
+                  ephemeral-storage: "100Mi"
+                limits:
+                  cpu: "200m"
+                  memory: "128Mi"
+                  ephemeral-storage: "100Mi"
+
+---
+# The refresher's egress: DNS, the API server, and HTTPS out. It is the only pod in this
```
jeanschmidt: probably not intentional?

---

<details>
<summary>🟡 The jwks-refresher NetworkPolicy comment claims it and the sigv4 proxy are the only pods in the namespace allowed off-cluster; task pods have 443 to anywhere (ai-generated section)</summary>

The comment states the refresher "is the only pod in this namespace besides the sigv4 proxy that is allowed off-cluster". `sandbox-task-egress` in the sibling manifest allows every task pod TCP 443 to `::/0` and `0.0.0.0/0` for the git clone, so task pods are a third — and the widest — class allowed off-cluster. The comment also mis-states its own rule: this policy allows DNS and 443, not the API server. kubernetes/base/networkpolicy.yaml:65-102, `sandbox-task-egress`, `podSelector: app: sandbox-task`, egress `ipBlock: "::/0"` and `0.0.0.0/0` on port 443, with its own comment at :58 calling it "the widest thing in this namespace — 443 to anywhere is an exfiltration path". README.md:299-302 documents the same. This file's later comment at oidc.yaml:191-193 already acknowledges "the same honest limitation the task-egress rule documents", contradicting line 165-166. The egress rules below (oidc.yaml:181-201) list only kube-system DNS and 443, no API-server CIDR. Someone reasoning about which pods in `ai-sandbox` can reach the internet — the question this comment exists to answer — reads that it is two and gets a materially wrong picture of the namespace's exfiltration surface. The API-server clause also sends a reader looking for a rule that is not there. Rewrite as: "# The refresher's egress: DNS and HTTPS out. Off-cluster egress in this namespace is the sigv4 proxy, the task pods (443 for the clone, see sandbox-task-egress) and this job; it is allowed out only because it holds no RBAC worth stealing beyond patching one ConfigMap.".

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/kubernetes/base/oidc.yaml
```diff
@@ -0,0 +1,201 @@
...
+# namespace besides the sigv4 proxy that is allowed off-cluster, and the only reason it
+# is allowed is that it holds no RBAC worth stealing.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: jwks-refresher-egress
```
jeanschmidt, line 197: better to keep a standard, if possible

---

<details>
<summary>🟡 jwks-refresher-egress is the only NetworkPolicy in the repository that does not live in a networkpolicy.yaml (ai-generated section)</summary>

oidc.yaml:165-201 declares a NetworkPolicy alongside the ConfigMap, ServiceAccount, Role, RoleBinding and CronJob. Everywhere else in this repository NetworkPolicies are collected in a file named networkpolicy.yaml, including the policies for workloads whose manifests live in sibling files of the same kustomize base. A repo-wide grep for `kind: NetworkPolicy` returns exactly four files: agent-sandbox/kubernetes/base/networkpolicy.yaml, buildkit/kubernetes/base/networkpolicy.yaml, pypi-cache/kubernetes/networkpolicy.yaml — and this one. This module's networkpolicy.yaml holds all six of its policies, including sandbox-agent-ingress (:29) and sandbox-dispatcher-egress (:115) for the Deployment declared in dispatcher.yaml, and proxy-ingress (:147) / proxy-egress (:172) for the Deployment declared in sigv4-proxy.yaml. buildkit does the same for its four policies while its workload sits in haproxy.yaml. The namespace's egress surface is no longer readable in one file. networkpolicy.yaml:170-171 still asserts the proxy rule is "the only sanctioned egress out of the namespace", and the stated plan to close the 443-to-anywhere rules (networkpolicy.yaml:52-61, README.md:299-302) would be carried out in that file and would leave the third such rule, at oidc.yaml:194-201, wide open. Move the jwks-refresher-egress document (oidc.yaml:164-201, comment included) into kubernetes/base/networkpolicy.yaml next to proxy-egress, leaving the ConfigMap, ServiceAccount, Role, RoleBinding and CronJob in oidc.yaml. No other change is needed — kustomization.yaml already lists both files.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/pyproject.toml
```diff
@@ -12,7 +12,15 @@ dependencies = [
...
 
 [dependency-groups]
-dev = ["pytest>=8.0", "pytest-cov>=5.0", "pytest-xdist>=3.5"]
+# pyjwt/cryptography are TEST-only here: the agent-sandbox dispatcher gets them from
+# apt inside its own image, and nothing in this repo imports them at deploy time.
+dev = [
+  "pytest>=8.0",
+  "pytest-cov>=5.0",
+  "pytest-xdist>=3.5",
+  "pyjwt>=2.0",
```
jeanschmidt, line 21: for consistency and reproduceability

---

<details>
<summary>🟡 pyjwt and cryptography are added to pyproject.toml but the committed uv.lock is not regenerated (ai-generated section)</summary>

osdc/uv.lock is the committed resolution of osdc/pyproject.toml, and the two are now out of agreement: the lock's `[package.dev-dependencies]` (uv.lock:292-297) and `[package.metadata.requires-dev]` (:308-313) list exactly pytest, pytest-cov and pytest-xdist, with no entry for the two dependencies added at pyproject.toml:21-22. Searching osdc/uv.lock for `name = "pyjwt"` or `name = "cryptography"` returns nothing, while every runtime dependency from pyproject.toml:6-12 has a matching entry at uv.lock:300-306. Every CI job installs with plain `uv sync` (.github/workflows/osdc-test.yml:43, osdc-lint.yml:43, _osdc-deploy.yml:79), and `just test` (osdc/justfile:2020) runs `uv run pytest`. `uv sync` re-resolves and rewrites the lock inside the runner, so the pyjwt/cryptography versions the test suite actually runs against are whatever resolves on the day rather than anything pinned in git; locally, the next `uv` invocation in this tree leaves an unrelated uv.lock diff for whoever runs it next to explain. It also means the new dependencies never pass through the review that the lock exists to give them. Run `uv lock` in osdc/ and commit the updated osdc/uv.lock alongside this change.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/README.md
```diff
@@ -59,16 +59,92 @@ N task pods, 3 fit per fleet node, and a pending pod adds one. The ceiling is
...
 
 - `GET /healthz` → `{"status":"ok","in_flight":int,"capacity":int}`
-- `POST /run` body `{"repo","ref","task","model"?,"wait"?}` →
+- `POST /run` body `{"ref"?,"task"?,"wait"?}` →
   `{"task_id":str,"cloned":bool,"file_count":int,"top_level":[str],"report":str,"errors":{…}}`
 
   Waits for the task by default, so a caller sees the result on the same connection —
   budget for a cold fleet, where the pod waits on a Karpenter node. `"wait": false`
-  returns `202 {"task_id"}` instead.
+  returns `202 {"task_id"}` instead. `repo` and `model` are still accepted for
```
jeanschmidt: nit, should be an easy fix

---

<details>
<summary>⚪ The README states two contradictory rules for `repo`/`model`, and no `model` value can satisfy the one it documents (ai-generated section)</summary>

Two passages added by this diff state opposite contracts, and each is true for only one of the two fields. Lines 67-69 say `repo` and `model` are "still accepted for compatibility but must agree with policy" — false for `model`, because `authorize.V1_MODEL` is the empty sentinel, so `grant.model` is `""` and `http_api.py:212` refuses every non-empty `model` string; there is no value that agrees. Lines 183-184 say "sending either is a `403`" — false for `repo`, which has exactly one accepted value. authorize.py:53 `V1_MODEL = "" # empty means "the dispatcher's configured default"`; http_api.py:212 `if spec.get("model") and spec["model"] != grant.model:` → 403 for any non-empty string. The model actually used comes from elsewhere: kube.py:161 injects `BEDROCK_DEFAULT_MODEL_ID` from the Deployment env, and agent/sandbox.py:26,213 falls back to it when `SANDBOX_MODEL` is empty — so a caller sending the cluster's configured model id sends the value policy uses and still gets a 403. For `repo`, test_dispatcher.py's `test_naming_the_policy_repository_is_still_accepted` asserts `_post(f"{server}/run", {"repo": authorize.V1_CLONE_REPO})["report"] == "ok"`, directly contradicting README:184. Separately, the pull request description claims "the README no longer documents either as a caller's choice", but README:244-245 still reads "the model id comes from the caller's `/run` body, so that boundary is the one bounding what can be invoked" — that sentence is the stated justification for the shape of the Bedrock IAM policy. An existing caller reading the Endpoints entry concludes it can keep sending `model` as long as it matches policy, sends the deployed `BEDROCK_DEFAULT_MODEL_ID`, and gets `403 the model is set by policy, not by the request` with no value that works. Anyone revisiting the Bedrock IAM policy from README:244 reasons from a request-body boundary that this change removed. In the Endpoints entry, split the two fields: `repo` is accepted only when it equals the policy repository, and any `model` at all is refused in v1. Correct README:184 ("sending either is a `403`") to match, and update README:244-245 so the IAM paragraph names the Grant rather than the request body. Alternatively, if `model` compatibility is genuinely wanted, compare against the effective model (`kube.DEFAULT_MODEL`) when `grant.model` is empty instead of against the `""` sentinel — I would keep the current refuse-everything behaviour, which the tests pin, and fix the prose.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/README.md
```diff
@@ -59,16 +59,92 @@ N task pods, 3 fit per fleet node, and a pending pod adds one. The ceiling is
...
+  call* below for why the request does not get to choose them.
   `top_level` is the clone's real top-level listing, which is also fed to the
   model — an empty one means the report was not grounded in the repo.
 - `GET /status/<task_id>` → `{"state":"running"}` or `{"state":"done", …result}`.
-  Results are kept in memory for an hour after the task finishes.
+  Results are kept in memory for an hour after the task finishes. A task belonging to
+  another caller answers `404`, not `403` — otherwise the endpoint would confirm that
+  other callers are running tasks.
+
+## Who may call, and what a call can do
```
jeanschmidt, line 81: nit, should be an easy fix

---

<details>
<summary>⚪ The Limitations section still states `/run` has no notion of who is asking and that caller identity does not exist, which this change adds (ai-generated section)</summary>

This change adds a whole "Who may call" section documenting OIDC authentication and an authorization policy, but the Limitations bullet at README.md:293-298 — the section headed "read before trusting it" — still says "`/run` has no notion of who is asking: any pod in `arc-runners` can call it, and the NetworkPolicy is the only gate" and "A caller identity and per-request ephemerality are the same piece of work and neither exists". README.md:294 and :297-298 (unchanged by this diff) versus dispatcher/authorize.py (new, decides the caller), http_api.py:142-156 (`_caller()` derives an identity for /status), http_api.py:220 `tasks.start_task(grant.caller)` and tasks.py:134 (`task.get("owner") != owner`). Caller identity now exists and the NetworkPolicy is no longer the only gate for a caller that presents a token. The two halves of the README disagree about whether caller identity exists. A reader who consults Limitations to decide what protections are absent is told to build one that this change already shipped, and the item most in need of tracking — that enforcement is off by default — is not the one the bullet names. Rewrite the bullet to describe the actual residual: "**Enforcement is off by default.** `/run` authenticates and authorizes a presented token (see *Who may call*), but `REQUIRE_AUTH` ships `false`, so a request with no `Authorization` header is still admitted and any pod in `arc-runners` can call it. Flip the flag once the GHA client sends tokens." Drop the "caller identity … does not exist" clause.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/README.md
```diff
@@ -59,16 +59,92 @@ N task pods, 3 fit per fleet node, and a pending pod adds one. The ceiling is
...
+The dispatcher holds create-Job RBAC, so it is the component that must not be able to
+reach the internet — its NetworkPolicy allows DNS and the Kubernetes API and nothing
+else. GitHub's signing keys therefore arrive as a mounted ConfigMap, refreshed every six
+hours by a CronJob (`kubernetes/base/oidc.yaml`) whose Role can patch that one named
+object and nothing more.
+
+The manifest declares that ConfigMap with **no `data`** — the content belongs to the
+refresher. That is load-bearing rather than tidy: seeding it in the manifest meant every
+`kubectl apply` put the seed back over live keys, so each deploy blanked them. `deploy.sh`
+runs a refresh immediately after applying, so the window with no keys is seconds rather
```
jeanschmidt: nit, should be an easy fix

---

<details>
<summary>⚪ README says the window with no signing keys is "seconds"; oidc.yaml says "a minute or two" for the same window and explains why seconds is wrong (ai-generated section)</summary>

Two prose passages added by this change give different magnitudes for the same window between `kubectl apply` and the dispatcher seeing keys. The README says "seconds"; kubernetes/base/oidc.yaml:33-35 says "a minute or two" and states the reason the README's figure cannot hold — a ConfigMap volume updates on the kubelet's sync period, not instantly. README.md:135-137 versus kubernetes/base/oidc.yaml:33-35 ("that window is a minute or two (a ConfigMap volume updates on the kubelet's sync period, not instantly) rather than up to six hours"). deploy.sh:184 additionally allows the refresh Job itself up to `--timeout=120s`. An operator watching a first deploy after enabling `REQUIRE_AUTH` expects authenticated requests to work within seconds, sees 401s for a minute or more, and starts debugging a system that is behaving as designed. Change "the window with no keys is seconds" to "the window with no keys is a minute or two (the mount updates on the kubelet's sync period)", matching oidc.yaml:34.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/README.md
```diff
@@ -59,16 +59,92 @@ N task pods, 3 fit per fleet node, and a pending pod adds one. The ceiling is
...
+decoration: a ConfigMap volume only updates when its content changes and GitHub rotates
+rarely, so judging freshness by file mtime would age out a perfectly healthy key set
+while a refresher dead for a month looked identical. The dispatcher refuses keys older
+than 24h, which is what turns a silently dead refresher into a loud failure.
+
+Task pods declare no volumes at all, and `kubernetes/base/admissionpolicy.yaml` enforces
+that in the API server. **Adding one is a policy change as well as a manifest change** —
+the `GITHUB_TOKEN` init-container work is the case this is waiting for.
 
 ## The two images
```
jeanschmidt, line 183: nit, should be an easy fix

---

<details>
<summary>⚪ README still says both images are "stdlib-only today", which this change makes false for the dispatcher image (ai-generated section)</summary>

The bullet at README.md:159-161, in the section that starts on this line, ends "both are stdlib-only today". This change installs `python3-jwt` and `python3-cryptography` into the dispatcher image, so only the task image is stdlib-only now. The Dockerfile's own header comment was rewritten for this (it no longer says "This one is stdlib-only too"); the README's copy of the same claim was not. dispatcher/Dockerfile:30-31 adds `python3-cryptography` and `python3-jwt` to the apt install; dispatcher/Dockerfile:9-11 replaces the old "This one is stdlib-only too (urllib against the in-cluster API...)" with "it is why adding a JWT library here costs the sandbox nothing". README.md:161 still reads "both are stdlib-only today". (The stale line is 12 lines below this heading and falls outside the diff hunk, which ends at line 150.). The README's "The two images" section now understates the dispatcher image's dependency surface, which is the one property the section exists to describe. Change README.md:161 to say the task image is stdlib-only and the dispatcher image adds PyJWT and cryptography from apt, which is the whole reason the two images are separate.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/dispatcher/test_admissionpolicy.py
```diff
@@ -359,7 +367,22 @@ def test_the_job_pass_is_scoped_to_the_dispatchers_service_account(policy):
 
 
 def _digest(expression: str) -> str:
-    return hashlib.sha256(re.sub(r"\s+", " ", expression).strip().encode()).hexdigest()[:12]
+    """The EXACT parsed expression, not a whitespace-normalised one.
+
+    Normalising first would have collapsed whitespace inside CEL string literals too, so
+    a real semantic change like 'a  b' -> 'a b' would keep its digest. YAML has already
+    folded these block scalars by the time we see them, so re-indenting or reflowing an
```
jeanschmidt: nit, should be an easy fix

---

<details>
<summary>⚪ The _digest docstring claims re-indenting cannot move a digest, but five of this diff's re-pinned digests are exactly the expressions where it can (ai-generated section)</summary>

YAML folding of a `&gt;-` scalar joins lines with a single space only while they sit at the block's own indentation; a *more-indented* continuation line is kept literally, newline and leading spaces included. Five of the policy's expressions are written that way, so their parsed strings still carry `\n` plus indentation, and re-indenting or reflowing one does move its digest — the opposite of what this sentence tells the next maintainer. The five digests this diff re-pins ("...allowPrivilegeEscalation: false and runAsNonRoot: true...", "...not add Linux capabilities", "...not unmask /proc", "...cpu, memory and ephemeral-storage limits", "...request exactly what they limit") are precisely the five whose continuation lines are indented past the first line — admissionpolicy.yaml:169-176, 177-182, 183-188, 204-210, 222-226. Every uniformly-indented multi-line expression kept its digest across the same change: "...not share host namespaces" (admissionpolicy.yaml:112-115) is still `1cafda1b80d7`, "task pods must set no sysctls" (128-132) is still `64b91d2f9f8f`, "...app: sandbox-task label" (249-254) is still `ab1151632d34`. That split is only possible if the parsed strings for the first group retain their newlines and indentation. Someone reflowing one of those five CEL expressions with no semantic change gets a failure from test_no_expression_changed_without_its_mirror_being_rechecked telling them the expression changed, after this docstring told them it would not. Replace the last sentence with something accurate, e.g. "YAML folds same-indent continuation lines into single spaces, but keeps the newline and indentation of a more-indented line — so reflowing one of the multi-line expressions does move its digest even with no semantic change.".

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

osdc/modules/agent-sandbox/kubernetes/base/oidc.yaml
```diff
@@ -0,0 +1,201 @@
...
+    namespace: ai-sandbox
+
+---
+# Every 6 hours. The dispatcher refuses keys older than 24h, so this has four chances to
+# succeed before a failing refresher becomes an outage — enough to ride out a GitHub
+# blip, short enough that a persistent failure is not silent for a week.
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: jwks-refresh
```
jeanschmidt: nit, should be an easy fix

---

<details>
<summary>⚪ The CronJob is named jwks-refresh while its ServiceAccount, Role, RoleBinding, NetworkPolicy and pod label are all jwks-refresher (ai-generated section)</summary>

Every object belonging to this component is named `jwks-refresher` except the CronJob itself, which drops the trailing `er`. Both other CronJobs in the repository, and both workloads in this module, use one name across the workload and its RBAC. oidc.yaml:48 (ServiceAccount), :57 (Role), :74 (RoleBinding), :171 (NetworkPolicy), :120 (pod label) and :124 (serviceAccountName) are all `jwks-refresher`; only :94 is `jwks-refresh`. zombie-cleanup/kubernetes/cronjob.yaml:4 and :24 both say `zombie-cleanup`; harbor-cache-recovery/kubernetes/cronjob.yaml:4 and :24 both say `harbor-cache-recovery`. Inside this module, rbac.yaml:14/:23/:51 and dispatcher.yaml:17 all say `sandbox-dispatcher`. `kubectl get cronjob jwks-refresher` — the name every other object in the file uses — returns NotFound; deploy.sh:183 has to spell the odd one out, and a later runbook or smoke test that guesses the consistent name fails against a healthy cluster. Rename the CronJob to `jwks-refresher` at oidc.yaml:94 and update `--from=cronjob/jwks-refresh` at deploy.sh:183. Those are the only two references, and since the object has not shipped there is nothing to prune from existing clusters.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 0a9a77006.</sub>

</details>

jeanschmidt (changes requested):
Overall not any big concern. But this PR will quickly become a landmine for agents that read docs, comments and docstrings

## Review summary

🟠 1 major · 🟡 14 minor · ⚪ 6 nit

- 🟠 The policy admits only github-hosted runners, but only arc-runners pods can reach /run, so flipping REQUIRE_AUTH leaves no caller that can be authorized — `osdc/modules/agent-sandbox/dispatcher/authorize.py:69`
- 🟡 deploy.sh still prints a /run example sending `repo`, the copy of the same example in README.md was updated to drop it — `osdc/modules/agent-sandbox/README.md:176`
- 🟡 Every deploy creates a jwks-refresh Job that nothing ever reaps, so Job and Pod objects accumulate in ai-sandbox without a ceiling — `osdc/modules/agent-sandbox/deploy.sh:183`
- 🟡 authorize.py's docstring says it is the only file that decides policy and a reviewer can read it and stop, but http_api.py decides the anonymous caller's grant — `osdc/modules/agent-sandbox/dispatcher/authorize.py:1`
- 🟡 ALLOWED_WORKFLOW_PREFIX repeats the caller slug as a single global, so a second entry in ALLOWED_CALLERS would be denied at the workflow check — `osdc/modules/agent-sandbox/dispatcher/authorize.py:47`
- 🟡 `Grant.caller` is the access-control key /status compares, while its comment says it is for the audit line and no decision — `osdc/modules/agent-sandbox/dispatcher/authorize.py:84`
- 🟡 authorize.py says `repo` in the body "is ignored rather than validated"; http_api.py validates it and answers 403 on a mismatch — `osdc/modules/agent-sandbox/dispatcher/authorize.py:159`
- 🟡 http_api.py's module docstring still documents `repo` as a required /run field and `model` as a caller choice, and still says the file never decides anything — `osdc/modules/agent-sandbox/dispatcher/http_api.py:131`
- 🟡 _caller() restates the authentication preamble of _grant_for() rather than calling it, so /status and /run hold two copies of one rule — `osdc/modules/agent-sandbox/dispatcher/http_api.py:142`
- 🟡 `_load_keyset`'s docstring says the key set is re-read when the file changes; there is no file-change detection, only the interval — `osdc/modules/agent-sandbox/dispatcher/oidc.py:95`
- 🟡 A JWKS file whose top-level JSON is not an object raises AttributeError out of _load_keyset, dropping the connection instead of answering 401 — `osdc/modules/agent-sandbox/dispatcher/oidc.py:117`
- 🟡 test_the_allow_list_is_code_not_configuration is denied by the module constants either way, so the two env vars it sets test nothing — `osdc/modules/agent-sandbox/dispatcher/test_authorize.py:163`
- 🟡 The jwks-refresher NetworkPolicy comment claims it and the sigv4 proxy are the only pods in the namespace allowed off-cluster; task pods have 443 to anywhere — `osdc/modules/agent-sandbox/kubernetes/base/oidc.yaml:165`
- 🟡 jwks-refresher-egress is the only NetworkPolicy in the repository that does not live in a networkpolicy.yaml — `osdc/modules/agent-sandbox/kubernetes/base/oidc.yaml:171`
- 🟡 pyjwt and cryptography are added to pyproject.toml but the committed uv.lock is not regenerated — `osdc/pyproject.toml:21`
- ⚪ The README states two contradictory rules for `repo`/`model`, and no `model` value can satisfy the one it documents — `osdc/modules/agent-sandbox/README.md:67`
- ⚪ The Limitations section still states `/run` has no notion of who is asking and that caller identity does not exist, which this change adds — `osdc/modules/agent-sandbox/README.md:77`
- ⚪ README says the window with no signing keys is "seconds"; oidc.yaml says "a minute or two" for the same window and explains why seconds is wrong — `osdc/modules/agent-sandbox/README.md:136`
- ⚪ README still says both images are "stdlib-only today", which this change makes false for the dispatcher image — `osdc/modules/agent-sandbox/README.md:149`
- ⚪ The _digest docstring claims re-indenting cannot move a digest, but five of this diff's re-pinned digests are exactly the expressions where it can — `osdc/modules/agent-sandbox/dispatcher/test_admissionpolicy.py:374`
- ⚪ The CronJob is named jwks-refresh while its ServiceAccount, Role, RoleBinding, NetworkPolicy and pod label are all jwks-refresher — `osdc/modules/agent-sandbox/kubernetes/base/oidc.yaml:94`

<details>

**What this changes**

The dispatcher gains GitHub Actions OIDC authentication and an authorization policy held in code. `oidc.py` verifies RS256 tokens against a JWKS mounted from a ConfigMap, with an in-document `fetched_at` and a 24h staleness bound; `authorize.py` turns verified claims into a frozen `Grant`, and `kube.job_manifest` is re-signatured to take that `Grant` instead of the request body, so the clone repository and model are policy rather than request. Tasks now carry an owner and `/status` reports another caller's task as absent. A CronJob with a Role scoped to one named ConfigMap keeps the keys current, and `deploy.sh` triggers one refresh after apply. Enforcement ships off: with `REQUIRE_AUTH=false`, a request with no `Authorization` header still receives a full v1 Grant built in `http_api._grant_for`.

**What it gets right**

The verifier is configured, not trusted: `algorithms=["RS256"]` is fixed, required claims go through the library's `options={"require": ...}`, and `test_oidc.py` drives real signatures — a hand-assembled HS256 token signed with the public key, a `kid` naming one key with another's signature, a rotated key picked up without a restart. The staleness bound is re-checked on the cached path in `_load_keyset`, not only on load, which is the window a naive cache leaves open. The refresher parses with `PyJWKSet` before writing and leaves the ConfigMap untouched on any failure, so an unloadable key set cannot replace a working one. `_flag` aborts at import on an unrecognised `REQUIRE_AUTH`. `_finish` carries the owner across, so finished results do not become readable by everyone.

**Where the risk is**

One behavioural gap and a large band of prose drift. The gap is `ALLOWED_RUNNER_ENVIRONMENTS = frozenset({"github-hosted"})` in `authorize.py` against `sandbox-agent-ingress`, which admits only the `arc-runners` namespace to a ClusterIP: the admissible set and the reachable set are disjoint, so the flip that `dispatcher.yaml` and the README call the point of the work admits nobody, and the module's own `test-agent-sandbox` job in `integration-test.yaml.tpl` is excluded twice over. Three reviewers reached that region independently.

The drift concentrates in exactly the comments that carry the design argument. `authorize.py`'s docstring claims it is the only file that decides policy and that nothing downstream reads the request body, while the `repo`/`model` refusals live in `do_POST`; the comment above them says `repo` "is ignored rather than validated". That pair is one edit and matters more than its severity: a future dispatch path following the stated rule silently reinstates the model-substitution bug this change just fixed. The README's `repo`/`model` contract is self-contradictory and unsatisfiable for `model`, the `http_api.py` module docstring still documents `repo` as required, the Limitations section still says caller identity does not exist, and the Bedrock IAM paragraph still reasons from a caller-supplied model. `Grant.caller` is labelled "not for any decision" while being the `/status` ownership key, and no test does an authenticated POST followed by a `/status` read, so changing that string breaks reads with the suite green. Separately, three reviewers each landed on the unregenerated `osdc/uv.lock` and on the per-deploy `jwks-refresh-deploy-&lt;epoch&gt;` Job that nothing reaps.

**Before merge, and after**

The runner-environment/reachability disjointness needs a decision on the caller shape, not a one-line edit, and the change ships the instruction to flip the flag. Correct the invariant-bearing comments and README passages in `authorize.py`, `http_api.py` and `README.md`, since they are one line each and they are what a reviewer is told to trust. Regenerate `osdc/uv.lock`.

After: `ttlSecondsAfterFinished` on the CronJob's jobTemplate; the non-object JWKS document that escapes `_load_keyset` as `AttributeError`; the vacuous `test_the_allow_list_is_code_not_configuration` and the missing authenticated `/status` test; the `_digest` docstring against the five re-pinned expressions; per-caller workflow prefix, `_caller()` collapsing to `_grant_for({}).caller`, the NetworkPolicy's file placement, and the `jwks-refresh`/`jwks-refresher` name split.

</details>

izaitsevfb:
@jeanschmidt , please re-review

huydhn (approved):
Stamped to unblock

jeanschmidt (approved):

PR #1046, opened 2026-08-26
Fix the ECR-gate smoke test crashing where the flag is off

huydhn:
`just smoke` fails on every cluster with `harbor.ecr_pullthrough` off — the ConfigMap is correctly absent and the test crashes instead of passing:

```
base/kubernetes/tests/smoke/test_base_kubernetes.py:31: in test_registry_mirror_ecr_gate_matches_cluster_config
    result = run_kubectl(["get", "configmap", "registry-mirror-ecr", "--ignore-not-found"], ...)
E   json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

`--ignore-not-found` prints nothing when the resource is missing, and `run_kubectl`'s default `json_output=True` feeds that empty string to `json.loads`. Reads it raw now and parses only when non-empty.

Affects `meta-staging-aws-uw1`, `meta-staging-aws-ue1`, `meta-prod-aws-ue1`, `lf-prod-aws-ue1`. My mistake in #1035 — I only exercised it on staging-ue2, where the flag is on and the output is valid JSON, so the absent branch never ran.

No infrastructure impact: the ConfigMap's absence on those clusters is correct.

**Tested:** `just test` OK. `just lint` fails only on the pre-existing hadolint finding at `base/docker/runner-base/Dockerfile:36`.

### DIFF
--- a/osdc/base/kubernetes/tests/smoke/test_base_kubernetes.py
+++ b/osdc/base/kubernetes/tests/smoke/test_base_kubernetes.py
@@ -1,5 +1,7 @@
 """Smoke tests for base Kubernetes resources (DaemonSets, StorageClass, nodes)."""
 
+import json
+
 import pytest
 from helpers import assert_daemonset_healthy, filter_services, run_kubectl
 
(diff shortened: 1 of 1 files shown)


huydhn:
This is a bug from my stack https://github.com/pytorch/ci-infra/pull/1035. I deployed and tested it on ue2, but one of the new check wrongly failed on ue1 https://github.com/pytorch/ci-infra/actions/runs/32937209085

zxiiro (approved):

Issue #1016, opened 2026-08-18
TorchTitan long queue on 8x a10g runners


huydhn:
Coordinate a strategy to improve the queue time for 8x a10g runners with TorchTitan team

https://chat.google.com/room/AAQAt6iU48s/YMr1lzP_yQQ/YMr1lzP_yQQ?cls=10

cc @wconstab @felipemello1 @jeanschmidt

huydhn:
cc @georgehong

wconstab:
Cc @wwwjn, who is working on fakepg migration for ci tests in titan

PR #1048, opened 2026-08-27
Add per-cluster Karpenter g6 weight ladder for LF clusters

zxiiro:
Karpenter NodePool weight is scoped fleet-wide today, so every cluster
gets the same launch priority order regardless of what capacity is
actually available to it. g6.48xlarge never launches on LF clusters
(0 successes in 211 CreateFleet attempts across lf-prod-aws-ue1 and
lf-prod-aws-ue2), and the remaining sizes are left in an order that
does not reflect cost per delivered runner-hour.

Add a weight field that can be a per-cluster map (a default plus
per-cluster overrides) instead of a bare int, threaded through the
generator via a new NODEPOOLS_CLUSTER env var. A resolved weight of 0
omits the NodePool for that cluster entirely.

Apply it to modules/nodepools/defs/g6.yaml only, for lf-prod-aws-ue1
and lf-prod-aws-ue2: g6.48xlarge is disabled there, and the remaining
sizes are reordered by measured cost per delivered runner-hour instead
of largest-first. Meta's default ladder is unchanged.


mtarsel (approved):

huydhn:
To provide some additional context here, a large instance is actually cheaper to use at scale as we can squeeze in more jobs with less overhead. But the capacity problem described here is real https://github.com/pytorch/ci-infra/issues/1016, and Meta fleets are observing similar capacity shortage. As LF fleets handles less traffics, it makes sense to try different order.

However, the blast radius is large that the new order is applied to the whole fleet, including those that it shouldn't need to change like `c7i-runner` where we legitimately want bigger node with high density. So, I would propose 2 changes here:

1. Apply this only to some GPU nodes like g5 or g6 where their capacity problem is legit. Leave the rest. This means that a fleet-wide knob like `smallest-first` is not sufficient
2. Check if we can set different weight per cluster (Claude tells me we could, the feature is already there). Then, we don't need to build a new mechanism to support different weight ordering. For example:

```
    - type: g6.48xlarge
      weight:
        default: 100                 # keep largest-first on meta-* clusters
        lf-prod-aws-ue1: 60
        lf-prod-aws-ue2: 60
      node_disk_size: 600
      has_nvme: true
    - type: g6.8xlarge
      weight:
        default: 70
        lf-prod-aws-ue1: 100
        lf-prod-aws-ue2: 100
      node_disk_size: 600
      has_nvme: true
```

huydhn (changes requested):
The blast radius of this change is fleet-wide which is too big, let's do this only for those node with capacity issues like g5/g6

zxiiro:
> The blast radius of this change is fleet-wide which is too big, let's do this only for those node with capacity issues like g5/g6

Okay I reworked this to apply only to the g6 instances for now. I took some data before and will review the data after this is deployed to see how it changes things before deciding if we want to deploy similar changes to other instance types like g5.

huydhn (approved):
LGTM! Ty for the fix!

IMO, there might be a legit need to use smaller instances on LF fleets as they don't receive as much traffics as the Meta ones. But we need to find a way measure this better to find the right balance.

huydhn:
@claude Do a final round of review to see if there is any other issues to add here

zxiiro:
I addressed the minor fixes from Claude's comment, will merge if everything passes the checks.

zxiiro:
The failure was due to being unable to acquire lock file for one of Meta's clusters. Seems unrelated to the change here.

PR #1052, opened 2026-08-31
[grafana] Add merge-retry panels to Core Metrics

georgehong:
Port HUD's "Merge retry rate (avg)" to a daily series. Introduce two panels:
- Direct analog with some slight changes to methodology aimed to be more precise than the HUD query
- Mergebot resistant panel that describes **percent of PRs that need a retry** - i.e., "one in three PRs needed another try"

The following are some chosen changes: 
- Bucket by merge day, not PR-creation day: HUD's cohort requires the PR to have merged, so it is right-censored and reads low on the trailing week off half the PRs.
- Dedupe pull_request before filtering state/labels; it is a ReplacingMergeTree with no version column. Verified equivalent to FINAL (73,313 PRs, zero mismatches); filtering first over-includes 10 PRs whose latest snapshot no longer qualifies.
- Count distinct merge workflow runs; 3.9% of PRs get the same run's "Merge started" comment posted twice.
- Winsorize the mean at 5: uncapped, merge-bot from internal sync constantly attempts to merge which skews the results.

### DIFF
--- a/grafana/pytorch_devx.json
+++ b/grafana/pytorch_devx.json
@@ -1000,7 +1000,7 @@
             "transformations": []
           }
         },
-        "description": "Mean pytorchmergebot merge attempts per landed PR — distinct workflow runs, winsorized at 5. 1.0 = landed first try; y-axis floor is 1.0, so the shaded area is retries. Blue = 7d PR-weighted avg. HUD's 'Merge retry rate (avg)', bucketed by merge day rather than PR-creation day (HUD's cohort is right-censored, ~10% low on the trailing week). Counts CI-driven retries only; ghstack members attribute to the stack's top PR. Amber 1.56 / red 1.78. Today excluded; empty for executorch.",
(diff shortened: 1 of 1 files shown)


georgehong:
Preview link: https://pytorchci.grafana.net/goto/shgnzp. I'm happy to consolidate this into a single chart, but I like that there's one chart that most closely matches the HUD metric and another for intuitively interpreting the common case. 

Part of https://github.com/pytorch/test-infra/issues/8679

grafana/pytorch_devx.json
```diff
@@ -955,6 +955,359 @@
...
+                      "color": "green",
+                      "value": 0
+                    },
+                    {
+                      "color": "#EAB839",
+                      "value": 50
+                    },
+                    {
+                      "color": "red",
+                      "value": 60
```
jathu: How did you arrive at these? 50% of PRs needing a merge retry seems high for a warning threshold

jathu (approved):

grafana/pytorch_devx.json
georgehong: Thresholds were chosen to be roughly stable/within the history for this year, but it's definitely not desirable to need retries so often. I'll update to target 35% or less as green state (and corresponding in the other chart).

PR #1054, opened 2026-09-02
[gpu-dev] Allow allowlisted users unlimited single-GPU extensions

bobrenjc93:
## Summary

- preserve the existing 48-hour extension policy and user experience
- bypass only the backend total-duration check for the authenticated emails `bobren@meta.com` and `huydhn@meta.com` when they reserve one GPU
- retain the 48-hour cap for non-allowlisted users and multi-GPU reservations

## Test plan

- `.venv/bin/pytest -m "not integration"` (1219 passed, 3 expected xfails)
- focused extension policy tests (4 passed)
- Ruff on the new tests and changed Lambda lines
- Black on the new tests


huydhn:
Allow unlimited reservation kind of defeat what we vision how gpu-dev are used in the first place. Ideally, we want to allow people to keep their reservation, but automatically detect if it's in use, and free up the resource if it's not.

As I don't have any data point to prove or disprove this. I agree with the suggestion what we keep the mechanism, but add an allow list where we can give people in that list more time, unlimited number of extension, for example.

huydhn (approved):

PR #1055, opened 2026-09-02
[gpu-dev] Don't block on (or restore from) third-party snapshot copies

huydhn:
Port of wdvr/osdc#227 now that the code lives here. Follow-up to the non-blocking snapshot retry (wdvr/osdc#226, already in this tree): that stopped the Lambda *blocking* on a pending snapshot; this fixes *which* snapshots we treat as ours in the first place.

## Why

`CopySnapshot` copies the source snapshot's tags onto the copy. The Wiz security scanner copies gpu-dev disks, so its copies arrive tagged `gpu-dev-user` + `disk_name` + `ManagedBy=gpu-dev-cli`, and our disk logic treats them as ours. Two failures, both observed in prod:

**1. Blocking.** The pending-snapshot gate matches the scanner's copy, which can sit pending for hours:

```
snap-093f829edf4210c90
  role:        arn:aws:iam::308535385114:role/WizScannerRole
  description: auto-generated by Wiz
  CopySnapshot    2026-09-02 01:33:41 UTC
  DeleteSnapshot  2026-09-02 17:48:38 UTC   (lifetime 16h15m)
```

That one blocked **12 reservation attempts** on disk `default` in 24h — most recently eellison@meta.com on an H100, who sat through ~40 min of Lambda-timeout retries before it cleared. An earlier copy cost drisspg@meta.com 3 failed reservations on 08-17 and armed the same trap for 8 users on 08-18.

**2. Silent stale restores** — the more dangerous one. A *completed* copy also matches, and its `StartTime` is the copy time, so it can win `max(StartTime)` and be picked as the restore source, returning a point-in-time older than it appears. Replayed against prod data, **7 of the 8** affected users would have restored from a scanner copy on their next reservation.

## What

- `is_foreign_snapshot()` — ignore snapshots we didn't author, in **both** the pending gate and the restore-source selection.
- Only wait on a pending snapshot **newer** than the newest completed one. An older pending snapshot has nothing to offer; the 08-17 blocker was 6h15m older than a complete, ready restore point.

Matched on copy markers (`wiz*` tags, `VolumeId == vol-ffffffff`, description) rather than requiring a gpu-dev tag. A whitelist on the `SnapshotType` tag looked cleaner, but 133 of 6612 real snapshots don't carry it — including 5 hand-made recovery snapshots (mreso, nshulga, ezyang) it would have orphaned.

## Validation

Against all 6612 production snapshots: **6596/6596 real kept, 16/16 scanner copies excluded**, zero false exclusions.

Replaying the new logic over the 8 users affected by the 08-18 scan batch — all go blocked → ok, and 7 get corrected restore sources:

```
user                    before      after       restore source
drisspg@meta.com        BLOCKED     ok          snap-0f2342ae1205868a8
simonlayton@meta.com    BLOCKED     ok          snap-0dd5f5005e704886e  (was a WIZ COPY)
eellison@meta.com       BLOCKED     ok          snap-01ef00259731e6c5c  (was a WIZ COPY)
vasiliy@meta.com        BLOCKED     ok          snap-092b7e618ec648584  (was a WIZ COPY)
tarindu@meta.com        BLOCKED     ok          snap-04b77d2f3f3917576  (was a WIZ COPY)
zainhuda@meta.com       BLOCKED     ok          snap-08fd9cb46fbf06400  (was a WIZ COPY)
calebkim@meta.com       BLOCKED     ok          snap-03b6b318de711ee99  (was a WIZ COPY)
felipemello@meta.com    BLOCKED     ok          snap-04c57fe289b87a8d6  (was a WIZ COPY)
```

## Test plan

Written TDD — 5 new tests in `gpu-dev/tests/unit/lambda_fn/test_foreign_snapshots.py`, all confirmed failing first (one reproduces the stale-restore path as `assert 'snap-wiz' == 'snap-ours'`).

```
cd gpu-dev && env -u SSH_AUTH_SOCK uv run pytest -m "not integration"
# 1224 passed, 3 xfailed
```

`tofu validate` not run — empty local provider cache; the tf change is a one-character `LAMBDA_VERSION` bump to 0.6.7.

## Deploy note

**Needs `tofu apply` on prod.** Prod is currently running lambda **0.6.5 (LastModified 2026-06-16)** — so the already-merged non-blocking retry (0.6.6) has never shipped either. Confirmed by the user-visible message: prod still emits the pre-0.6.6 wording "Waiting for disk snapshot to complete (from previous session)". Worth checking whether other merged lambda changes are also sitting undeployed.

Until then, `--no-persist` bypasses the gate entirely.

## Out of scope

- The unbounded `queued` state from the non-blocking-retry change: a genuinely stuck snapshot now queues forever with no deadline, where it previously failed after ~45 min. Separate PR.
- `shared/snapshot_utils.py::get_latest_snapshot` has the same hole. `migrate_ebs_across_az` (`index.py:877`) is the nearest relative and has no callers — dead code. (Corrected: an earlier version of this description cited a `get_latest_user_snapshot` wrapper, which does not exist in this repo; that note was ported from the osdc version unverified.)
- Read-side guard in Step 2b for already-persisted `clone_source_snapshot` references — a semantics decision, see review thread.
- The unbounded `queued` state inherited from the non-blocking-retry change: a genuinely stuck snapshot now queues forever with no deadline, where it previously failed after ~45 min.
- The Wiz tag-inheritance behaviour is worth raising with that integration's owners — it affects anything keying off `gpu-dev-user`/`disk_name`, not just this gate.


huydhn:
cc @drisspg @eellison

gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py
```diff
@@ -111,6 +111,28 @@ class SnapshotPendingError(RuntimeError):
     """A disk snapshot is still pending and allocation should be retried later."""
 
 
+def is_foreign_snapshot(snapshot: dict[str, Any]) -> bool:
```
jeanschmidt: Maybe we should consider this?

---

<details>
<summary>🟠 Clone-source selection in process_clone_disk_action still picks scanner copies, so a clone can record a Wiz copy as its permanent clone_source_snapshot (ai-generated section)</summary>

The description promises the new helper is applied "in **both** the pending gate and the restore-source selection", and enumerates the remaining holes under "Out of scope" (only `shared/snapshot_utils.py::get_latest_snapshot`). A third site is neither fixed nor listed: `process_clone_disk_action` picks the clone source with the identical unfiltered `max(StartTime)` over snapshots tagged `gpu-dev-user` + `disk_name`, filtering only `delete-date`. That site is not touched by this diff, but its output is consumed by the very function this diff changes. index.py:10270-10273 builds `active_snapshots` filtering only `delete-date`; index.py:10313 `source_snapshot = max(active_snapshots, key=lambda s: s['StartTime'])`; index.py:10345 persists it as `'clone_source_snapshot': source_snapshot_id`. Step 2b of the changed function reads it back at index.py:7472 and assigns it to `latest_snapshot` at index.py:7490, which then feeds `create_volume(SnapshotId=...)` at index.py:7516 with no `is_foreign_snapshot` check. The path is live: the SQS dispatcher calls `process_clone_disk_action` at index.py:2127-2128. This is the same `max(StartTime)` mechanism the description calls "the more dangerous one". A user who clones a disk while a scanner copy is the newest snapshot of the source disk gets a clone whose recorded source is that copy: first the clone restores an older point-in-time than it appears, and then, once the scanner deletes its copy (16h15m in the PR's own example), every reservation for that cloned disk fails hard and permanently — `describe_snapshots` raises InvalidSnapshot.NotFound, index.py:7496-7501 re-raises, and index.py:7858-7862's named-disk branch marks the reservation failed. Recovery needs a manual DynamoDB edit. Narrow fix: add the filter to the comprehension at index.py:10270-10273, e.g. `if 'delete-date' not in {...} and not is_foreign_snapshot(s)`. Broader fix I would also take: guard the read side in step 2b by checking `is_foreign_snapshot(clone_snapshot)` before assigning it to `latest_snapshot` at index.py:7490 — that also protects clone entries already written to DynamoDB, which the write-side fix cannot repair. If you intend to leave both for a follow-up, say so in the Out of scope list the way you did for `get_latest_snapshot`.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 4a3e88242.</sub>

</details>

gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py
```diff
@@ -7420,15 +7418,49 @@ def create_disk_from_snapshot_or_empty(user_id: str, availability_zone: str, dis
...
 
+        # Now check for a pending snapshot of our own (from a recent expiry)
+        pending_filters = [
+            {"Name": "tag:gpu-dev-user", "Values": [user_id]},
+            {"Name": "status", "Values": ["pending"]},
+        ]
+        if disk_name:
+            pending_filters.append({"Name": "tag:disk_name", "Values": [disk_name]})
+
+        pending_response = ec2_client.describe_snapshots(
```
jeanschmidt: great point, so maybe we need a different approach or a more elaborate control

---

<details>
<summary>🟠 Moving the pending-snapshot query after the completed query opens a window where a just-completed snapshot is invisible to both queries (ai-generated section)</summary>

`create_disk_from_snapshot_or_empty` now runs the paginated `status=completed` query first (index.py:7410-7419) and the `status=pending` query second (index.py:7438). A snapshot that is still pending when the completed query is evaluated, but has transitioned to completed by the time the pending query is evaluated, appears in neither result set. `latest_snapshot` is then the previous session's snapshot and the volume is restored from it, silently. The old ordering had no such gap: the pending query ran strictly first, so any in-flight snapshot was either caught as pending (raise `SnapshotPendingError`, retry later) or, if it had already completed, was returned by the completed query that ran afterwards. Order in the head file: pages consumed at index.py:7418-7419 (`for page in page_iterator: snapshots.extend(...)`), then `pending_response = ec2_client.describe_snapshots(...)` at index.py:7438. The removed code issued the pending `describe_snapshots` before the paginator (the `-` hunk at base lines 7377-7396). Nothing re-reads the completed set after the pending query, and the restore at index.py:7504-7516 uses `latest_snapshot` computed at 7428. The path is retried about once a minute for a queued reservation (`aws_cloudwatch_event_rule.reservation_processor_schedule`, `rate(1 minute)`, lambda.tf:385-404), so each waiting user takes many draws at the window. The window width itself (one API round trip between the two evaluations) is inferred from the call ordering, not measured. When the timing lands in the window, the user's most recent session snapshot is never restored: the disk comes back at the previous session's point-in-time with no warning and no error, and the skipped snapshot is thereafter older than the one taken at the end of the new session, so the lost work is only recoverable by hand. This is the same silent-stale-restore failure the PR sets out to eliminate. Issue the pending `describe_snapshots` call before the paginated completed query and keep the comparison where it is: move lines 7431-7441 above line 7410, leaving the `if pending_snapshots:` evaluation at 7443. With pending queried first, a snapshot that completes in between shows up in the completed set, and the existing `latest_pending['StartTime'] &lt;= latest_snapshot['StartTime']` test (equality, same snapshot in both lists) correctly declines to block on it.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 4a3e88242.</sub>

</details>

gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py
jeanschmidt: I guess there are more failure modes...

---

<details>
<summary>🟡 is_foreign_snapshot() is named as a general ownership check but implements three Wiz-specific markers, so any other copier reads as ours (ai-generated section)</summary>

The name and the docstring's first line state a general predicate ("produced by something other than gpu-dev"), but the body is a three-marker deny-list for one specific tool: a tag key starting with `wiz`, `VolumeId == "vol-ffffffff"`, and the literal description `auto-generated by Wiz`. Everything else answers "ours". The two call sites (index.py:7425 and index.py:7445) read as `not is_foreign_snapshot(snap)`, i.e. as a general ownership test, and neither the call sites nor the exclusion itself emit anything when a snapshot is dropped. index.py:128-133 is the whole implementation: `any(key.lower().startswith("wiz") ...)`, `snapshot.get("VolumeId") == "vol-ffffffff"`, `"auto-generated by Wiz" in (...Description...)`. Two of the three markers are the string "wiz". The tag inheritance the PR describes is a property of `CopySnapshot`, not of Wiz: any copier of a gpu-dev snapshot gets `gpu-dev-user` + `disk_name` + `ManagedBy=gpu-dev-cli` and therefore matches the filters at index.py:7403-7407 and index.py:7432-7436. Meanwhile the sibling decision one block down does log its outcome (index.py:7451-7454, "Ignoring pending snapshot ... older than completed ..."), so the foreign-exclusion is the only silent decision in a function that logs every other branch. When a second copier appears (AWS Backup, DLM, a replacement scanner, a manual `aws ec2 copy-snapshot`), its copies are classified as ours and the exact prod incident recurs: a pending copy gates reservations, a completed copy wins `max(StartTime)` and silently restores stale data. Whoever debugs it reads `not is_foreign_snapshot(snap)` at the call site, concludes copies are already excluded, and looks elsewhere. Symmetrically, if the heuristic ever misfires on a real gpu-dev snapshot, the user silently gets an older restore point or an empty disk with no log line explaining why. Narrow fix: rename to `is_wiz_copy` (or `is_scanner_copy`) so the call sites state what is actually detected, and log the exclusion, e.g. in the loop at index.py:7423-7426 `logger.info(f"Excluding third-party copy {snap['SnapshotId']} from restore candidates")` before skipping. Broader fix: keep the deny-list but assert positive ownership too (e.g. require `SnapshotType`, `Project=gpu-dev-servers`, or a description starting with `gpu-dev `, with the hand-made recovery snapshots grandfathered by id) so unknown copiers default to excluded rather than included. I would take the narrow fix now — it costs nothing and removes the false sense of generality.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 4a3e88242.</sub>

</details>

gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py
```diff
@@ -111,6 +111,28 @@ class SnapshotPendingError(RuntimeError):
...
+    """Whether a snapshot was produced by something other than gpu-dev.
+
+    `CopySnapshot` copies the source snapshot's tags onto the copy, so
+    `tag:gpu-dev-user` does not prove we authored a snapshot -- a security
+    scanner copying a user's disk yields one tagged as ours. Such copies must
+    never gate a reservation (they can stay pending for hours) nor be used as a
+    restore source (they hold a point-in-time older than their StartTime, so
+    they can win `max(StartTime)` and silently restore stale data).
+
+    Matched on copy markers rather than a required gpu-dev tag: ~2% of real
```
jeanschmidt: worth fixing the docstring

---

<details>
<summary>🟡 is_foreign_snapshot's docstring blames missing SnapshotType tags on hand-made recovery snapshots; the live clone path also creates snapshots without it (ai-generated section)</summary>

The added docstring records why a `SnapshotType`-tag whitelist was rejected, and attributes every tagless snapshot to hand-made recovery: "~2% of real snapshots are hand-made recovery ones carrying no SnapshotType tag". That attribution is wrong about this repository's own code: `process_clone_disk_action` creates live-clone snapshots with no `SnapshotType` tag, so gpu-dev itself keeps producing untagged snapshots. index.py:10290-10300 (`process_clone_disk_action`, dispatched live from index.py:2128) calls `ec2_client.create_snapshot` with only `gpu-dev-user`, `disk_name` and `Name` tags — no `SnapshotType`. The dead `migrate_ebs_across_az` at index.py:902-914 is the same. By contrast `safe_create_snapshot` in lambda/shared/snapshot_utils.py:57-63 always sets `{"Key": "SnapshotType", ...}`, so the shutdown/cancellation snapshots the docstring implicitly contrasts against are the tagged ones and the clone path is an untagged producer the sentence does not account for. No behaviour breaks today. The cost lands on the next person who revisits this filter: the docstring is the only record of the whitelist decision, and it tells them the tagless set is a fixed historical handful of hand-made snapshots. Tagging those and switching to a `SnapshotType` whitelist would then silently exclude every clone-created snapshot, so a cloned disk's first reservation would restore an empty volume instead of the clone source. Reword lines 124-125 to name the live producer, e.g. "Matched on copy markers rather than a required gpu-dev tag: not everything we author carries SnapshotType -- process_clone_disk_action's live-clone snapshots omit it, as do ~2% of existing snapshots (including hand-made recovery ones)." The broader alternative is to also add a SnapshotType tag at index.py:10293-10300 so a whitelist becomes viable later; I would take the reword now, since it keeps the recorded rationale accurate without changing behaviour.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 4a3e88242.</sub>

</details>

gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py
```diff
@@ -7420,15 +7418,49 @@ def create_disk_from_snapshot_or_empty(user_id: str, availability_zone: str, dis
         for page in page_iterator:
             snapshots.extend(page.get('Snapshots', []))
 
-        # Filter out soft-deleted snapshots (those with delete-date tag)
+        # Drop soft-deleted ones (delete-date tag) and third-party copies
         active_snapshots = []
         for snap in snapshots:
             tags = {tag['Key']: tag['Value'] for tag in snap.get('Tags', [])}
-            if 'delete-date' not in tags:
+            if 'delete-date' not in tags and not is_foreign_snapshot(snap):
```
jeanschmidt: yeah, this failure mode is bad...

---

<details>
<summary>🟡 `process_clone_disk_action` picks its clone source with the same max(StartTime) logic but does not get the new `is_foreign_snapshot` filter (ai-generated section)</summary>

`create_disk_from_snapshot_or_empty` and `process_clone_disk_action` are twin implementations of the same selection: paginate `describe_snapshots` on `tag:gpu-dev-user` + `tag:disk_name` + `status=completed`, drop entries carrying `delete-date`, then take `max(StartTime)`. This change hardens only the first copy (lines 7421-7428) with `is_foreign_snapshot`; the second copy (lines 10269-10273, feeding `max(StartTime)` at 10313) is left byte-for-byte on the old rule, and nothing links the two. index.py:7421-7428 now reads `if 'delete-date' not in tags and not is_foreign_snapshot(snap)`. index.py:10269-10273 still reads `if 'delete-date' not in {t['Key']​: t['Value'] for t in s.get('Tags', [])}` over an identical filter set built at 10252-10256, and 10313 does `source_snapshot = max(active_snapshots, key=lambda s: s['StartTime'])`. That id is persisted as `clone_source_snapshot` (index.py:10345) and later consumed unfiltered by the very function this change edits, at index.py:7489-7490 (`latest_snapshot = clone_snapshot`). The path is user-reachable: `gpu-dev disk clone` (cli.py:5446) and `client.clone_disk` (sdk/python/src/gpu_dev/_sync/client.py:300) both enqueue `action: clone_disk` (disks.py:576, _backend/aws.py:378), dispatched at index.py:2127-2128. A third copy of the same loop lives in shared/snapshot_utils.py:311-324 (`get_latest_snapshot`), also unfiltered. A user who clones a disk while a scanner copy is that disk's newest completed snapshot gets `clone_source_snapshot` pinned to the copy. Unlike the restore path, that id is written to DynamoDB, so every future reservation on the cloned disk restores the same stale point-in-time via index.py:7490 — the failure the PR body measured on 7 of 8 users, now durable rather than transient. It also means the next person reading either block has no way to tell the two must agree. Narrow: add the same predicate to index.py:10270-10273, e.g. `active_snapshots = [s for s in snapshots if 'delete-date' not in {t['Key']​: t['Value'] for t in s.get('Tags', [])} and not is_foreign_snapshot(s)]`. Broader: move `is_foreign_snapshot` into `shared/snapshot_utils.py` beside `get_latest_snapshot` and call it from all three selection sites (index.py:7425, index.py:10270, snapshot_utils.py:315), which also puts it where the expiry lambda can reach it. I would apply the narrow fix in this PR — it closes the clone path now — and do the move separately.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 4a3e88242.</sub>

</details>

jeanschmidt (changes requested):
## Review summary

🟠 2 major · 🟡 3 minor

- 🟠 Clone-source selection in process_clone_disk_action still picks scanner copies, so a clone can record a Wiz copy as its permanent clone_source_snapshot — `gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py:114`
- 🟠 Moving the pending-snapshot query after the completed query opens a window where a just-completed snapshot is invisible to both queries — `gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py:7438`
- 🟡 is_foreign_snapshot() is named as a general ownership check but implements three Wiz-specific markers, so any other copier reads as ours — `gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py:114`
- 🟡 is_foreign_snapshot's docstring blames missing SnapshotType tags on hand-made recovery snapshots; the live clone path also creates snapshots without it — `gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py:124`
- 🟡 `process_clone_disk_action` picks its clone source with the same max(StartTime) logic but does not get the new `is_foreign_snapshot` filter — `gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py:7425`

<details>

**What this changes**

The reservation-processor Lambda gains a module-level helper, `is_foreign_snapshot()` at index.py:114, which classifies a snapshot as third-party on three markers: a tag key beginning `wiz`, `VolumeId == "vol-ffffffff"`, or the description `auto-generated by Wiz`. `create_disk_from_snapshot_or_empty` applies it at two points, the completed-snapshot filter (index.py:7425) and the pending list (index.py:7445). The same function's two EC2 queries swap order so completed runs first, and the pending gate now raises `SnapshotPendingError` only when the newest pending snapshot is strictly newer than the newest completed one. `LAMBDA_VERSION` moves 0.6.6 to 0.6.7, and five unit tests are added.

**What it gets right**

The `StartTime` comparison keeps the case that matters working: a genuine in-flight snapshot still gates, and `test_pending_snapshot_newer_than_completed_still_blocks` pins that so the loosened gate cannot silently become a no-op. The stale-restore path and the no-own-snapshot fallback to an empty volume both have tests. The docstring records the rejected alternative, a `SnapshotType` whitelist, with the reason, and the newly-skipped pending snapshot gets a log line naming both ids.

**Where the risk is**

The findings concentrate in two regions, and both are the same defect class as the one being fixed.

The larger cluster is the clone path. Six findings share `cluster_id` c-001, arriving from independent angles and converging on `process_clone_disk_action` at index.py:10270-10313 and its consumer, Step 2b at index.py:7464-7501. That is the third selection site in the file, with the identical `max(StartTime)` pick over the same tag-filtered set, and it is untouched. Unlike the restore path, the pick is persisted to DynamoDB as `clone_source_snapshot`, so a bad choice is durable rather than re-evaluated. The underlying shape: the new rule is enforced by hand at each call site instead of at the query or selection boundary, which makes it a convention rather than an invariant.

The second cluster, c-002, is the query reordering itself. A snapshot that is pending when the completed query runs and completed by the time the pending query runs appears in neither result set, and the restore silently uses the previous session's snapshot. That is the same silent-stale-restore outcome the change exists to eliminate, reached by timing rather than by a scanner copy.

Two findings share c-003 and are one edit: the helper's general name and docstring against a body that detects one specific tool, and the fact that a foreign exclusion emits nothing, in a block where every other branch logs. That silence compounds both clusters above, because each of those failure modes produces a plausible-looking restore with nothing in CloudWatch naming the cause. Two more findings are documentation accuracy: the docstring's account of which snapshots lack `SnapshotType`, and an out-of-scope note in the description citing a call chain that does not exist in the repository.

**Before merge, and after**

Fix before merge: the reordering window in `create_disk_from_snapshot_or_empty` — moving the pending query back above the completed one is a block move, and it removes a newly-reachable stale restore. Also the clone-source filter at index.py:10270-10273, which is a one-line change with a durable consequence.

Reasonable as follow-up: guarding the read side in Step 2b, which is a design call about how to treat already-written references; renaming and logging the exclusion in `is_foreign_snapshot`; and the two documentation corrections, on tagless snapshots created by the clone path and on the real caller of `shared/snapshot_utils.py::get_latest_snapshot`.

</details>

gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py
```diff
@@ -7420,15 +7418,49 @@ def create_disk_from_snapshot_or_empty(user_id: str, availability_zone: str, dis
...
                 active_snapshots.append(snap)
 
         latest_snapshot = max(active_snapshots, key=lambda s: s['StartTime']) if active_snapshots else None
```
izaitsevfb, line 7461: please consider if you want to address this race.

---

<details>
<summary>🟡 Swapping the two snapshot queries opens a window in which a snapshot that completes between them is invisible to both. (ai-generated section)</summary>

The completed-snapshot query now runs first and the pending query second. A snapshot that is still pending when the first query runs, and completed by the time the second one runs, is returned by neither — so the reservation neither waits for it nor restores from it, and provisions the disk from an older snapshot instead. That is the silent stale restore this change exists to prevent, and it is new: in the previous order the same timing produced an unnecessary wait, which fails safe. The window is only the gap between the two API calls, so it is rare rather than routine, but it is also cheap to close — one `describe_snapshots` asking for both states and splitting the result on `State` removes it entirely.

<sub>Reviewed by codex gpt-5.6-sol @ xhigh + claude-fable-5, concurrent at xhigh effort, against 4a3e88242.</sub>

</details>

gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py
```diff
@@ -111,6 +111,28 @@ class SnapshotPendingError(RuntimeError):
...
+    Matched on copy markers rather than a required gpu-dev tag: ~2% of real
+    snapshots are hand-made recovery ones carrying no SnapshotType tag.
+    """
+    tags = {tag["Key"]: tag["Value"] for tag in snapshot.get("Tags", [])}
+    if any(key.lower().startswith("wiz") for key in tags):
+        return True
+    # A copy has no source volume in this account.
+    if snapshot.get("VolumeId") == "vol-ffffffff":
+        return True
+    return "auto-generated by Wiz" in (snapshot.get("Description") or "")
```
izaitsevfb: consider adding test coverage.

---

<details>
<summary>⚪ Two of the three copy markers have no test coverage, and the copy in the description is the kind those two catch. (ai-generated section)</summary>

`is_foreign_snapshot` matches three independent markers and returns on the first that hits. Every `_wiz_copy` fixture in the new test file carries a `wiz` tag key, so all five tests return from that first check and neither the `VolumeId` nor the description branch is ever evaluated; there is also no direct test of the predicate itself. Meanwhile the copy the description holds up as the production case is identified there by its `auto-generated by Wiz` description and by the copying role, with no `wiz` tag listed — so the branch most likely to fire in production is one of the two with no coverage, and a change in the scanner's tagging would be invisible to this suite. Two more fixtures, one carrying only the description and one carrying only `vol-ffffffff`, would close it.

<sub>Reviewed by codex gpt-5.6-sol @ xhigh + claude-fable-5, concurrent at xhigh effort, against 4a3e88242.</sub>

</details>

izaitsevfb (approved):
our ai overlords say approve, but please take a look at the nits

izaitsevfb:
https://github.com/pytorch/ci-infra/blob/4a3e8824287ee2844e3de3f4d4965d1bb0db3d54/gpu-dev/terraform-gpu-devservers/lambda/reservation_processor/index.py#L10269-L10273

perhaps makes sense to fix that third place as well?

---

<details>
<summary>🟡 The disk-clone path picks its source snapshot the same unfiltered way, so a cloned disk can be pinned to a copy that is about to disappear. (ai-generated section)</summary>

`process_clone_disk_action` selects the newest completed snapshot matching the `gpu-dev-user` and `disk_name` tags, excluding only soft-deleted ones, and writes that id into DynamoDB as `clone_source_snapshot`. Step 2b of the function this PR changes reads it back and assigns it to `latest_snapshot` unchecked, for as long as the cloned disk has no snapshot of its own. A copy inherits both tags and sorts newest, so it can win that selection exactly as it did in the two places fixed here — and a copy is short-lived: once the scanner deletes it, `describe_snapshots(SnapshotIds=[clone_source])` raises and Step 2b re-raises rather than falling back, so every reservation of that disk fails, and the disk can never take a snapshot of its own to escape the reference. Applying `is_foreign_snapshot` to the clone path's `active_snapshots` list closes it.

<sub>Reviewed by codex gpt-5.6-sol @ xhigh + claude-fable-5, concurrent at xhigh effort, against 4a3e88242.</sub>

</details>

huydhn:
Thanks both — the two majors were real and are fixed in 0610147. Detail per point:

## Fixed before merge

**🟠 Clone-source selection (`process_clone_disk_action`)** — @jeanschmidt and @izaitsevfb both landed on this, and the durability argument is what makes it worse than the two sites I did fix: the pick is persisted as `clone_source_snapshot`, so it isn't re-evaluated. Once the scanner deletes the copy, Step 2b's `describe_snapshots(SnapshotIds=[clone_source])` raises and re-raises rather than falling back, and the disk can never take a snapshot of its own to escape the reference. Now filtered.

I checked prod for damage already done: **25 cloned disks carry a `clone_source_snapshot`, resolving to 15 distinct snapshots — all legitimate gpu-dev shutdown/cancellation/live-clone snapshots, all `completed`, none deleted, none a copy.** So nothing is wedged today; the bug was latent. No remediation needed, just the filter.

**🟠 Query reordering window** — correct, and it's the same failure class the PR removes, so thank you for catching it. Rather than move the block back and lose the staleness comparison, pending is queried **first** again and the comparison still works: a snapshot pending during the first query and completed by the second now appears in *both* sets, compares equal under `latest_pending['StartTime'] <= latest_snapshot['StartTime']`, and is used as the restore source. Pinned by `test_pending_query_runs_before_completed_query`, which asserts the call order so the block can't drift back.

## Also done

**🟡 Naming** — fair; the name promised more than the body delivered. Renamed to `is_third_party_snapshot_copy`, and the markers are now named constants (`_SNAPSHOT_COPY_TAG_PREFIXES`, `_SNAPSHOT_COPY_DESCRIPTIONS`) so adding a copier is a one-line change. The docstring now says plainly that this detects known copiers and that an unlisted copier still reads as ours.

**🟡 Silent exclusion** — right that it was the only branch in that decision that logged nothing. Exclusions now go through `filter_own_snapshots()`, which names the skipped snapshot ids and the context (`pending`, `completed`, `clone source for '<disk>'`).

**🟡 Docstring accuracy** — confirmed: `process_clone_disk_action` tags live-clone snapshots with only `gpu-dev-user`/`disk_name`/`Name`, no `SnapshotType`. Docstring corrected to cite the clone path alongside the hand-made recovery ones. 4 of the 15 clone sources above are exactly these tagless live-clone snapshots, which is also a concrete reason the `SnapshotType` allowlist would have been wrong.

**PR description error** — you were right that the out-of-scope note cited a call chain that doesn't exist here. `get_latest_user_snapshot` isn't in this repo at all (I ported that note from the osdc version without re-checking). `migrate_ebs_across_az` does exist at `index.py:877` and still has no callers. Description updated.

## Deferred, with reasoning

**Step 2b read-side guard.** Agreed this is the design call rather than a bug fix, so I'd rather not decide it in this PR. Filtering at write time stops new bad references, and none exist today. The open question is what a clone *should* do when its source is genuinely gone: falling back to an empty volume silently hands someone a blank disk they believe is a clone, which I'd argue is worse than the current hard failure. The real fix is probably a clear terminal error plus a way to re-point or drop the reference — worth its own PR and its own opinion. Happy to take it next if you'd like it sooner.

Also still outstanding from my original description: the unbounded `queued` state inherited from the non-blocking-retry change (a genuinely stuck snapshot now queues forever with no deadline, where it previously failed after ~45 min).

## Verification

```
1226 passed, 3 xfailed        (7 in test_foreign_snapshots.py, 2 new)
```

Both new tests confirmed failing first: `query order was ['completed', 'pending']` and `assert 'snap-wiz' == 'snap-ours'`.

Re-validated the renamed helper over all **6799** currently gpu-dev-tagged snapshots: **6799/6799 real snapshots kept, zero false exclusions.** One caveat on my earlier "16/16 copies excluded" number — the scanner has since deleted all of its copies, so there are none in the account right now to re-measure against. That half is covered by unit tests today; the 16/16 figure stands from when they existed.

One note for whoever merges: prod's `live` alias was pinned to a June 16 version until earlier today, so merged lambda changes weren't actually running. Worth confirming the alias advances after the apply for this one.

jeanschmidt (approved):

PR #1057, opened 2026-09-04
[node-compactor] Quarantine GPU nodes that black-hole every GPU pod

huydhn:
## What

One dead GPU silently takes an entire GPU node out of service. `nvidia-device-plugin`'s `alignedAlloc` builds the NVLink topology across **all** GPUs on the node before filtering to the ones the pod requested, so a single lost GPU fails `GetPreferredAllocation` for every GPU pod — including pods that would have landed on a healthy GPU. The node keeps reporting `Ready` and keeps advertising full allocatable GPUs, so it accepts work at full rate and destroys all of it.

Today on `meta-prod-aws-ue2`: `ip-10-4-164-212` rejected **554 pods over 1h46m** (first 14:49:22Z, last 16:35Z). Every rejection aborted on the *same* GPU pair:

```
300  devices (0, 1)
```

It never evaluated GPUs 2 or 3, and never looked at which GPU the pod asked for. The node was only removed at 16:42:23Z by emptiness consolidation — not by health.

Verbatim kubelet message:

```
Allocate failed due to device plugin GetPreferredAllocation rpc failed with err:
rpc error: code = Unknown desc = error getting list of preferred allocation devices:
unable to get device link information: error getting NVLink for devices (0, 1):
failed to get nvlink remote pci info: failed to get nvlink state: GPU is lost,
which is unexpected
```

## Changes

**Detection + quarantine** (`base/node-compactor/`)
- New `gpu_health.py`: matches pods killed by a device-plugin allocation failure, selects nodes over threshold.
- `discovery.py` records those failures per node in the pod pass it already runs every 7s — they were discarded one line later.
- `compactor.quarantine_gpu_black_holes()` taints before any capacity decision, since a black hole otherwise looks like usable capacity.
- Defaults: 3 failures / 5 min, capped at 20% of a fleet. Configurable via `clusters.yaml` → `node_compactor.gpu_quarantine_*`.
- **No RBAC change** — `nodes: patch` and `pods: list` were already granted.

**Kernel logs** (`modules/logging/pipelines/base.alloy`)
- The journal `keep` matched `__journal__systemd_unit` against a regex including `kernel`, but kernel messages carry no `_SYSTEMD_UNIT` — so every kernel line was dropped while the config read as though it collected them. Reordered so `unit` is derived first (kernel labelled from `__journal__transport`) and `keep` applies to the derived label. This is why we have no Xid record for today's incident.

## Design notes

- **Separate taint key** (`node-compactor.osdc.io/gpu-unhealthy`). Every untaint path is keyed on `cfg.taint_key`, so the quarantine can't be cleared by burst demand, cooldown, or `min_nodes`.
- **No job is killed.** The node drains as healthy in-flight pods finish; existing `WhenEmpty` reaps it.
- **Fleet cap** guards the worse failure: a bad AMI or driver rollout would otherwise cordon every GPU node at once. Karpenter's own node-repair path applies the same 20% guard.
- **Symptom, not hardware probe.** No NVML, no privileged DaemonSet, no per-node agent — and it fires on any device-plugin allocation fault, not only ones a probe anticipated. Cost: it's post-hoc, so ~3 pods die before it trips. Against 554, that's the trade.

## Rejected alternatives

- **Bumping the device plugin.** Doesn't fix it. `go-gpuallocator` is pinned at `v0.6.0` in v0.19.0, v0.19.3 *and* v0.20.0, and `NewDevices()` is byte-identical across all three; v0.20.0 only reworked the non-aligned path. No config knob either — `GetPreferredAllocationAvailable: true` is hardcoded.
- **Karpenter `NodeRepair` alone.** It never cordons: it waits the full 10-min toleration, then force-deletes with `terminationGracePeriod=0`, killing healthy in-flight jobs. At today's rate that's still ~60 lost jobs per episode. Worth adding *on top of* this later, not instead of it.

## Testing

`just lint` (13/14 — the `hadolint` failure on `runner-base/Dockerfile:36` pre-exists on a clean tree) and `just test` pass at 98.82% coverage.

31 new tests. The regression test uses the verbatim kubelet message above and asserts the taint lands on the third rejection while the other 19 g5 nodes stay untouched.

## Not verified

The Alloy fix is reasoned from systemd semantics plus the observed fact that `unit="kernel"` returns nothing cluster-wide while the other four units work. Confirming it is one query for `{unit="kernel"}` after the logging module rolls out.

Refs: https://github.com/meta-pytorch/pytorch-gha-infra/issues/1470


huydhn:
Superseded by the ghstack stack #1058 → #1061, which splits this into a minimal base commit plus three follow-ups.

PR #1063, opened 2026-09-10
Remove the algif and DirtyFrag CVE mitigation DaemonSets

huydhn:
Both were stopgaps for kernels we no longer run: CVE-2026-31431 needed 6.12.85+, CVE-2026-43284/43500 needed 6.12.83+, and every node is on 6.12.100+ or 6.18.x once #1062 has rolled. Also drops the startup-taint registry entries, the TODO markers on the four AMI pinnings, the two Grafana panels, and the docs references.

⚠️ Stacked on #1062 and must not merge until the base node groups have actually rotated — 24 of them are still on 6.12.73, where both mitigations are doing real work.


osdc/base/kubernetes/algif-mitigation.yaml
```diff
@@ -1,237 +0,0 @@
...
-# cluster any unprivileged workflow code can escalate to root on the node.
-# CVSS 7.8 / CISA KEV.
-#
-# AL2023 ships algif_aead as a loadable module (CONFIG_CRYPTO_USER_API_AEAD=m),
-# verified on kernel 6.12.79-101.147.amzn2023. The module is not loaded by
-# default — this DaemonSet writes /etc/modprobe.d/disable-algif.conf to
-# block on-demand loading, then defensively rmmod's it if anything has
-# already loaded it.
-#
-# Runs on EVERY node (base infra + Karpenter ARC + BuildKit + GPU + H100/B200).
```
jeanschmidt, line 14: Ok, this is a blocker, especially on the ai-sandbox, so please make sure that we're rebuilding those AMIs before shipping this change (confirmed the AMIs don't contain the kernel with the fix)

---

<details>
<summary>🔴 The deleted DaemonSets ran on every node, but the ai-sandbox fleet pins a tag-selected AMI the repo documents as not picking up AL2023 kernel CVE fixes (ai-generated section)</summary>

This line refers to the pre-change file. Both mitigation DaemonSets ran on literally every node (`tolerations: [{operator: Exists}]`, "NO nodeSelector — must run on every node in the cluster"). Removing them rests on the premise that node rotation has moved every node onto a patched kernel, which holds only for fleets whose EC2NodeClass tracks `alias: al2023@​latest` or the `amazon-eks-node-al2023-x86_64-nvidia-*` glob. The `ai-sandbox` fleet selects its AMI by tag `osdc.io/ami: ai-sandbox-gvisor`, whose kernel is frozen at whatever the SSM `recommended` parameter returned the last time someone ran `just build-agent-sandbox-ami`. osdc/modules/nodepools/scripts/python/generate_nodepools.py:284-287 states it outright: "`alias: al2023@​latest` picks up AL2023 CVE fixes automatically on node rotation; a custom AMI only moves when it is rebuilt." osdc/modules/nodepools-agent-sandbox/defs/ai-sandbox.yaml:11-12 repeats it: "It pins the AL2023 base, so unlike the al2023@​latest fleets it does not pick up CVE fixes on node rotation — rebuild to refresh." osdc/modules/nodepools-agent-sandbox/packer/ai-sandbox-gvisor.pkr.hcl:62-70 resolves `source_ami` from `/aws/service/eks/optimized-ami/.../recommended/image_id` at build time only. The `nodepools-agent-sandbox` module is enabled on two clusters (osdc/clusters.yaml:312 and :419). The stacked #​1062 bump this PR cites (`base_node_ami_version: "v20260903"`, osdc/clusters.yaml:30) drives the managed base node group only and does not touch Karpenter fleets. On any cluster where the tagged gVisor AMI predates the AL2023 build carrying the 6.12.85 / 6.12.83 fixes, ai-sandbox nodes lose the `algif_aead` and `esp4`/`esp6`/`rxrpc` blacklists while still running the vulnerable kernel — and after this PR there is nothing in the repo, the docs, or the dashboard that records that these nodes ever had a mitigation. I could not read that AMI's kernel version from the repository, so the exposure is asserted as unverified rather than confirmed. Narrow: before merging, rebuild the gVisor AMI (`just build-agent-sandbox-ami` for the clusters at osdc/clusters.yaml:312 and :419), confirm `uname -r` on an ai-sandbox node is &gt;= 6.12.85, and record that in the PR. Broader: add an assertion next to `test_ami_exists_in_region` (osdc/modules/nodepools-agent-sandbox/tests/smoke/test_nodepools_agent_sandbox.py:115) that fails when the tagged AMI's CreationDate predates a declared minimum, so a stale gVisor image can never silently outlive a kernel-CVE window again. I would do the rebuild now and add the smoke assertion.

<sub>Reviewed by claude-opus-5[1m] at max effort, against b2a65ce7d.</sub>

</details>

osdc/base/kubernetes/kustomization.yaml
```diff
@@ -12,8 +12,6 @@ resources:
   - nvidia-device-plugin.yaml
   - harbor-namespace.yaml
   - registry-mirror-config.yaml
-  - algif-mitigation.yaml
```
jeanschmidt, line 15: we need to include cleanup logic, otherwise the daemonset will continue to exists and can become a annoyance or a problem in the future with no code tracing it down

---

<details>
<summary>🔴 Two base manifests are dropped from the kustomization with no orphan-cleanup script, unlike the git-cache removal that shipped scripts/cleanup-git-cache.sh (ai-generated section)</summary>

This line number refers to the pre-change file. `algif-mitigation.yaml` and `dirtyfrag-mitigation.yaml` are deleted from the repo and dropped from `resources:`, but base k8s is deployed with a plain `kubectl apply -k` (osdc/justfile:394) which never prunes. Each manifest carried a DaemonSet, a ConfigMap, a ServiceAccount, a ClusterRole and a ClusterRoleBinding, so ten cluster objects are left running on every existing cluster with nothing in the repo that references or reaps them. This repository already has an established shape for exactly this situation and the change does not follow it. `osdc/scripts/cleanup-git-cache.sh` was added for the identical case; its header states the rationale at lines 4-7 ("Delete every in-cluster resource left orphaned by the removal of the git-cache module. The git-cache codebase ... was deleted in a code-only PR; this script handles the cluster-side cleanup that was deliberately deferred to operators") and lines 44-47 ("Why not kustomize delete: The git-cache manifests are gone from the repo, so kustomize can no longer produce the manifest set required to delete-by-ref"). It enumerates exactly the object kinds this change strands — DaemonSet, ConfigMap, ServiceAccount, ClusterRole, ClusterRoleBinding — at lines 25-34, plus node taints at line 38. `osdc/modules/buildkit/deploy.sh:83` states the same rule for its own resources: "Apply never deletes, so sweep by label." `osdc/modules/nodepools/deploy.sh:127-157` implements the same stale sweep. Nothing else reaps these: `just undeploy` selects on `osdc.io/module` (osdc/justfile:664-683) and neither deleted manifest carried that label — they used `app`/`app.kubernetes.io/name` only. Grep for `algif|dirtyfrag` over the worktree returns zero hits, so no cleanup path was added anywhere. After this merges and `just deploy-base` runs, both mitigation DaemonSets keep running on every node of every existing cluster, keep re-writing `/etc/modprobe.d/disable-algif.conf` and `/etc/modprobe.d/disable-dirtyfrag.conf` (blocking algif_aead, esp4, esp6, rxrpc) on every newly provisioned node, and keep holding cluster-wide `nodes get/patch` RBAC — so the change does not actually remove the mitigations it is titled after. The leftovers are invisible to the repo, so the next operator who finds `dirtyfrag-mitigation` in `kubectl get ds -n kube-system` has no manifest to trace it back to. Add `osdc/scripts/cleanup-cve-mitigations.sh` modelled directly on `scripts/cleanup-git-cache.sh` (same OSDC_CONFIRM dry-run gate, same enumerate/delete/verify structure), enumerating in kube-system: daemonset/configmap/serviceaccount named `algif-mitigation*` and `dirtyfrag-mitigation*`, the cluster-scoped `algif-mitigation-taint-remover` and `dirtyfrag-mitigation-taint-remover` ClusterRole/ClusterRoleBinding, and any node still carrying `node-init.osdc.io/algif-mitigation` or `node-init.osdc.io/dirtyfrag-mitigation`. Narrower alternative if you do not want a script: land the deletions but keep a documented one-liner in the PR/runbook (`kubectl delete ds,cm,sa -n kube-system -l app in (algif-mitigation,dirtyfrag-mitigation)` plus the two clusterrole/clusterrolebinding pairs). I would take the script — it matches the precedent, is idempotent, and covers the node-taint case the one-liner misses.

<sub>Reviewed by claude-opus-5[1m] at max effort, against b2a65ce7d.</sub>

</details>

osdc/.claude/skills/osdc-runners-nodepools/SKILL.md
```diff
@@ -148,8 +146,6 @@ Generated EC2NodeClass templates set `spec.metadataOptions.httpProtocolIPv6: ena
 GPU pools (H100, B200) optionally append a `text/x-shellscript` MIME part via the per-def `user_data_script` field — used today for one-shot containerd registry mirror config that must be in place before any image pull. Prefer DaemonSets where possible; reserve `user_data_script` for boot-critical setup only.
 
 **TEMPORARY mitigations** — two kernel-mod blacklist DaemonSets with identical shape and the same lifecycle:
```
jeanschmidt, line 148: We should cleanup documentation with this change, there might be more pointers

---

<details>
<summary>🟠 The "TEMPORARY mitigations" heading in osdc-runners-nodepools/SKILL.md survives with both of its bullets deleted, asserting two DaemonSets that no longer exist (ai-generated section)</summary>

The diff removes the two bullet items under this heading (the `base/kubernetes/algif-mitigation.yaml` and `base/kubernetes/dirtyfrag-mitigation.yaml` entries) but leaves the heading line itself. The result is a sentence that declares "two kernel-mod blacklist DaemonSets" exist and ends in a colon introducing an empty list, immediately followed by the `### NodePool Def File Schema` heading. Hunk `@@ -148,8 +146,6 @@` of `osdc/.claude/skills/osdc-runners-nodepools/SKILL.md` removes exactly two lines (old count 8 → new count 6) and keeps the heading as a context line. The manifests it describes are deleted by this same diff (`osdc/base/kubernetes/algif-mitigation.yaml` and `osdc/base/kubernetes/dirtyfrag-mitigation.yaml`, both `deleted file mode`), and `osdc/base/kubernetes/kustomization.yaml:14-15` no longer lists them. Grep for `algif|dirtyfrag` across the checkout returns no matches, so nothing the sentence describes is left in the repository. Anyone (or any coding assistant) reading this skill to learn what runs on OSDC nodes is told two kernel-CVE blacklist DaemonSets are deployed with a "TEMPORARY lifecycle", and finds nothing to look at. The same skill's other sections are the canonical reference for what is under `base/kubernetes/`, so the dangling assertion contradicts the list at line 31 that this diff already corrected. Delete line 148 (the `**TEMPORARY mitigations** — two kernel-mod blacklist DaemonSets with identical shape and the same lifecycle:` paragraph) along with one of the surrounding blank lines, so the section ends at the `user_data_script` paragraph and `### NodePool Def File Schema` follows directly.

<sub>Reviewed by claude-opus-5[1m] at max effort, against b2a65ce7d.</sub>

</details>

osdc/docs/architecture.md
```diff
@@ -13,7 +13,7 @@ Every cluster gets the same base infrastructure:
...
-- **Base k8s resources** — `osdc-system` namespace (used by deploy audit ConfigMaps), gp3 StorageClass, NVIDIA device plugin, node performance tuning DaemonSet, registry mirror config, Harbor namespace, image-cache-janitor (prunes stale image content from node disks), NodeLocal DNSCache (per-node CoreDNS DaemonSet binding `fd00::10`, intercepts pod DNS via iptables-mode NOTRACK), ENIConfigs (one per AZ; currently inert pending VPC CNI Custom Networking enablement), and two transient CVE-mitigation DaemonSets (`algif-mitigation` for CVE-2026-31431, `dirtyfrag-mitigation` for CVE-2026-43284) that will be removed once a kernel-patched AMI is in use.
+- **Base k8s resources** — `osdc-system` namespace (used by deploy audit ConfigMaps), gp3 StorageClass, NVIDIA device plugin, node performance tuning DaemonSet, registry mirror config, Harbor namespace, image-cache-janitor (prunes stale image content from node disks), NodeLocal DNSCache (per-node CoreDNS DaemonSet binding `fd00::10`, intercepts pod DNS via iptables-mode NOTRACK), and ENIConfigs (one per AZ; currently inert pending VPC CNI Custom Networking enablement).
```
jeanschmidt, line 16: more docs fixes, that eventually will become a landmine for agents

---

<details>
<summary>⚪ osdc/docs/architecture.md:93 still lists "CVE-mitigation DaemonSets" among what `kubectl apply -k base/kubernetes/` deploys (ai-generated section)</summary>

The change updates the "Base k8s resources" bullet at line 16 of this file to drop the two CVE-mitigation DaemonSets, but the deployment-flow diagram lower in the same file still annotates the `kubectl apply -k base/kubernetes/` step with "CVE-mitigation DaemonSets". That line is untouched by the diff and now names a class of resource the kustomization no longer contains. I am anchoring on line 16 because line 93 falls outside every hunk in this file. `osdc/docs/architecture.md:93` reads `│ ├── kubectl apply -k base/kubernetes/ ← StorageClass, NVIDIA, CVE-mitigation DaemonSets, etc.`. The diff removes `algif-mitigation.yaml` and `dirtyfrag-mitigation.yaml` from the `resources:` list in `osdc/base/kubernetes/kustomization.yaml` (lines 14-15 of the base file) and deletes both manifests. A repo-wide grep for `mitigation` (case-insensitive) matches only this line and an unrelated "flakiness mitigation" note in `osdc/compactor_tests_change.md:87`. A reader following the deployment-flow diagram to find out what `deploy-base` applies goes looking for CVE-mitigation manifests under `base/kubernetes/` and finds none; the diagram is the doc's canonical summary of the base apply step. Change line 93 to drop the stale item, e.g. `│ ├── kubectl apply -k base/kubernetes/ ← StorageClass, NVIDIA, registry-mirror config, etc.`.

<sub>Reviewed by claude-opus-5[1m] at max effort, against b2a65ce7d.</sub>

</details>

jeanschmidt (approved):
Rebuild the AMIs, add a cleanup logic, fix the docs, but stamping to unblock

## Review summary

🔴 2 blocker · 🟠 1 major · ⚪ 1 nit

- 🔴 The deleted DaemonSets ran on every node, but the ai-sandbox fleet pins a tag-selected AMI the repo documents as not picking up AL2023 kernel CVE fixes — `osdc/base/kubernetes/algif-mitigation.yaml:14 (LEFT)`
- 🔴 Two base manifests are dropped from the kustomization with no orphan-cleanup script, unlike the git-cache removal that shipped scripts/cleanup-git-cache.sh — `osdc/base/kubernetes/kustomization.yaml:15 (LEFT)`
- 🟠 The "TEMPORARY mitigations" heading in osdc-runners-nodepools/SKILL.md survives with both of its bullets deleted, asserting two DaemonSets that no longer exist — `osdc/.claude/skills/osdc-runners-nodepools/SKILL.md:148`
- ⚪ osdc/docs/architecture.md:93 still lists "CVE-mitigation DaemonSets" among what `kubectl apply -k base/kubernetes/` deploys — `osdc/docs/architecture.md:16`

<details>

**What this changes**

The change retires two kernel-CVE mitigation DaemonSets. It deletes `osdc/base/kubernetes/algif-mitigation.yaml` and `dirtyfrag-mitigation.yaml`, drops both from the base kustomization, removes their two entries from `STARTUP_TAINTS` in `generate_nodepools.py` along with the tests asserting them, removes the same two keys from node-compactor's ignore list in `base/node-compactor/scripts/python/taints.py`, deletes Grafana panels 141 and 142 plus their grid items, and strips the `TODO(CVE-...)` markers from `clusters.yaml`, `generate_buildkit.py`, `generate_nodepools.py` and the pypi-cache EC2NodeClass template. Net is 7 added lines against 820 removed. The code matches the description; the description additionally states a merge gate that no file enforces.

**What it gets right**

The registry-and-DaemonSet lockstep rule is honoured: both `STARTUP_TAINTS` entries leave in the same commit as the manifests they belong to, which is exactly the deadlock the deleted comment in `generate_nodepools.py` warned about. The rule itself is generalised in `docs/node-warmup-and-scheduling-gates.md:92` rather than deleted with its example, so the invariant outlives the case that motivated it. The node-compactor ignore list and the registry unit tests are updated in the same pass instead of being left to fail later.

**Where the risk is**

One line carries most of it. Several reviewers independently landed on `osdc/base/kubernetes/kustomization.yaml`, and those findings share a cluster: `deploy-base` runs a bare `kubectl apply -k` (`osdc/justfile:394`) with no prune, so dropping entries from `resources:` unmanages ten live objects — two privileged DaemonSets, two script ConfigMaps, two ServiceAccounts, two ClusterRole/ClusterRoleBinding pairs granting cluster-wide `nodes: get,patch` — rather than deleting them. The same cluster holds the leftover `/etc/modprobe.d/disable-*.conf` files on nodes that already ran the mitigations, and the capacity-model drift, since `daemonset_overhead.py` discovers DaemonSets by scanning `base/kubernetes/` and stops counting requests those pods still hold. One cleanup script in the shape of the existing `osdc/scripts/cleanup-git-cache.sh` settles that whole group.

A second cluster attacks the premise rather than the code: "every node is on 6.12.100+" holds for fleets tracking `alias: al2023@​latest`, but the ai-sandbox gVisor fleet selects its AMI by tag and moves only on rebuild. Three reviewers converged there. It compounds with the unenforced merge gate, since the `clusters.yaml` hunk removes the only in-tree record of the rotation precondition. The recurring class across the change is the same one each time: complete in the repository, incomplete in live clusters, on node filesystems, on the dashboard, and for one non-rotating fleet.

**Before merge, and after**

Blocking, in order: cluster-side cleanup for the orphaned objects (the kustomization findings), the ai-sandbox AMI kernel check, and recorded confirmation that the base node groups have actually rotated. Two low-severity items are one-line edits and belong here rather than in follow-up — the `TEMPORARY mitigations` heading left with no bullets in `osdc/.claude/skills/osdc-runners-nodepools/SKILL.md`, and the stale deploy-flow line at `osdc/docs/architecture.md:93`.

Reasonable afterwards: re-tiling the grid hole at y=12 in `grafana/osdc_services_status.json`; the panel-removal timing, which is a split-the-PR call; and removal of the on-node modprobe.d files, which needs a design decision between a one-shot pass and forced rotation.

</details>

osdc/base/kubernetes/algif-mitigation.yaml
huydhn: Yup, I need to rebuild those AMI in the context of T285302716 too, so this is needed

jeanschmidt: dismiss this one, I re-checked the ai-sandbox and the current AMI have those fixes

PR #1064, opened 2026-09-10
docs: where the GPU driver comes from, and moving ahead of AWS

huydhn:
Parks what we found checking CUDA 13.4 support on the A10G / L4 runners: every GPU fleet takes its driver from the stock EKS NVIDIA AMI via a name glob, so we have no lever on the version — 580.178.04 (R580) today, while CUDA 13.4 wants R615 and AWS has never shipped past R580.

Docs-only, no decision made. Writes up the three routes (custom packer AMI, `cuda-compat-13-4` in the image, GPU Operator) with what each costs, so we can pick this up later.

### DIFF
--- a/osdc/docs/nvidia-driver-ahead-of-aws.md
+++ b/osdc/docs/nvidia-driver-ahead-of-aws.md
@@ -1,0 +1,163 @@
+# Moving the NVIDIA driver ahead of the AWS AMI schedule
+
+Status: **investigation, no decision made.** Captured 2026-09-10 so we can pick it
+up later. Trigger was a question about CUDA 13.4 support on the A10G / L4 runners.
+
+## The gap
+
+Every GPU fleet takes its driver from the stock EKS NVIDIA AMI. There is no
+driver pin anywhere in `osdc/` — `generate_nodepools.py` selects GPU AMIs with a
+name glob and the driver is whatever that image happens to carry:
+
+```python
+ami_selector_block = """  amiSelectorTerms:
+    - name: "amazon-eks-node-al2023-x86_64-nvidia-*\""""
+```
+
+This covers **all** GPU fleets — g5 (A10G), g6 (L4), g4dn (T4), p4d, p5 (H100),
+p6-b200. `ai-sandbox` is the only def in the repo that overrides the AMI, and it
+is CPU-only.
+
+Consequence: the driver moves when AWS decides it moves, and we have no lever.
+
+## Current state (verified 2026-09-10)
+
+| | |
+|---|---|
+| Driver | `580.178.04` (`kmod-nvidia-latest-dkms-580.178.04-1.amzn2023`) |
+| Driver branch | **R580** |
+| nvidia-container-toolkit | `1.20.0` |
+| AMI | `amazon-eks-node-al2023-x86_64-nvidia-{1.35,1.36}-v20260903` |
+
+AWS driver history across recent EKS AMI releases:
+
+| AMI release | Driver |
+|---|---|
+| v20260903 | 580.178.04 |
+| v20260827 | 580.178.04 |
+| v20260818 | 580.178.04 |
+| v20260810 | 580.159.03 |
+| v20260801 | 580.159.03 |
+| v20260728 | 580.159.03 |
+
+AWS has stayed on R580 throughout. They have not shipped R590, R595, R610 or
+R615 in any EKS AMI to date.
+
+## Why this matters: CUDA 13.x is branch-gated
+
+The CUDA 13.4 release notes map toolkits to *driver branches*, not to point
+releases the way 12.9 and earlier did:
+
+| CUDA | Driver branch |
+|---|---|
+| 13.4 | **R615** |
+| 13.3 | R610 |
+| 13.2 | R595 |
+| 13.1 | R590 |
+| 13.0 | R580 |
+
+Two statements from those notes set the boundary:
+
+> Existing CUDA 13.x applications run on drivers >=580 under CUDA minor version
+> compatibility.
+
+> CUDA 13.4 new features and newly enabled platforms require an R615 or later
+> driver that supports them.
+
+So R580 clears the `>= 580` minor-compat floor — 13.4-built code that stays on
+13.0-era driver APIs runs today. Anything using actual 13.4 features does not,
+and R615 is four branches beyond what AWS ships.
+
+Note CUDA 13.4 is currently a **Developer Preview**, not GA, and NVIDIA labels it
+unfit for production. Nothing is urgent yet. Two packaging changes land with it
+that affect any build we do ourselves: on Linux the driver stops shipping with
+the toolkit as of 13.4, and R615 packages no longer include the proprietary
+kernel modules.
+
+## Options
+
+### 1. Custom GPU AMI (leading candidate)
+
+Build on the stock EKS NVIDIA AMI and upgrade the driver, then select it by tag.
+The machinery already exists and is proven: `modules/nodepools-agent-sandbox/packer`
+does exactly this shape for gVisor, and the nodepool generator already supports
+`ami_selector_tags` per def.
+
+What we take on:
+
+- **CVE fixes stop being automatic.** This is the real cost. The glob currently
+  picks up kernel fixes on node rotation with no action — see the base node
+  groups in `clusters.yaml`, which were the *only* nodes left unpatched for
+  CVE-2026-64561 precisely because they were pinned. A custom GPU AMI puts every
+  GPU node in that same category, and GPU nodes are the bulk of the fleet.
+- **We own EKS version skew.** The glob has no Kubernetes version in it, which is
+  already causing drift worth fixing independently (see below). A tag selector
+  has the same blind spot — the `ai-sandbox` def carries a TODO about exactly
+  this.
+- Per-region builds, since AMIs are regional.
+
+Worth scoping a rebuild-on-a-schedule job if we go this way, so the image does
+not silently rot.
+
+### 2. `cuda-compat-13-4` in the CI container image
+
+Ship the forward-compatibility package in the image rather than touching nodes at
+all. Confirmed viable for our hardware: it requires a base driver `>= r580`
+(we have 580.178.04) and is supported on NVIDIA Data Center GPUs, which covers
+A10G, L4, T4, A100, H100 and B200.
+
+It installs driver-615 libraries under `/usr/local/cuda-13.4/compat/` and only
+provides the libraries — the consumer sets `LD_LIBRARY_PATH`. CUDA/OpenGL and
+CUDA/Vulkan interop are not supported through it.
+
+Cheapest option by a wide margin and it changes no infrastructure, but it is
+per-image rather than fleet-wide, so it only helps workloads we control.
+
+### 3. NVIDIA GPU Operator driver containers
+
+Fleet-wide and decoupled from the AMI, but it has to displace the driver already
+baked into the EKS NVIDIA AMI, which is the awkward part. Heaviest of the three.
+
+## Open questions
+
+- Do we actually need R615 features, or is minor-version compatibility enough?
+  This decides whether any of this is worth doing. Nothing has demanded 13.4 yet.
+- If we build a custom AMI, what keeps it current on kernel CVEs?
+- Does AWS have a public position on when EKS AMIs move off R580?
+
+## Unrelated issue found while investigating
+
+The GPU AMI glob has no Kubernetes version in it, so it matches the newest NVIDIA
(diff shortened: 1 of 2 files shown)


osdc/docs/nvidia-driver-ahead-of-aws.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Do we actually need R615 features, or is minor-version compatibility enough?
+  This decides whether any of this is worth doing. Nothing has demanded 13.4 yet.
+- If we build a custom AMI, what keeps it current on kernel CVEs?
+- Does AWS have a public position on when EKS AMIs move off R580?
+
+## Unrelated issue found while investigating
+
+The GPU AMI glob has no Kubernetes version in it, so it matches the newest NVIDIA
+AMI of *any* minor. In us-east-2 and us-west-1 that is now the 1.36 image, while
```
jeanschmidt, line 131: nit, should be an easy fix

---

<details>
<summary>⚪ The skew section names us-east-2 and us-west-1 as affected, but its evidence table has no us-west-1 row and instead shows unaffected us-east-1 (ai-generated section)</summary>

The prose at lines 130-132 asserts two affected regions, us-east-2 and us-west-1. The table immediately below at lines 134-137 lists `meta-prod-aws-ue1` (us-east-1, GPU kubelet v1.35.7 — not affected) and `meta-prod-aws-ue2` (us-east-2, v1.36.3 — affected). us-west-1 is claimed as affected but has no row, and the one region the table does contradict-check is a region the prose never names. `osdc/clusters.yaml:483-485` maps `meta-prod-aws-ue1` to `region: us-east-1`, `osdc/clusters.yaml:525-527` maps `meta-prod-aws-ue2` to `us-east-2`, and `osdc/clusters.yaml:434-436` maps `meta-prod-aws-uw1` to `us-west-1` — so the table's two rows are us-east-1 and us-east-2, not the us-east-2/us-west-1 pair the prose claims. uw1 does run GPU fleets subject to the same glob (`osdc/docs/h100-fabric-handles-imex-channels.md:222` records a verification on `meta-prod-aws-uw1`, `p5.48xlarge`). Whoever picks this up to fix the glob or to drain the skewed nodes cannot tell from the doc whether us-west-1 is affected, and the table invites them to conclude only ue2 needs attention — leaving a second production cluster running an unsupported kubelet/API-server pair. Add a `meta-prod-aws-uw1` row with its measured control-plane and kubelet versions so the table covers every region the prose names; if uw1 was not actually measured, drop it from the sentence at line 131 and say the check was run on ue1 and ue2 only.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 950231398.</sub>

</details>

osdc/docs/nvidia-driver-ahead-of-aws.md
```diff
@@ -0,0 +1,163 @@
...
+A kubelet newer than the API server is outside the supported skew. The CPU path
+does not have this problem because `alias: al2023@latest` tracks the cluster's
+Kubernetes version. Fixing the glob to include `eks_version` is worth doing
+regardless of what we decide about drivers.
+
+## Verifying current state
+
+```bash
+# Driver the fleet is actually running
+kubectl --context <cluster> get nodes -l node-fleet=g5 \
```
jeanschmidt, line 148: nit, should be an easy fix

---

<details>
<summary>⚪ The verification recipe uses `kubectl --context` and a typed `&lt;region&gt;`, not the `just kubeconfig` / `cluster-config.py` pattern its sibling runbooks use (ai-generated section)</summary>

The "Verifying current state" block reaches a cluster with `kubectl --context &lt;cluster&gt;` and passes `--region &lt;region&gt;` to the AWS CLI by hand. Sibling runbooks in the same directory establish the opposite shape: run `just kubeconfig "$CLUSTER"` first, then plain `kubectl`, and derive the region from `scripts/cluster-config.py` rather than typing it. The context this block names does not exist in a kubeconfig until `just kubeconfig` has created it. `osdc/docs/s3-gateway-endpoint.md:96` derives the region with `REGION=$(uv run scripts/cluster-config.py "$CLUSTER" region)` and :129 runs `just kubeconfig "$CLUSTER"` before an unqualified `kubectl run` at :134. `osdc/docs/cluster-recreation.md` does the same at :67, :83, :182, :235 and :261 (`just kubeconfig ${CLUSTER}`) and derives config at :102-103 and :223-224 (`CNAME=$(uv run scripts/cluster-config.py ${CLUSTER} cluster_name)` / `REGION=$(uv run scripts/cluster-config.py ${CLUSTER} region)`). `osdc/justfile:9-13` states the rule outright: anything touching a live cluster "MUST call `just kubeconfig &lt;cluster&gt;` or the equivalent `aws eks update-kubeconfig`". That recipe at `osdc/justfile:172-178` is what creates the named context, via `--alias "$CNAME"`. A reader copy-pasting the block gets `error: no context exists with the name` unless they happened to run `just kubeconfig` in an earlier session, and then has to look the region up by hand for the `describe-instances` call on line 150 — the two steps every other runbook in `osdc/docs/` removes. Open the block with the sibling preamble and drop `--context`: ```bash CLUSTER=meta-prod-aws-ue2 REGION=$(uv run scripts/cluster-config.py "$CLUSTER" region) just kubeconfig "$CLUSTER" kubectl get nodes -l node-fleet=g5 \ -o jsonpath='{.items[0].spec.providerID}' | sed 's|.*/||' aws ec2 describe-instances --region "$REGION" --instance-ids &lt;id&gt; \ --query 'Reservations[].Instances[].ImageId' --output text ``` This matches `osdc/docs/s3-gateway-endpoint.md:94-109` exactly.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 950231398.</sub>

</details>

jeanschmidt (approved):
docs for agents is aways a great thing! S2

## Review summary

⚪ 2 nit

- ⚪ The skew section names us-east-2 and us-west-1 as affected, but its evidence table has no us-west-1 row and instead shows unaffected us-east-1 — `osdc/docs/nvidia-driver-ahead-of-aws.md:131`
- ⚪ The verification recipe uses `kubectl --context` and a typed `&lt;region&gt;`, not the `just kubeconfig` / `cluster-config.py` pattern its sibling runbooks use — `osdc/docs/nvidia-driver-ahead-of-aws.md:148`

<details>

**What this changes**

The change adds one document, `osdc/docs/nvidia-driver-ahead-of-aws.md` (163 lines), and one row in the reference table in `osdc/CLAUDE.md`. Nothing else is touched: no Terraform, no generator, no config. The document records that every GPU fleet takes its driver from the stock EKS NVIDIA AMI selected by a version-less name glob, pins the current state at 580.178.04 / R580, explains why CUDA 13.4 wants R615, lays out three routes with their costs, and then appends two things that are not about drivers — a live kubelet/control-plane version skew found during the investigation, and a shell recipe for re-verifying the driver. The code and the description agree.

**What it gets right**

The load-bearing technical claims are checkable against the tree and they hold. The Python snippet quoted at lines 13-14 matches `generate_nodepools.py:267-268` verbatim; the CPU/GPU asymmetry the doc rests on is real (`alias: al2023@​latest` at line 288 versus the glob at 267); and `ai-sandbox` really is the only def carrying `ami_selector_tags`, and really is `gpu: false`. The cost named against the leading option mirrors the trade-off comment already sitting at `generate_nodepools.py:312-315`, so the doc argues from the repository's own stated position rather than around it. The status line says investigation, no decision, and the option-1 section leads with the security downside instead of burying it.

**Where the risk is**

Everything lands in two paragraphs and one code block of a single new file, and the reviewers converged hard. Three separate readings hit the CVE identifier at line 91 (`c-001`): `CVE-2026-64561` occurs exactly once in the whole tree, on that line, with no advisory URL, while the pin the sentence points at — `clusters.yaml:34` — is annotated with `CVE-2026-31431` and `CVE-2026-43284`, both linked, following a convention `node-warmup-and-scheduling-gates.md:152` states outright. One edit settles all three. Three more hit the skew table (`c-003`): line 131 names us-east-2 and us-west-1 as affected, and the table below it has rows for us-east-1 and us-east-2. One row, or one deleted clause, settles all three. Two hit tracking (`c-002`): the skew is filed under a heading that calls it unrelated, in a doc named after drivers, with no marker at `generate_nodepools.py:267` where two CVE TODOs three lines above prove readers already look. Two hit the verification block (`c-004`): `kubectl --context` deviates from the `just kubeconfig` preamble every other runbook in `osdc/docs/` uses, and the `describe-instances` query yields an `ami-…` ID while the next comment says to look up the name.

The recurring class is one thing: the doc's checkable cross-references do not resolve. That compounds, because both of them sit under the strongest arguments the doc makes. Two questions cannot be closed from the repository at all — whether that CVE identifier exists externally, and whether the skew is still live in production.

**Before merge, and after**

Before: the CVE identifier at line 91, and the missing us-west-1 row or the narrowed sentence at line 131. Both are one-line edits under the doc's two central claims, which is exactly why they are cheap to fix and expensive to leave. The missing `describe-images` step in the verification block is a third one-liner.

After: getting the skew tracked where the glob is generated, since that touches code this change deliberately leaves alone and needs a call on whether the `eks_version` fix rides along. Aligning the verification block with the `just kubeconfig` / `cluster-config.py` shape is straightforward and not urgent.

</details>

huydhn:
Thanks @jeanschmidt — both nits fixed, plus the CVE link and the `describe-images` step from the extended analysis. Also refreshed for CUDA 13.4 going GA.

**The skew table (nit 1)** — you were right, and the prose was the accurate half. I re-measured live on 2026-09-14 rather than adding a row from memory:

| Cluster | Region | GPU kubelet | CPU kubelet | GPU nodes affected |
|---|---|---|---|---|
| meta-prod-aws-ue1 | us-east-1 | v1.35.7 | v1.35.7 | 0 of 287 |
| meta-prod-aws-ue2 | us-east-2 | **v1.36.3** | v1.35.7 | **560** |
| meta-prod-aws-uw1 | us-west-1 | **v1.36.3** | v1.35.7 | **2** (both p5/H100) |

All three control planes are v1.35.6. So uw1 is affected as the prose claimed, and ue2 is worse than the old table implied — 560 nodes, not one row. That's 562 production GPU nodes outside supported skew right now.

Digging into why ue1 escapes turned up something worth having in the doc: it isn't structural, it's a one-second margin. AWS registers every minor of an AMI release within a few seconds, and the glob takes whichever is newest:

| Region | Newest match | Registered | Runner-up |
|---|---|---|---|
| us-east-1 | `…nvidia-1.35-v20260903` | 22:43:17 | `…-1.36-` at 22:43:16 |
| us-east-2 | `…nvidia-1.36-v20260903` | 22:43:26 | `…-1.35-` at 22:43:25 |
| us-west-1 | `…nvidia-1.36-v20260903` | 22:41:55 | `…-1.35-` at 22:41:53 |

us-east-1 could flip on any future release. That also covers `lf-prod-aws-ue2` by region without needing to measure it — I have no access to the lf account, and the doc says so.

**The verification block (nit 2)** — rewritten to the `just kubeconfig` / `cluster-config.py` shape, `--context` dropped, and it now ends at an AMI *name* via the missing `describe-images` call instead of leaving you holding an `ami-…` ID. Added a second block for re-checking the skew itself.

**CVE-2026-64561 (c-001)** — linked to its ALAS advisory and named the fix (#1062), matching the convention the `clusters.yaml` annotations follow.

**Tracking the skew at `generate_nodepools.py:267` (c-002)** — not done, and I'd like your call. It needs a decision on whether the `eks_version` fix rides along, and this PR is deliberately docs-only. What the doc does now is state that explicitly, and say that fixing the glob is the actual fix while draining the skewed nodes only helps until the next scale-up. Happy to do either here or in a follow-up — given 562 nodes are affected, a follow-up that actually fixes the glob seems better than a TODO comment.

**Two other changes** since you looked:

CUDA 13.4 is **GA** now — 13.4.1 shipped this month, so the "Developer Preview, unfit for production" framing is gone. Conclusion is unchanged: minor version compatibility still means R580 runs 13.4 binaries, so the question remains whether we need 13.4 *features*, and nothing has asked for them.

I also built option 1 to see what it costs (#1068, draft, parked). Two findings folded back in here because they change the option's price:

- `nvidia_driver_major_version=615` **cannot work** upstream. The build resolves `min(kmod-nvidia-open-dkms, AWS GRID runfile)` and AWS's runfile bucket stops at `595.91.07`. Swapping the driver on the finished AMI is the only route.
- R615 dropping the proprietary kmod is a **boot failure**, not a packaging note. `nvidia-kmod-load.sh` probes the version with `rpmquery kmod-nvidia-latest-dkms`, absent at 615, so selection falls through to a flavor that doesn't exist and the node loads no driver at all. Same shape as awslabs/amazon-eks-ami#2768. It also rules out g4dn/g5/g5g, which upstream hardcodes to the proprietary module for the GSP workaround.

PR #1066, opened 2026-09-11
harbor-cache-recovery: purge the failing artifact, not the whole repository

jeanschmidt:
**Impact:** OSDC - harbor-cache-recovery 
**Risk:** medium

## What
Reworks how the recovery job decides what to delete: it now matches verbatim containerd content-corruption strings, holds an explicit never-purge deny list, and issues an artifact-scoped `DELETE .../artifacts/<tag|digest>` instead of a repository-scoped delete. The script is split into `pull_failures.py` (scan + classify), `harbor_client.py` (Harbor API), and `harbor_cache_recovery.py` (orchestration), with tests split to match.

## Why
The job was doing damage and not fixing much:

- **Detection was far too broad.** `failed to copy` and `failed precondition` are generic containerd wrappers — a TLS fault, a full disk, or any node-local error came out as "cache corruption". `unexpected content digest` isn't a string containerd ever emits, so it never matched anything.
- **Every match deleted the entire repository**, evicting every good tag in it for every other consumer, to fix one bad artifact. Combined with the above, a node with a full disk could wipe the cached repo cluster-wide.
- **The dominant real failure, `unexpected media type text/html`, isn't fixable by purging.** Harbor rejects unregistered media types before caching, so there is nothing cached to remove and the purge only did harm. It's now explicitly denied and logged for investigation (it points at a Harbor front-door routing problem — separate open issue).
- **Layer-blob faults were purgeable but unfixable.** Deleting an artifact drops DB rows; the blob survives until GC and Harbor re-serves it, so those would have looped forever. They're excluded by design.
- **The critical alert paged on permanent states.** 404-on-digest and 412 (pinned as a child of a multi-arch index) now have their own outcomes; only genuine request failures set a non-zero exit.

Also: references are validated against the OCI tag/digest grammar and the URL is fully percent-encoded, because `..` in a reference or a `?` in a repo path would have silently widened the delete back to repository scope. Tag beats digest when an image pins both — Harbor recomputes proxied multi-arch index digests, so the upstream digest usually addresses nothing.

# Notes
- Purges are capped at 200 per run; the remainder rolls to the next tick rather than risking `DeadlineExceeded` against the 300s deadline.
- Alert descriptions and the Harbor skill/docs are updated for a fact that cost debugging time: `MANIFEST_UNKNOWN` never reaches a kubelet message — containerd renders registry 404s as `not found`. Grep for that, not the uppercase code.
- Worth reviewing the two indicator lists closely; they are the whole behaviour of this job, and each entry is annotated with the containerd source line it comes from.

### DIFF
--- a/osdc/.claude/skills/osdc-harbor/SKILL.md
+++ b/osdc/.claude/skills/osdc-harbor/SKILL.md
@@ -116,7 +116,7 @@ Three PDBs cover the multi-replica components only: `harbor-core`, `harbor-regis
 | `base/kubernetes/harbor/pdb.yaml.tpl` | PodDisruptionBudget template for core/registry/nginx (sed-substituted by `_deploy-harbor`, no native chart PDB support) |
 | `modules/eks/terraform/modules/harbor/main.tf` | S3 bucket + IAM role (IRSA) + IAM user (static-keys workaround) |
 | `modules/eks/images.yaml` | Bootstrap images to mirror to ECR |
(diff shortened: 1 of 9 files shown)


huydhn:
@claude Do a round of review on this change, highlight any potential blocking issues

huydhn (approved):
Oh well, Claude complains about a doc inconsistency, but not blocking issue IMO

jeanschmidt:
thanks for flagging the issue on the skill :)

I also folded up a small fix for smoke that seems to have broken with newer version of some dependency, just a string format

PR #1068, opened 2026-09-11
nodepools: park a Packer build for an NVIDIA 615 GPU AMI (CUDA 13.4)

huydhn:
Draft — parked, not proposed for adoption yet.

Builds an EKS AL2023 GPU AMI with the NVIDIA driver swapped from the stock **580** to **615.71.09** (R615), the branch CUDA 13.4 asks for. **Nothing selects it.** Every GPU pool still tracks the `amazon-eks-node-al2023-x86_64-nvidia-*` glob. This is the implementation sitting ready if we decide we need R615; adopting it is a separate per-def change.

### Overlaps #1064 — needs a call before this leaves draft

#1064 (`docs/nvidia-driver-ahead-of-aws.md`) is the broader survey of this same question, and lists "custom packer AMI" as the leading candidate. This is that candidate, built. The two docs overlap and both append a row to the same `CLAUDE.md` table, so they will conflict on merge.

Options, roughly in order of how much I like them:

1. Land #1064 first, then rebase this and fold my doc's net-new findings into `nvidia-driver-ahead-of-aws.md`, keeping only the packer README here.
2. Keep both docs — mine now cross-references #1064 as the decision doc and scopes itself to mechanics.
3. Close this if we're happy to wait for AWS, and keep the branch for later.

One correction for #1064 while it's open: it says CUDA 13.4 is a Developer Preview. **13.4.1 went GA in September 2026**; the July developer preview was 13.4.0.

### Why 580 is probably still fine

Worth stating plainly, because it argues against merging this:

| Need | Minimum driver |
|------|----------------|
| Run existing CUDA 13.x applications (minor version compatibility) | **>= 580** |
| CUDA 13.4's *new* features and newly enabled platforms | **R615** |

We're at 580. CUDA 13.4 binaries run today. R615 buys 13.4-specific features — Vera Rubin (dev preview), PTX ISA 9.4, GCC 16 / Clang 22 with nvcc — none of which is on the critical path. `pypi-cache` tops out at `cu130`, so nothing in the fleet targets 13.4 either. If the real goal is "run CUDA 13.4 workloads", the correct amount of infra work here is zero plus a `cu134` slug.

### Why a driver swap and not a build flag

The upstream EKS AMI build takes `nvidia_driver_major_version`, so `=615` looks like it should just work. Two independent blockers:

- It resolves the full version as `min(kmod-nvidia-open-dkms, AWS GRID runfile)`, and `s3://ec2-linux-nvidia-drivers` stops at **595.91.07**. No 610, no 615. `archive-grid-kmod` hard-errors with no matching runfile, and the `min()` would pin back to 595 anyway.
- `archive-proprietary-kmod` installs `kmod-nvidia-latest-dkms-<version>`. **R615 ships no proprietary kernel module** — the NVIDIA AL2023 repo's proprietary packages stop at 610.

Both are fixable only by forking the upstream template. Swapping the driver on the finished AMI keeps us on AWS's image and confines the delta to one script. (The GPU Operator third option doesn't apply: it doesn't manage drivers on Amazon Linux.)

### The part that would have bitten us

`/etc/eks/nvidia-kmod-load.sh` has to be patched or **nodes boot with no driver at all**. Upstream probes the major version with:

```bash
KMOD_MAJOR_VERSION=$(rpmquery kmod-nvidia-latest-dkms --queryformat '%{VERSION}' | cut -d. -f1)
```

That's the proprietary kmod, which doesn't exist at 615. The probe fails, `devices-support-open` returns 1, and selection falls through to a proprietary flavor that isn't there. Same shape as the g7 bug in awslabs/amazon-eks-ami#2768. A second guard redirects any flavor whose archive is absent (the g4dn/g5 GSP override, GRID subdevices) to `nvidia-open`, and drops `nvidia-disable-gsp.conf` with it since the open kmod requires GSP firmware.

Both patches are anchored on exact upstream text and **fail the build if an anchor moves**, so a base-AMI refresh can't ship a silently broken image.

### What the swap does

Follows upstream's own sequence: rebuild the DKMS module and re-archive it under `/var/lib/dkms-archive` (what `nvidia-kmod-load.service` unpacks at boot), rebuild `gdrdrv` against it, move userspace RPMs to the matching version, and generate `/etc/eks/nvidia-open-supported-devices-615.txt` from the `supported-gpus.json` the `nvidia-driver` RPM ships — upstream only publishes lists up to 595. `nvidia` (proprietary) and `nvidia-open-grid` archives are deleted and not replaced. Closing assertions fail the build if any `nvidia-*` package is left at 580 or the archived kmod isn't the target version.

### Scope

Compute GPUs — p4d (A100), p5 (H100), p6-b200 (B200). **Not** g4dn/g5/g6, which are exactly the pools depending on the proprietary kmod and the GSP workaround; the fallback guard means such a node boots rather than breaks, but that path is untested.

Two R615 items that sound alarming and don't apply: **CDMM** is GB200-class hardware only (all our GPU hosts are x86, and the EKS AMI already writes `NVreg_CoherentGPUMemoryMode=driver` unconditionally), and the **610→615 in-place upgrade hang** needs an in-place upgrade, which Karpenter nodes never do.

The real cost is that a fleet on this AMI stops picking up AL2023 and driver CVE fixes on node rotation, and nothing enforces a rebuild on an EKS version bump. That's the strongest argument for waiting — #1064 makes the same point about the base nodes being the only ones left unpatched for CVE-2026-64561.

### Test plan

- `just test` — pass, 98.84% coverage.
- `just lint` — 13/14. The failure is hadolint `DL3066` on `base/docker/runner-base/Dockerfile:36`, pre-existing on `main` and untouched here.
- `packer validate` clean; the `>= 615` variable guard correctly rejects 595.
- Both `nvidia-kmod-load.sh` patches replayed against the real upstream file: anchors match, patched result passes `bash -n`, and the fallback guard lands after the g4dn/g5 `esac` and before `kmod-util load`.

**Not done: the build has never been run against a real AMI.** It needs AWS (a `c7a.4xlarge` — DKMS compiles without a GPU, so it's cents). Treat the script as reviewed but unrun. Adopting it also needs a real GPU job end-to-end, and on p5 a fabric-handles check per `docs/h100-fabric-handles-imex-channels.md`, since IMEX is the most driver-sensitive thing in the fleet.

### If we want a cheap signal on R615 first

`gpu-dev/` already `dnf install`s the driver directly in `terraform-gpu-devservers/templates/al2023-user-data.sh` (landing on 595 today). Moving that to 615 is roughly a one-line change with no AMI build, and would tell us whether R615 is well-behaved on this hardware before OSDC commits.

### DIFF
--- a/osdc/CLAUDE.md
+++ b/osdc/CLAUDE.md
@@ -83,3 +83,4 @@ Reference documentation in `docs/`:
 | `docs/pypi-package-cache.md` | PyPI wheel cache architecture, slug naming, S3 layout, runner integration |
 | `docs/h100-fabric-handles-imex-channels.md` | Enabling CUDA fabric handles (`CU_MEM_HANDLE_TYPE_FABRIC`) on H100 runners — IMEX channel creation, the gated fabric runner, and verification |
 | `docs/dcgm-exporter-memory-tuning.md` | Why dcgm-exporter OOMs on high-vCPU GPU nodes (Go 1.24 GOMAXPROCS defaults to host vCPU count) and the GOMAXPROCS=2 fix |
+| `docs/nvidia-driver-615-ami.md` | CUDA 13.4 / NVIDIA R615 — why the stock EKS GPU AMI cannot reach it, and the parked (unadopted) Packer driver-swap in `modules/nodepools/packer/` |
--- a/osdc/docs/nvidia-driver-615-ami.md
+++ b/osdc/docs/nvidia-driver-615-ami.md
@@ -1,0 +1,170 @@
+# NVIDIA driver R615 for the compute GPU fleets (CUDA 13.4)
+
+Investigation and a parked implementation. **Nothing is adopted** — every GPU
+pool still runs the stock EKS GPU AMI at driver 580. This records what CUDA 13.4
+actually requires, why the obvious upgrade paths do not work, and what the build
+in `modules/nodepools/packer/` does if we decide we need it.
+
+Verified September 2026 against CUDA 13.4.1 GA and EKS AMI `v20260903`.
+
+> **Companion doc:** `docs/nvidia-driver-ahead-of-aws.md` (PR #1064) is the
+> broader survey — observed driver versions across recent AMI releases, the
+> `cuda-compat-13-4` and GPU Operator options, and the separate GPU-AMI-glob
+> Kubernetes skew bug it turned up. This doc is narrower: it is the build for
+> that survey's leading candidate, plus the two things that only surface once
+> you try it (the GRID runfile ceiling, and the boot-time kmod selection
+> failure). Read that one first for the decision, this one for the mechanics.
+> Note 13.4 has since gone GA, which that doc predates.
+
+## What CUDA 13.4 actually requires
+
+CUDA 13.4.1 went GA in September 2026 (13.4.0 was a July developer preview). Its
+corresponding driver branch is **R615** — `615.71.09` on Linux.
+
+The distinction that decides whether any of this is worth doing:
+
+| Need | Minimum driver |
+|------|----------------|
+| Run existing CUDA 13.x applications (minor version compatibility) | **>= 580** |
+| CUDA 13.4's *new* features and newly enabled platforms | **R615** |
+
+The fleet is already at 580. **CUDA 13.4 binaries run on it today.** R615 buys
+13.4-specific features — Vera Rubin support (developer preview), PTX ISA 9.4,
+GCC 16 / Clang 22 / libc++ 22 with nvcc, Windows-on-Arm — none of which is on
+the critical path for PyTorch CI as of this writing. `pypi-cache` currently
+builds `cu126`, `cu128` and `cu130` (`clusters.yaml`), so nothing in the fleet
+targets 13.4 yet either.
+
+If the only goal is "run CUDA 13.4 workloads", the correct amount of
+infrastructure work is **zero**, plus a `cu134` slug in pypi-cache.
+
+## Where the driver comes from today
+
+Nothing in `osdc/` pins a driver version. GPU nodepools take whatever the newest
+EKS GPU AMI ships:
+
+```python
+# modules/nodepools/scripts/python/generate_nodepools.py
+if is_gpu:
+    ami_family_block = "  amiFamily: AL2023"
+    ami_selector_block = """  amiSelectorTerms:
+    - name: "amazon-eks-node-al2023-x86_64-nvidia-*\""""
+```
+
+That glob covers g5/g6/g4dn, p4d, p5 (`nodepools-h100`) and p6-b200
+(`nodepools-b200`). `base_node_ami_version: "v20260903"` in `clusters.yaml` pins
+only the CPU `-standard-` image for base nodes. The H100 and B200 userData
+scripts do persistence mode and IMEX channel setup but explicitly rely on the
+AMI for the driver and Fabric Manager.
+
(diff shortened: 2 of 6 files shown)


huydhn:
Reworked (`22e687e`) — this now patches upstream's AMI build instead of swapping the driver on a finished image. **I had the blocker wrong**, and the correction changes the approach.

**What I got wrong.** I claimed a from-source build could not reach 615 because AWS's GRID runfile bucket stops at `595.91.07` and upstream resolves the version as `min(kmod-nvidia-open-dkms, GRID runfile)`. But `nvidia_grid_runfile_bucket_name` is an ordinary Packer variable — mirroring the bucket is supported configuration. I presented an overridable input as a wall. The hedge about it "silently pinning back to 595" was also wrong: the lookup greps for the *requested* major, so at 615 it returns empty and the build exits 1 cleanly.

**The real blocker is narrower.** `archive-proprietary-kmod` runs unconditionally and installs `kmod-nvidia-latest-dkms-<version>`, which NVIDIA stopped building at 610:

| | present in the NVIDIA AL2023 repo |
|---|---|
| `kmod-nvidia-open-dkms` | …595.91.07, 610.43.02, 610.57.04, **615.71.09** |
| `kmod-nvidia-latest-dkms` | …595.91.07, 610.43.02, 610.57.04 — **none at 615** |

That's a dnf package, not an S3 object, so no bucket helps. But it's one patch, not an impossibility.

**And skipping it has a consequence I'd missed.** `archive-proprietary-kmod` is also what leaves the `kmod-nvidia-latest-dkms` *RPM* installed on the finished image — it archives the module and `kmod-util remove`s it, but never `dnf remove`s the package. `nvidia-kmod-load.sh` then reads the driver's major version from exactly that RPM at boot. So the boot probe breaks on *any* route to 615. The swap never avoided that patch; it just applied it by sed-ing a shipped file with anchors. Patching upstream puts the same fix somewhere it reads as a bug report.

Once `archive-grid-kmod` is skipped too — a compute-only image never selects the GRID flavor — the GRID lookup drops out of version resolution and the runfile mirror stops being a question at all. **No bucket needed.**

**What's here now:**

```
build-gpu-ami.sh   clone upstream @ v20260903 -> git apply patches/ -> make -> tag the AMI
patches/0001-…     build only the open kmod on branches that ship only the open kmod
patches/0002-…     stop the boot selector depending on the proprietary kmod's RPM
```

Both patches gate on `NVIDIA_DRIVER_MAJOR_VERSION >= 615`, so they're no-ops on every branch AWS currently builds. They're written to be upstreamable — `0002` also fixes the g7 selection bug in awslabs/amazon-eks-ami#2768 as a side effect, which might be worth sending regardless of what we decide here.

`git apply` runs without `--3way` and without fuzz: if a patch stops applying, upstream moved and that needs a human, not a merge.

**Verification.** Both patches apply cleanly to a fresh clone at the tag; both patched files pass `bash -n`; and I exercised the patched version-resolution against stubbed inputs:

| case | result |
|---|---|
| 615, no GRID runfile published | `open_only=true full=615.71.09` |
| 580, stock path, versions equal | `open_only=false full=580.178.04` |
| 580, GRID older → must pin to GRID | `open_only=false full=580.159.03` |
| 615, open kmod also missing | exits 1, no fallback |

The two 580 rows matter most: stock behaviour is unchanged, including the "GRID is older, pin to it" case.

`just lint` 13/14 (pre-existing hadolint `DL3066` on `base/docker/runner-base/Dockerfile`), `just test` passing.

**Still unrun end to end** — the build needs AWS and has never produced an actual AMI. The swap implementation is preserved at `e5b1184` if we decide maintaining a patched fork is worse than patching a shipped file.

PR #1069, opened 2026-09-13
[grafana] Add PyTorch Test Metrics dashboard

georgehong:
Standalone page tracking config multiplicity: how many distinct build x config combinations execute the same test on one main commit. Kept out of Core Metrics since this is diagnostic, not a mainline KPI.

The commit selector reads the viable/strict advance log, filtered to commits whose runs have all finished and none cancelled. viable/strict alone is not enough: it gates on required jobs, so a recent commit can still be mid-flight. Arbitrary shas can be typed in, so a completeness panel guards that path. A config counts only where the test executed, xfail included and skips not. Configs are counted distinct, so the replicator's duplicate inserts cannot inflate them, and shard indices are stripped so shards of one config fold together.

Authored with Claude Code.


georgehong:
Preview at https://pytorchci.grafana.net/goto/sr4bs4. We can observe pull vs trunk differences:
- When filtered to trunk, one can observe heavy concentration at multiplicity 3 -- these are our three GPU runs, CUDA 13.0, 13.2, and ROCm
- When filtered to pull, this distributes heavily to 6-7, and another at 10 or so corresponding to build variations.

huydhn (approved):
I need to sit with you to understand what these metrics mean and how to use this dashboard, but this looks like an internal dashboard only for us, so stamp to unblock

jathu (approved):
💯

PR #1071, opened 2026-09-14
image-cache-janitor: drop the unused util-linux package from the image

jeanschmidt:
**Impact:** image-cache-janitor DaemonSet on all clusters
**Risk:** low

## What
Removes the `util-linux` apk install from the janitor's Dockerfile and cleans up the three comments that still claimed the janitor reaches the host via `nsenter`.

## Why
The janitor talks to containerd directly: it runs the `crictl` binary bundled in the image, pointed at the host CRI socket via `CONTAINER_RUNTIME_ENDPOINT` and a hostPath mount. Nothing shells out through `nsenter`, so `util-linux` was dead weight — and a pinned version (`=2.41.6-r1`) that breaks the build every time Alpine rotates the package out of its repo index.

# Notes
The deploy tag is a hash of the Dockerfile, so this forces one rebuild and push of both arch images on the next `deploy-base`.

Two leftovers not touched here, both fine as follow-ups: the `.trivyignore` KSV-0010 comment still justifies the janitor's `hostPID: true` as "nsenter-based debugging access", and it's worth checking whether `hostPID` is needed at all now.


huydhn (approved):
LGTM!

huydhn:
I vaguely remember seeing this before, but deleting code is fine as long as the image can be built

PR #1073, opened 2026-09-15
Fix the GPU release runner's disk request so it can schedule

huydhn:
`rel-l-x86aavx2-29-113-l4` asked for **600Gi of ephemeral storage on a g6.8xlarge whose nodepool provisions a 600GB disk**. Allocatable is below that once the AMI, container images and kubelet reserves are taken, so Karpenter could never place the pod. The scale set registered with GitHub and looked healthy while every job sat queued — pytorch/vision#9668's nightly CUDA wheels queued indefinitely while the CPU rows on `rel-l-x86iavx512-44-340` finished normally in the same run.

```
rel-l-x86aavx2-29-113-l4   disk_size: 600   node_disk_size: 600   ← never fits
l-x86aavx2-29-113-l4       disk_size: 150   node_disk_size: 600   ← CI runner, same instance, works
rel-l-x86iavx512-44-340    disk_size: 600   node_disk_size: 900   ← CPU release runner, works
```

600 came from copying the CPU release runner in #1067, which works only because its nodepool provisions 900 for that request. 150Gi is what both the CI runner on this same instance and the `linux.g5.4xlarge.nvidia.gpu` EC2 runner this replaces already use for the same builds, so it is not a reduction in practice.

Renders as `ephemeral-storage: "150Gi"`; runner and nodepool generator tests pass.


atalman (approved):
lgtm

PR #1076, opened 2026-09-15
Retune OSDC Service Status so a non-green tile means something

jeanschmidt:
**Impact:** OSDC Service Status dashboard
**Risk:** low

deployed dashboard: https://pytorchci.grafana.net/d/ci-infra-f4vfmq-osdc_services_status/osdc-services-status

## What

Rewrites the queries, thresholds and not-deployed handling of `osdc_services_status.json`, and fixes the alert rules that had the same defects. The nine DaemonSet `% healthy` tiles become stuck-pod severity tiles (pods unavailable for 15 continuous minutes, in excess of the cluster's NotReady node count, over a budget of 0.5% of desired pods with a floor of 1.5). Optional modules (buildkit, keda, hf-cache, dcgm) now render `NOT DEPLOYED` / `NO DATA — DEPLOYED` instead of red. The four summary rollups count broken workloads and degraded services rather than unavailable replicas and individual down scrape targets, and explicitly count "kube-state-metrics is down" as a signal instead of quietly reporting green. Remaining thresholds were recalibrated against 90 days of fleet data. Panel titles get the `[cluster]` variable suffix the repo convention requires.

On the alerting side: `HarborCacheRecoveryFailing` could never clear (the gauge is sticky and the CronJob retains failed Jobs), `HarborCacheRecoveryStale` aggregated fleet-wide instead of per cluster, the conntrack alerts were tuned to saturation levels the fleet has never approached, `ControlPlaneCrashLoop` moves 5 → 10 restarts/h to match the dashboard's red line, and a new `NodeInitTaintStuck` alert catches nodes wedged in node-init.

## Why

The dashboard was unreadable. Over a 14-day window at least one summary tile was non-green in 100% of minutes and at least one was RED in 100% of minutes; the ten rollup tiles changed colour ~96 times an hour against a 30 s refresh. Nine tile-instances were permanently red purely because the module isn't deployed on that cluster. Zero genuine DaemonSet failures happened in that window, yet the `% healthy` tiles ran 7.5–27.7% non-green — node churn was being counted as service failure. Meanwhile three real production problems were on screen and indistinguishable from the noise, and a wedged GPU node that had been dead for 52 hours showed six DaemonSet tiles green.

Every threshold in here is now backed by measured distributions (or is explicitly labelled engineering judgement in the panel description, with the observed range stated).

# Notes

What is genuinely firing after these changes, and why:

| tile | state | what it is |
|---|---|---|
| Workloads, uw1 | red 1 | orphaned `ai-sandbox/sandbox-agent` — real, zero user impact, debris from a deploy misfire | 
| Worst Container Restart Rate, lf-ue1 | red ~19/h | node-exporter crashloop — live incident, still accelerating |
| OOM / Restarts, lf-ue1 | yellow 1 | same incident | | KEDA Scalers, both LF | purple -1 | real IaC drift, 81 days old |
| DCGM Exporter, uw1 (intermittent) | purple -1 | GPU nodes present, nothing scraping them — 5.46% of the last 7 days |
| Harbor Cache Recovery, ue1/ue2 | yellow during bursts | the unpaginated-pod-list OOM |

Everything else is green, and each of the above is a thing someone should actually go fix.


jeanschmidt:
fixed a small discrepancy between main board and service detail board

huydhn:
@claude Verify if all the queries here are ready to go, any edge cases or blocking issue to highlight

huydhn (approved):
Claude says go :)

PR #1077, opened 2026-09-15
[grafana] Add Total tests run panel, key configs by workflow

georgehong:
Total tests run counts distinct test x configuration executions, the weighted sum of the config multiplicity distribution.

Prefix the config key with the workflow name. aarch64 test (default) is the only job name duplicated across pull and trunk, and the two run different hardware (Graviton4/SVE128 vs Graviton3/SVE256), so the old key merged two real configurations. Scopes are now additive.

Accept hyphens when stripping shard indices so configs like inductor-halide normalize. Correct the multiplicity bucket description and drop an unused bc column from Unique tests run.


georgehong:
preview at https://pytorchci.grafana.net/goto/s6vp89
- Chart already reflects the CUDA update changing the 3 --> 2 peak. Periodic is a scheduled workflow so this chart does not display it.

jathu (approved):

PR #1078, opened 2026-09-15
Give the g6 release nodepool a fallback ladder

huydhn:
The 1-GPU release runner's workflow pod asks Karpenter for **~126Gi** once its sidecars are counted, not the 113Gi in the runner def, and that does not fit a `g6.8xlarge`. Pinned to `g6.8xlarge` alone the nodepool had nowhere to fall back to:

```
no instance type has enough resources, requirements=instance-type In [g6.8xlarge], …
resources={"cpu":"30355m","ephemeral-storage":"150Gi","memory":"125790Mi","nvidia.com/gpu":"1"}
```

The runner pod carries no runner-class affinity so it scheduled fine, and the scale set registered and looked healthy — only the workflow pod, which is the one pinned to the release nodepool, stayed Pending. pytorch/vision#9668's nightly CUDA wheels queued for hours against it while the CPU rows on `rel-l-x86iavx512-44-340` finished normally in the same run.

**The sizing was never the problem.** The identically specced CI runner `l-x86aavx2-29-113-l4` — same instance, disk, vcpu, memory, gpu — never shows this, because the CI g6 fleet offers five sizes and Karpenter places the same pod on a `g6.12xlarge`. That is also why no `g6.8xlarge` has ever been provisioned in these clusters. #1067 inherited an instance choice from a def that had never actually landed on it.

Ordered smallest-first, unlike the CI ladder: release nodes are isolated, so a larger size is only worth taking when the smaller cannot be had. A `g6.12xlarge` holds four of these runners.

Renders `g6-{12,16,24,48}xlarge-release` with `osdc.io/runner-class: release`; nodepool and runner generator tests pass.


huydhn:
Superseded by the sizing fix — a ladder would have scheduled, but onto a g6.12xlarge that fits one of these pods and strands three GPUs. Root cause was the 113Gi request exceeding a g6.8xlarge's allocatable once pod and daemonset overhead are counted.

PR #1079, opened 2026-09-15
Size the GPU release runner to fit the g6.8xlarge it names

huydhn:
rel-l-x86aavx2-29-113-l4 asks for 113Gi, which does not fit. Karpenter schedules
the job plus its sidecars and the node's daemonsets: 125790Mi for a 113Gi job,
9.8Gi above the job alone and 4.4Gi over the 121241Mi that instance_specs.py
records as g6.8xlarge allocatable. 104Gi leaves 4.6Gi of margin.

The CI runners on the same instance carry the same figure but never surfaced it,
because their fleet offers five sizes and Karpenter quietly takes the next one
the weight ladder allows. This runner's fleet generates a single release
nodepool, g6-8xlarge-release, so there is nothing to fall through to: the pod
stays Pending and pytorch/vision's nightly CUDA wheels queue for hours while the
CPU rows finish.

#1067 inherited the memory figure from a def whose comment accounted for node
overhead but not pod overhead. The two CI defs are wrong the same way and are
left for a follow-up, where the cost of their silent fallthrough is the point.

The label keeps -113-. Renaming is a cross-repo contract change -- pytorch's
.github/arc.yaml maps EC2 labels onto these names -- and the number Karpenter
reads is the request, not the label.


huydhn:
Folded into #1080, which applies 104Gi to all three 1-GPU runners rather than only the release one.

PR #1080, opened 2026-09-16
Right-size the 1-GPU GPU runners, and add 4xlarge-sized ones

huydhn:
Two related changes to the g5/g6 GPU fleets. **Folds in #1079.**

### Right-sizing the existing 1-GPU runners

`l-x86aavx2-29-113-a10g`, `l-x86aavx2-29-113-l4` and `rel-l-x86aavx2-29-113-l4` all ask for a 113Gi job, which does not fit the 8xlarge they name. Karpenter charges 1355m / 10078Mi on a workload node, so 113Gi demands `115712 + 10078 = 125790Mi` against 121241Mi of capacity — the exact number it reported when the release runner sat Pending and pytorch/vision's nightly CUDA wheels queued for hours. 104Gi leaves 4667Mi of margin.

The two CI runners never surfaced it: their fleets offer five sizes, so Karpenter quietly took the next one the ladder allowed and paid for a larger node. The release runner had a single pool with nothing beneath it, so it never scheduled at all. Both now land on the 8xlarge they are sized for.

Labels keep `-113-`. Renaming is a cross-repo contract change — pytorch's `.github/arc.yaml` maps EC2 labels onto these names — and Karpenter reads the request, not the label.

### New 4xlarge-sized runners

The fleet has no 1-GPU runner that fits a 4xlarge, so this adds `l-x86aavx2-11-41-a10g` (g5.4xlarge), `l-x86aavx2-11-41-l4` (g6.4xlarge) and the release-class `rel-l-x86aavx2-11-41-l4`. A 4xlarge is the smallest node in either family carrying a GPU, and a 1-GPU pod takes a whole node whatever its size — two would fit an 8xlarge on CPU and memory, but it has one GPU.

11 vCPU / 41Gi is the largest pod that still packs **GPU-bound on every size**, so nothing is stranded:

| node | GPUs | pods | bound by | cpu | mem |
|---|---|---|---|---|---|
| 4xlarge | 1 | 1 | GPU | 12355/16000 | 52062/60620 |
| 8xlarge | 1 | 1 | GPU | 12355/32000 | 52062/121241 |
| 12xlarge | 4 | **4** | GPU | 45355/48000 | 178014/181862 |
| 16xlarge | 1 | 1 | GPU | 12355/64000 | 52062/242540 |
| 24xlarge | 4 | 4 | GPU | 45355/96000 | 178014/363724 |
| 48xlarge | 8 | **8** | GPU | 89355/192000 | 345950/727449 |

The 12xlarge sets the ceiling — `(48000m - 1355m)/4 = 11661m` and `(181862Mi - 10078Mi)/4 = 41.9Gi`. On the 4xlarge that leaves 3645m and 8.4Gi spare, which also absorbs risk in `memory_mi` being the 0.925 estimate: no 4xlarge GPU node has been provisioned, so run `collect_instance_memory.py` against the first one.

Sizing throughout follows Karpenter rather than `analyze-utilization`, per the "Caveat for any near-full-node runner" in `docs/node-utilization-optimization.md` — the tool's DaemonSet figure is several Gi light, which is how the 113Gi defs were arrived at in the first place.

### Ladders

```
g5/default   48xl=100  12xl=90  24xl=80  4xl=75  8xl=70  16xl=60
g6/default   48xl=100  12xl=90  24xl=80  4xl=75  8xl=70  16xl=60
g6/LF        4xl=100   8xl=95   24xl=90  16xl=80 12xl=70 48xl=0
g6/release   4xl=100   8xl=95   12xl=90
```

In CI the 4xlarge ranks above the other 1-GPU sizes and below the multi-GPU ones. On g5 and g6 default that is a free slot with no existing weight moved; on LF the 8xlarge gives up the 100 cap.

The release block goes from one pool to three, smallest node that fits first. Packing four 11-41 pods onto a 12xlarge would be cheaper per runner-hour, but release builds are bursty and scale from zero, so one pod alone on a 4-GPU node is the case to avoid. The 12xlarge takes the bottom rung instead, because a release class with nothing below its smallest size is what stranded the release runner to begin with.

Resulting placement:

| runner | lands on | GPUs idle |
|---|---|---|
| `l-x86aavx2-29-113-{l4,a10g}` @104Gi | 8xlarge | 0 |
| `rel-l-x86aavx2-29-113-l4` @104Gi | g6.8xlarge | 0 |
| `l-x86aavx2-11-41-{l4,a10g}` | 4xlarge | 0 |
| `rel-l-x86aavx2-11-41-l4` | g6.4xlarge | 0 |


huydhn:
I found out this issue today about g5/g6 CI runners we are using, i.e. https://github.com/pytorch/ci-infra/blob/main/osdc/modules/arc-runners/defs/l-x86aavx2-29-113-l4.yaml. Their spec is 29 CPU and 113GB of memory. That's corresponding to a 8xlarge GPU runner https://aws.amazon.com/ec2/instance-types/g6. This doesn't sound right, we only need g5/g6.4xlarge before, so the mapping https://github.com/pytorch/pytorch/blob/main/.github/arc.yaml#L73 is over spec.

This PR prepares the runner def corresponding to g5/g6 4xlarge, so that we can try it out on CI.

zxiiro:
> I found out this issue today about g5/g6 CI runners we are using, i.e. https://github.com/pytorch/ci-infra/blob/main/osdc/modules/arc-runners/defs/l-x86aavx2-29-113-l4.yaml. Their spec is 29 CPU and 113GB of memory. That's corresponding to a 8xlarge GPU runner https://aws.amazon.com/ec2/instance-types/g6. This doesn't sound right, we only need g5/g6.4xlarge before, so the mapping https://github.com/pytorch/pytorch/blob/main/.github/arc.yaml#L73 is over spec.
> 
> This PR prepares the runner def corresponding to g5/g6 4xlarge, so that we can try it out on CI.

Looks like we found the same issue. This was on my list to investigate further before I went to China last week. Adding my own notes in case it's useful.

> The dominant g6 runner (l-x86aavx2-29-113-l4, GitHub label linux.g6.4xlarge.experimental.nvidia.gpu) requests 29 vCPU / 1 GPU, while cluster-wide GPU utilization is 100% (GPU is the binding resource per just simulate-cluster). Every vCPU bought without a GPU attached is waste, and the current 29:1 vCPU:GPU ratio is why multi-GPU instances (e.g. g6.12xlarge, 48 vCPU / 4 GPU) can't be filled by four 1-GPU jobs (4x29=116 > 48).
> Notably the GitHub label says "g6.4xlarge" (16 vCPU / 1 GPU) but the runner def requests 29 vCPU — roughly double the labeled size — and we back it with a g6.8xlarge, giving 1 runner/node (confirmed via just simulate-cluster: 377 g6.8xlarge nodes for 377 runners).

zxiiro (approved):

huydhn:
Going to try to deploy this to staging and test it out first before merging

jeanschmidt (approved):
great optimization!

PR #1087, opened 2026-09-19
Add a release A10G runner so release matches non-release

huydhn:
The release GPU runner is an L4 while its non-release counterpart is an A10G, so a PR and the nightly that follows validate the same wheel on different architectures.

executorch needs this. The torch wheel its CUDA rows resolve ships sm_75/80/86/90/100/120, so the smoke test passes on the A10G (sm_86) a PR uses and fails on the L4 (sm_89) the nightly uses:

```
AssertionError: the torch resolved for this x86_64 row carries no code for
sm_89, the runner's own GPU, so the execution check cannot run
```

No PR can reproduce it — the L4 is only reachable from a published ref.

`rel-l-x86aavx2-11-41-a10g` is the release twin of `l-x86aavx2-11-41-a10g`: same g5.4xlarge, same 11 vCPU / 41Gi. g5 gains a `release:` block to host it, laid out like g6's so an oversized release pod falls through instead of stranding in Pending (#1079).

Pointing the matrix at it is a separate test-infra change, and `rel-l-x86aavx2-11-41-l4` stays until that lands.


osdc/modules/arc-runners/defs/rel-l-x86aavx2-11-41-a10g.yaml
```diff
@@ -0,0 +1,24 @@
+# ARC runner definition: rel-l-x86aavx2-11-41-a10g
+# Release wheel-build runner (x86, GPU) — dedicated runner group + node isolation.
+# Instance type: g5.4xlarge — 1-GPU A10G, 16c/64Gi node (pool: g5-4xlarge-release).
+# Release-class twin of l-x86aavx2-11-41-a10g; see it for the sizing.
+#
+# A10G rather than L4 so the release runner matches the architecture of the
+# non-release runner the same wheel rows build on. rel-l-x86aavx2-11-41-l4 is an
+# L4 (sm_89) while its non-release counterpart is an A10G (sm_86), and the torch
+# wheel those rows resolve carries no sm_89, so executorch's nightly smoke test
```
jeanschmidt: I dont think this is broken, but just a heads up if you want to follow up

---

<details>
<summary>🟡 The sm_89 premise this change is built on is contradicted by rel-l-x86aavx2-29-113-l4.yaml, which stays on an L4 and is left unaddressed (ai-generated section)</summary>

The change's entire justification is that the CUDA x86_64 release rows cannot execute on an L4 because the torch they resolve carries no sm_89. Two existing records in the same repository assert the opposite for the same rows, and the diff reconciles neither. The repo's other CUDA-executing release runner, rel-l-x86aavx2-29-113-l4, is left on a g6/L4 pool with a comment stating its smoke test runs a model on the device — by the premise of this diff that runner is broken the same way, yet nothing in the diff or the description touches it. osdc/modules/arc-runners/defs/rel-l-x86aavx2-29-113-l4.yaml:11-14 says "Exists for the CUDA x86_64 wheel rows, whose smoke test exports a model through the CUDA partitioner and runs it, so unlike the CPU release runners this one needs a device. L4 is sm_89, which the x86_64 rows already compile for (.ci/scripts/wheel/cuda_arch_list.sh: 8.0 8.6 8.9 9.0 10.0 12.0)." osdc/modules/nodepools/defs/g6.yaml:19-20 says "g6.8xlarge rather than g5.8xlarge: same 32c/128Gi 1-GPU shape, cheaper per runner-hour, and L4 (sm_89) is in the arch list those wheels compile for." The new comment here and at osdc/modules/nodepools/defs/g5.yaml:13-15 says the opposite: the torch "ships sm_75/80/86/90/100/120" so the "execution smoke test ... fails on an L4 (sm_89)". Both existing files scope themselves to "the CUDA x86_64 wheel rows", the same rows the description names. rel-l-x86aavx2-29-113-l4 keeps instance_type g6.8xlarge and runner_class release (lines 17 and 22) and is untouched by this diff. The step I inferred, and could not verify from this repository, is which test-infra matrix rows resolve to which of the two release runners; the mapping lives outside this repo. If the new premise is correct, the executorch CUDA nightly keeps failing on whatever rows land on rel-l-x86aavx2-29-113-l4 after this ships, and the next person debugging it starts from a def whose comment tells them L4/sm_89 is fine. If the recorded rationale is correct instead, this change adds a scale set and three g5 release NodePools per cluster to solve a problem that does not exist. Nothing in the merged tree lets a reader tell which. Narrow fix: state in the PR description (and in the new def's comment) why rel-l-x86aavx2-29-113-l4 is unaffected — e.g. that its rows build but never run the execution check — and correct the now-wrong sm_89 sentence at osdc/modules/nodepools/defs/g6.yaml:19-20 and rel-l-x86aavx2-29-113-l4.yaml:13-14 so the two rationales no longer disagree. Broader fix, which I would pick if the 29-113 rows do execute on the device: add rel-l-x86aavx2-29-113-a10g on g5.8xlarge in the same commit, which the new g5 release block already has a pool for, so both CUDA release rows move off the L4 together. Either way the contradictory comments must be resolved rather than left side by side.

<sub>Reviewed by claude-opus-5[1m] at max effort, against 95b5f8532.</sub>

</details>

jeanschmidt (approved):
## Review summary

🟡 1 minor

- 🟡 The sm_89 premise this change is built on is contradicted by rel-l-x86aavx2-29-113-l4.yaml, which stays on an L4 and is left unaddressed — `osdc/modules/arc-runners/defs/rel-l-x86aavx2-11-41-a10g.yaml:9`

<details>

**What this changes**

Two files, additive only. A new ARC runner def, osdc/modules/arc-runners/defs/rel-l-x86aavx2-11-41-a10g.yaml, adds a release-class GPU scale set on g5.4xlarge at 11 vCPU / 41Gi — the release twin of l-x86aavx2-11-41-a10g and a same-shape alternative to the L4-backed rel-l-x86aavx2-11-41-l4, which stays in place. osdc/modules/nodepools/defs/g5.yaml gains a release: block that generates g5-4xlarge-release, g5-8xlarge-release and g5-12xlarge-release NodePools per cluster (us-west-1 excluded), labelled osdc.io/runner-class: release. Nothing in the diff points any matrix row at the new runner. Roughly half the added lines are comment arguing why A10G rather than L4.

**What it gets right**

The nodepool half is correct and necessary rather than incidental: release-class runners get required node affinity onto runner-class release nodes, so a g5-backed release runner cannot schedule at all without a g5 release pool. The ladder mirrors g6's, including the 12xlarge at the bottom purely as a fallthrough, which is the shape that stops an oversized release pod stranding in Pending (#​1079). The def omits runner_group and records why — release-class runners take the derived per-region group, so a def-level value would be ignored — matching the sibling release defs, as do the capacity fields. Keeping rel-l-x86aavx2-11-41-l4 leaves the change decoupled from the matrix repoint.

**Where the risk is**

Everything but one nit lands in one place: this repository's record of which GPU the CUDA x86_64 release wheel rows need, and which release runner serves them. Three findings share cluster c-001, and one edit settles all three. The new comments argue A10G on the grounds that the resolved torch wheel ships no sm_89; two untouched comments argue the opposite conclusion for the same rows — osdc/modules/nodepools/defs/g6.yaml:19-20, and osdc/modules/arc-runners/defs/rel-l-x86aavx2-29-113-l4.yaml:11-14, which describes itself as the runner whose smoke test runs a model on the device. Strictly, the old text is about executorch's own cuda_arch_list.sh and the new text about the torch dependency it resolves, so both sentences can be true at once; what conflicts is the conclusion each draws, and nothing marks either as superseded. Three findings converged on that region independently. What cannot be settled from this checkout is which release runner the failing rows actually target — that mapping lives in test-infra. If it is the 29-113 runner, this adds a scale set of the wrong shape and the nightly keeps failing; if it is the 11-41 runner, the 29-113 rationale is merely stale. Separate from all of that, the new comment's "its non-release counterpart" reads, under the repo's naming convention, as l-x86aavx2-11-41-l4, which is an L4, not the A10G the sentence claims.

**Before merge, and after**

Reconcile the record. Whichever way the matrix question resolves, "g6.yaml:20 and rel-l-x86aavx2-29-113-l4.yaml:13 still document L4/sm_89 as the right GPU" should not sit beside the opposite claim in g5.yaml. The counterpart naming point is a one-line edit: name l-x86aavx2-11-41-a10g explicitly. Severity labels split these across the range; the edit that settles them is the same one, and it is cheap.

Follow-up: "the def this repo records as serving the CUDA x86_64 wheel rows stays L4-only" is a design decision, not a typo. If the 29-113 rows do execute on the device, they need a rel-l-x86aavx2-29-113-a10g on g5.8xlarge — a pool the new ladder already provides — sequenced with the test-infra repoint and the eventual removal of rel-l-x86aavx2-11-41-l4.

</details>

osdc/modules/arc-runners/defs/rel-l-x86aavx2-11-41-a10g.yaml
huydhn: The bot comment here is just junk IMO, let me remove it. Basically, I want to have an a10g release runner in addition to the existing l4 one. The user (ET) can use whatever they want, the release fleet `proactive_capacity` is 0 anyway. Let me remove it.

osdc/modules/arc-runners/defs/rel-l-x86aavx2-11-41-a10g.yaml
```diff
@@ -0,0 +1,18 @@
...
+  instance_type: g5.4xlarge
+  disk_size: 150
+  vcpu: 11
+  memory: 41Gi
+  gpu: 1
+  runner_class: release
+  # No runner_group here: release-class runners always land in the per-region
+  # group {cluster_group}-release-runners, derived by generate_runners.py from
+  # the cluster config (a def-level runner_group would be ignored for them).
+  proactive_capacity: 0
```
jeanschmidt, line 16: hahahaha LOL claude is neurotic most of the times

---

<details>
<summary>⚪ The description's "an unused one costs nothing" omits the per-scale-set ARC listener pod, which runs continuously at 250m CPU / 1331Mi (ai-generated section)</summary>

The PR description justifies keeping both the L4 and the A10G release runners with "both sit at `proactive_capacity: 0`, so an unused one costs nothing". The `proactive_capacity: 0` on this line does suppress pre-warmed EC2 capacity, but each runner def becomes its own AutoscalingRunnerSet and ARC keeps one AutoscalingListener pod long-polling GitHub for it regardless of runner count, so the marginal cost of an unused scale set is not zero. osdc/modules/arc-runners/templates/runner.yaml.tpl:150-158 gives the listener container requests and limits of `cpu: "250m"` / `memory: "1331Mi"`, and that block is rendered once per def — generate_runners.py:626-638 loops over every file in modules/arc-runners/defs/. osdc/modules/arc-runners/tests/smoke/test_runners.py:315-318 asserts listener pods are present in the arc-systems namespace for deployed scale sets, and test_capacity_parity.py:294-300 separates "scale sets with proactive_capacity &gt; 0" (the ones that also get placeholder pods) from the rest, confirming the listener runs either way. This diff takes the def count in that directory from 48 to 49. On each cluster that deploys arc-runners outside us-west-1 (g5 is excluded there), the new scale set adds a permanently-running listener pod reserving 250m CPU and 1331Mi of memory on the system nodes even if no release job ever targets the A10G. The claim is correct about EC2 spend and wrong about cluster footprint, so anyone weighing "keep both" against "replace the L4" is deciding against a zero that is not zero. No YAML change is needed — the code is right. Restate the description as "an unused one costs no EC2 capacity; the marginal cost is one listener pod (250m CPU / 1331Mi) per cluster", so the keep-both decision is made against the real number.

<sub>Reviewed by claude-opus-5[1m] at max effort, against b33cf64e9.</sub>

</details>

jeanschmidt (approved):
No issues, just merge

## Review summary

⚪ 1 nit

- ⚪ The description's "an unused one costs nothing" omits the per-scale-set ARC listener pod, which runs continuously at 250m CPU / 1331Mi — `osdc/modules/arc-runners/defs/rel-l-x86aavx2-11-41-a10g.yaml:16`

<details>

**What this changes**

The change adds one ARC runner definition, `osdc/modules/arc-runners/defs/rel-l-x86aavx2-11-41-a10g.yaml`: a release-class scale set on g5.4xlarge at 11 vCPU / 41Gi / 1 GPU, with `proactive_capacity: 0`, `hud_failure_base_capacity: 0`, and `max_burst_capacity: 10`. It also adds a `release:` section to `osdc/modules/nodepools/defs/g5.yaml` listing g5.4xlarge, g5.8xlarge and g5.12xlarge at weights 100/95/90, which makes `generate_nodepools.py` emit `g5-*-release` NodePools labelled `osdc.io/runner-class: release` (generate_nodepools.py:732-740). Nothing is removed: the existing L4 release runner and the g5 `instances:` ladder are untouched.

**What it gets right**

The two halves land together. Release-class runners get a required node affinity on `osdc.io/runner-class: release` (generate_runners.py:390-400), so a release def on a g5 instance type with no g5 release pool would have nowhere to schedule; the nodepool section that makes it schedulable is in the same change. The `release:` block copies g6's shape exactly — same three sizes, same weights, smallest first with a larger size at the bottom — which is the arrangement that keeps an oversized release pod from pending against a class whose only pool is too small. The def mirrors `rel-l-x86aavx2-11-41-l4.yaml` line for line, including the comment about omitting `runner_group`, and that comment matches the code: `generate_runners.py:336-337` derives the group for release-class runners and ignores a def-level value.

**Where the risk is**

The review turned up nothing wrong in the YAML. What it did surface sits in the pull request's prose rather than the code: the claim that an unused runner costs nothing, anchored at the `proactive_capacity: 0` line of the new def. Each def becomes its own AutoscalingRunnerSet, and ARC keeps one listener pod per scale set regardless of runner count, requested and limited at 250m CPU / 1331Mi in `templates/runner.yaml.tpl:150-158`. So the marginal cost of keeping both the L4 and the A10G is zero EC2 and one listener pod per cluster. That is a single observation, not a pattern — there is no cluster of related findings here and no region of the diff that several passes converged on. The shape of the risk is that the change's mechanics are small and copied from a working precedent, while the judgment call it encodes — keep both architectures rather than replace one — rests on a cost figure that is right about instances and incomplete about cluster footprint.

**Before merge, and after**

Nothing in the code blocks this. The one thing worth doing first is the wording fix behind "an unused one costs nothing" on the new def in `osdc/modules/arc-runners/defs/` — it is a one-line edit to the description, and it is worth doing before merge precisely because that sentence is the argument for carrying both release GPU runners. Restating it as no EC2 cost plus one listener pod per cluster makes the keep-both decision reviewable on the real number. As follow-up: if both release GPU scale sets stay idle, the listener footprint is the figure to weigh when deciding whether the fleet keeps both.

</details>

huydhn:
Oh darn, I forgot to merge this. I thought I merged this already, merging it now.

